PR #5, opened 2016-10-05
fix init.rc replacement for non-standard devices

jemyzhang:
not all "init.environ.rc“ is imported from the init.rc. for example, Nokia-N1:
init.rc

> import /init.aosp.rc

init.aosp.rc

> import /init.environ.rc
> ...


topjohnwu:
This is already done, forgotten to close

PR #64, opened 2017-02-11
fix up bump support for LG g2/g3 devices

Drgravy:
unfortunately "d80*" wouldn't work for me, but it works like this


topjohnwu:
Thank you very much!

PR #97, opened 2017-02-27
fix syntax error in variable use

laggardkernel:
fix link copy


topjohnwu:
Thanks for pointing out, fixed.

PR #103, opened 2017-03-03
Rename bootimgtools to magiskboot

Yeradon:
Renaming is missing within build.sh

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -57,10 +57,10 @@ EXIT /B %ERRORLEVEL%
   COPY /Y libs\arm64-v8a\* zip_static\arm64
   COPY /Y libs\x86\* zip_static\x86
   COPY /Y libs\x86_64\* zip_static\x64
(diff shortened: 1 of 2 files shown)


Yeradon:
a839cb787e564df2c4781649426ea915cd2a4129

lexyan:
Hi Yeradon,
there is also reference to bootimgtools in the script directory, specially in  scripts/flash_script.sh

Yeradon:
Yeah saw that when i tried to flash. Gonna trie to find them all. I don't know much details about magisk's internal structures so its more of a trial and error for me. Gonna add the rest when i found all direct and indirect references.

Yeradon:
Ok a quick search showed that it's only used during un/installation process, which is what you expect. But when modifying the installer script, there is a 255 Error during installation. Which is not really helpfull,

Yeradon:
Fixed by topjohnwu in lässt commits

PR #139, opened 2017-03-24
Use busybox for bind mounts

steevp:
My BLU R1 HD's mount binary just segfaults with the bind option. This change allows Magisk to work correctly on my phone.

### DIFF
--- a/scripts/magic_mask.sh
+++ b/scripts/magic_mask.sh
@@ -57,7 +57,7 @@ unblock() {
 
 bind_mount() {
   if [ -e "$1" -a -e "$2" ]; then
-    mount -o bind "$1" "$2" || log_print "Mount Fail: $1 -> $2"
+    $BINPATH/busybox mount -o bind "$1" "$2" || log_print "Mount Fail: $1 -> $2"
   fi
 }
 
--- a/scripts/magic_mask.sh
+++ b/scripts/magic_mask.sh
@@ -57,7 +57,7 @@ unblock() {
 
 bind_mount() {
   if [ -e "$1" -a -e "$2" ]; then
-    $BINPATH/busybox mount -o bind "$1" "$2" || log_print "Mount Fail: $1 -> $2"
+    mount -o bind "$1" "$2" || $BINPATH/busybox mount -o bind "$1" "$2" || log_print "Mount Fail: $1 -> $2"
   fi
 }
 


DeicPro:
in some devices busybox's mount may fail... better use built-it mount and if fail use busybox one and if fail equally the regular return.

steevp:
Didn't think of that, thanks.

topjohnwu:
All commands running in the magic_mount scripts should already be running with busybox by default (the PATH variable is changed)
May you double check the issue?

steevp:
Uninstalled and reinstalled 11.6. Here is the log:
 https://hastebin.com/bepulitoku.md

Then I added some additional log printing:
```
bind_mount() {
  if [ -e "$1" -a -e "$2" ]; then
    log_print "PATH: $PATH"
    log_print "mount: $(which mount)"
    [ -e "$TOOLPATH/mount" ] || log_print "$TOOLPATH/mount not found"
    mount -o bind "$1" "$2" || log_print "Mount Fail: $1 -> $2"
  fi
}
```

Here is the log from that:
```
* Enabling BusyBox
PATH: /dev/busybox:/dev/busybox:/sbin:/vendor/bin:/system/sbin:/system/bin:/system/xbin
mount: /system/bin/mount
/dev/busybox/mount not found
Mount Fail: /dev/busybox -> /system/xbin
```

So it seems that, though $TOOLPATH is in the path, the $TOOLPATH/mount link hasn't been created yet, so it uses /system/bin/mount instead.

steevp:
OK I think I figured it out.

The symlinks exist but they point to /system/xbin/busybox, which might not exist before busybox is mounted. That's why the busybox bind_mount was using /system/bin/mount.

steevp:
OK I think I figured it out.

The symlinks exist, but they point to /system/xbin/busybox, which might not exist until after busybox is mounted. That's why the busybox bind mount was using /system/bin/mount.

Also looks like the "cp -afc" command in mount_busybox (and possibly other places) will use system cp if the user doesn't have a busybox binary in /system/xbin

topjohnwu:
This is really weird, because they should point to /data/magisk ($BINPATH), not /system/xbin

topjohnwu:
Try uninstall with uninstaller, then install Magisk again. Maybe something went wrong under /data/magisk

steevp:
Ah I see what happened now - I had an old busybox that didn't uninstall completely.

So when the magic_mask.sh was doing:
` cp -afc /system/xbin/. $TOOLPATH`
It would copy the old symlinks over. 

Sorry about that, I'll close this now.

Thanks.

PR #142, opened 2017-03-25
proc_monitor: Support older & newer kernels am_proc_start format

thassan:


topjohnwu:
Very nice! Thank you very much!

TheComputerGuy96:
Causes high CPU usage

PR #145, opened 2017-03-25
Fixes for A/B devices like Google Pixel

goodwin:


GabrielTK:
+1

JonnyTech:
@goodwin should all instances of `/system` be replaced with `$SYSTEM` in [Magisk/scripts/flash_script.sh](https://github.com/topjohnwu/Magisk/blob/master/scripts/flash_script.sh) especially since `SYSTEM=/system` is defined at the start?

ChrisM-Rogers:
I get an error unpacking the boot image on a stock device. I'll try to investigate later, currently in work so I can't.

goodwin:
@JonnyTech 
Maybe i missed some parts. If yes - sure, all /system should be replaced by variable. Main point is that on Pixel overall system structure is different as on other devices (and i suppose this will be same on all new A/B devices that will be shipped with Nougat on beginning)

ChrisM-Rogers:
I believe I've got it working on stock, to be confirmed after I've got Magisk Manager working. I'll pull in @topjohnwu's new commits, rebase and push it up. I guess the best way to solve this would be to send a PR to your fork and then you can add it to this PR maybe?

Probably be a good idea to get a few people to test it as well on your ROM, a couple extra stock testers and for good measure, other devices to ensure I/we haven't broken compatibility with other currently supported devices.

I can't get the Magisk Manager app to build on my laptop at the moment, so I'll have to wait till I get home to do anymore work. Currently, the PlayStore version shows my version stub in Magisk Manager, but Magisk Hide won't work yet and thus fails Safty Net, this is only when using Chainfire's SuperSU though, since I'd imagine Magisk Manager doesn't yet know where to find it's own binaries on multi-slot devices, so I hope to fix this for stock users too. :)

goodwin:
@ChrisM-Rogers 
You can send me PR and i will be happy to check your changes. And test if Magisk will keep working on RR with your changes (we should get it working everywhere ideally)

ChrisM-Rogers:
@goodwin Much agreed! That's the idea :) I haven't really added much at all to be honest, just conditions mostly.

crivchri:
Any news for A/B devices?

mleinart:
FWIW this tested successfully for me on a stock Google Fi Pixel XL running 7.1.2 with the May security update with the following caveat:

Install of this build *only* works if TWRP is flashed to recovery. If one boots TWRP via `fastboot boot` (as many guides recommend), the kernel cmdline has no reference to slots and so the new code in this patch doesn't detect this as an AB device.

IMO this solution is good enough if that caveat is put in the release notes.

An alternate may be to look in /dev/block/bootdevice/by-name/ 
    
    # ls -la /dev/block/bootdevice/by-name/boot
lrwxrwxrwx    1 root         root                36 Mar 20  1970 /dev/block/bootdevice/by-name/boot -> /dev/block/bootdevice/by-name/boot_b

(slot b is active on my device). Though I have no idea how reliable that is across devices

ChrisM-Rogers:
@mleinart that sounds promising. When I had a look around, I was using `dev/block/bootdevice/by-name/boot` as well. Seems like a viable option.

goodwin:
WIll close this request in next days. Have already working snapshot of v13 on my XL - so will prepare new pull request, targeted to upcoming release

ChrisM-Rogers:
@goodwin awesome! Is that on your ROM or on stock?

goodwin:
I'm testing everything on Resurrection Remix at the moment

ChrisM-Rogers:
Awesome, if you've got it pushed to your repository I'll take it for a spin on stock :)

ChrisM-Rogers:
@mleinart Can you double check to ensure you are running the May security patches? 

On this version, both root and TWRP fail to work for me resulting in a boot loop every time.

YHNdnzj:
@ChrisM-Rogers Flash [https://forum.xda-developers.com/android/software-hacking/signing-boot-images-android-verified-t3600606](url)

ChrisM-Rogers:
Ah thank you @YHNdnzj I wasn't aware that boot images had to be signed now.

goodwin:
@topjohnwu 
Hi man. I have working solution for latest sources of Magisk, that are working really fine on Pixel family. Are you interested at all? I'm trying to reach you on XDA, but no response on all my PMs

PR #146, opened 2017-03-26
list_monitor: only update_list() when file is actually written to

thassan:
this fixes the issue where magisk log gets flooded (over time) with messages about list being updated when its not


thassan:
for some reason for me with kernel 4.3.113 update_list() was called spuriously, causing flooding of the log over time. not sure if inotify blocks for a limited time period.

Issue #99, opened 2017-02-28
Magisk 11.1 SafetyNet not passed


Artrix:
I have Xiaomi Mi5 with miui 8.1.4.0, on Magisk v9 all work and safetynet is passed, on Magisk 11.1 cts profile mismatch. And app allegro (app for shopping) see root, app is added to hide in Magisk. 

Log from app: http://pastebin.com/C16H9H3K

amarpreetosahan:
I confirmed that issue

andrewleech:
I've been investigating this issue and discussing some workarounds here:
https://forum.xda-developers.com/mi-5/how-to/mi-5-passing-safetynet-unlocked-t3557845/post71119896#post71119896

Basically for any MIUI based roms I've tested safetynet passes with magisk v9 and phh-superuser-magisk-r266-2 as long as your build fingerprint matches a cts passing (stable) version (I've got a magisk module which will set this for you). I've done this on a few roms developer and a number of realeases of xiaomi.eu roms.

magisk v11.1 release and beta v12 fail both cts and basic integrity. 

I've got twrp recovery install log and logcat's of running magisk manager safetynet check attached here for both v9 and v12beta
the v12beta logcat also includes trying to run safetynet helper as well.

Before each test the boot image was reflashed in twrp to the original image for my rom.

My module to set props relating to safetnet passing, this was installed and enabled for all tests below:
[magisk-safetynet-props.zip](https://github.com/topjohnwu/Magisk/files/846225/magisk-safetynet-props.zip)

v9 testing
[recovery_v9_phh_install.txt](https://github.com/topjohnwu/Magisk/files/846229/recovery_v9_phh_install.txt)
[logcat_v9_safetynet_check.txt](https://github.com/topjohnwu/Magisk/files/846226/logcat_v9_safetynet_check.txt)
[magisk_error_log_v9.txt](https://github.com/topjohnwu/Magisk/files/846231/magisk_error_log_v9.txt)

v12 beta testing
[recovery_v12_170312_install.txt](https://github.com/topjohnwu/Magisk/files/846232/recovery_v12_170312_install.txt)
[logcat_v12_safetynet_check.txt](https://github.com/topjohnwu/Magisk/files/846233/logcat_v12_safetynet_check.txt)
[magisk_error_v12.txt](https://github.com/topjohnwu/Magisk/files/846234/magisk_error_v12.txt)

I'd be happy to help further in diagnosing the regression in v11 / v12
Cheers.

PR #149, opened 2017-03-28
Update magiskhide enable script to run daemon in su subshell

andrewleech:
I'm using magisk / magisk hide on a xiaomi Mi5 miui based rom. 

It hasn't worked properly for me since v9 + phh however, as discussed in https://github.com/topjohnwu/Magisk/issues/99#issuecomment-286910905

This issue report however got me on a track to fixing it: https://github.com/topjohnwu/MagiskManager/issues/112

Running the magiskhide enable script from adb su shell worked fine with recent releases!

Interestingly when enable is run through magisk manager, the magiskhide daemon appears as such in ps:
```
gemini:/ # ps | grep magiskhide
root      7815  1     10536  600   hrtimer_na 7f94df30a0 S /magisk/.core/bin/magiskhide
root      7816  7815  7456   372    pipe_wait 7f94df31a8 S /magisk/.core/bin/magiskhide
```

but when run from adb su shell it shows up as: 
```
root      7976  1     10536  528    pipe_wait 7f798ce1a8 S /magisk/.core/bin/magiskhide
root      7977  7976  7456   372    pipe_wait 7f798ce1a8 S /magisk/.core/bin/magiskhide
```
the WCHAN column always shows the parent process stuck waiting at a different point in the kernel.
Not sure if this is a permissions thing related to starting from an app rather than shell (even with su) related to https://doridori.github.io/Android-Security-welcome-to-shell/#sthash.LS0yvVWX.dpbs ?

regardless of the reason, the attached change appears to completely fix it on my device; just simply running magisk daemon within a separate su process 

The manager app is certainly already running the enable/disable under su, the fact that the disable script still works means it has permission to kill the daemon, but there's clearly something different about running it inside a separate su shell because it works every time for me.

Not understanding the root cause, I don't know if this is the "correct" way to fix the issue, but at least it's simple.

### DIFF
--- a/zip_static/common/magiskhide/enable
+++ b/zip_static/common/magiskhide/enable
@@ -72,4 +72,4 @@ while read PROCESS; do
 done < $MODDIR/hidelist
 
 log_print "Starting MagiskHide daemon"
-$BINPATH/magiskhide --daemon
+su -c $BINPATH/magiskhide --daemon


chhapil:
thanks for the fix, i'll try this at my end. Does this fix has issue Magisk Manager detecting the root?

chhapil:
Magisk Manager has issue detecting root with this fix, it doesnt detect device as rooted

pi-anl:
I didn't have any issue with Magisk Manager detection root, it always showed root green for me... until I just switched back to Magisk Manager and it was showing no root. 
I killed Magisk Manager and started it again (no reboot, just restart app) and it's all green again.
I think that's a completely different issue, I've seen this in the past and I wouldn't expect this change to affect it. This is only run at startup and when you toggle the magisk hide enabled setting.

For me, without this fix, magisk hide will not work (even though it says it's enabled). 
safetynet will not pass without the fix and root is never hidden.
It doesn't directly have anything to do with root.

andrewleech:
I didn't have any issue with Magisk Manager detection root, it always showed root green for me... until I just switched back to Magisk Manager and it was showing no root.
I killed Magisk Manager and started it again (no reboot, just restart app) and it's all green again.
I think that's a completely different issue, I've seen this in the past and I wouldn't expect this change to affect it. This is only run at startup and when you toggle the magisk hide enabled setting.

For me, without this fix, magisk hide will not work (even though it says it's enabled).
safetynet will not pass without the fix and root is never hidden.
It doesn't directly have anything to do with root.

chhapil:
thanks for your comments, working well at my end as well! 
Cheers!
Magisk Manager detects root and hide also working

chhapil:
MagiskManager works well after first boot after flashing but after subsequent reboots it does not recognise root.

andrewleech:
Again, I think you're talking about a completely different issue.
It works perfectly for me (over many reboots) and always shows root. 

Have you looked at the changes for this pull request? 
I doubt there's any way what I'm doing could affect the detection of root. 

Perhaps you should open a new issue report.

andrewleech:
Thanks

PR #187, opened 2017-05-01
Update cpio.c for Mate 9

Ante0:
Updated to work with Huawei Mate 9 (and probably future Huawei devices). All it does is change fileencryptioninline to encryptable, as they decided to use file based encryption.


topjohnwu:
Thanks, added in 20903784a47286b45b40037c6584e9deb4c1a95e

PR #224, opened 2017-06-03
Include uninstaller in a full Magisk build

TheComputerGuy96:
@topjohnwu


topjohnwu:
Added in 17684ed8a8738c11543a109616d3f75aa9418d97, thanks :)

PR #268, opened 2017-07-11
Fix core-only mode

goodwin:


TheComputerGuy96:
@goodwin Please remove Pixel support commit, if you want @topjohnwu to merge it

PR #277, opened 2017-07-12
Core-only mode, pixel compatibility

goodwin:
1. Fix for core only mode
2. Bringing Pixel compatibility - works fine currently on my device. Shouldn't break compatibility with all other devices


TheComputerGuy96:
@goodwin Why can't you make the magisk binary unified for A/B (Pixel) devices?

goodwin:
it's unified - after boot. but we need magisk_utils and magiskpolicy compiled as static - in other casse they will not work under TWRP on pixel. Something related to linker most probably, but easiest way was to make separate binaries.
So in short - after boot we are using normal "magisk" binary as others. During installation in TWRP we are using two extra binaries, compiled as static.

TheComputerGuy96:
@goodwin You should cleanup this PR for Magisk v13.1

goodwin:
@TheComputerGuy96
If you have some doubts or notes regarding any part of this PR - please write more detailed comment. Otherwise - it is already based on latest source code, so i dont see what i should change. And there are no merge conflicts

JpegXguy:
image

goodwin:
When i was writing that comment - there was no conflicts ;) sure, time is going, and during past 5 days there are fresh updates that are bringing merge conflicts, this is normal

JpegXguy:
That is so

The-MAZZTer:
Hey, your fork doesn't build, commit 3865acceb0af7417679235badd1649fe442a5fc4 breaks the build process with: 

* Building Magisk binaries

Android NDK: Trying to define local module 'sepol' in jni/selinux/libsepol/Android.mk.
Android NDK: But this module was already defined by jni/selinux/libsepol/Android.mk.
C:/Users/mzzt/AppData/Local/Android/sdk/ndk-bundle/build//../build/core/build-module.mk:34: *** Android NDK: Aborting.    .  Stop.

Build Magisk binary failed!

I think the change you made to Android.mk did it. Previous commit builds fine.

goodwin:
@The-MAZZTer 
Looks like i have missed some change. Thank you that you found it - i'll fix in next days

tycho:
@goodwin If/when you get your changes applied on v13.x I'd be happy to test it on my Pixel (sailfish).

The-MAZZTer:
It looks like @goodwin has not been keeping his github up to date and he does not post builds there, here is where he posts builds:

https://www.androidfilehost.com/?w=files&flid=183895

The latest is  Magisk-v13.3-Pixel.zip but some people on reddit are saying it is not stable and to use the earlier v13.0 new10. I have not tried either yet.

goodwin:
Updated PR with latest changes

tycho:
@goodwin Tested Magisk-v13.3-Pixel-b4.zip from https://www.androidfilehost.com/?w=files&flid=183895 and it seems to forget it has root after a while. I boot up, run a few apps requiring SU (e.g. Root Explorer) and they behave fine. Then when I try later I find that I've "lost" root until I reboot the device. "b1" seems to behave with respect to keeping root, but apparently fails SafetyNet.

tycho:
"b3" also forgets root after a couple of su's

goodwin:
@tycho discussion thread is here https://forum.xda-developers.com/apps/magisk/unofficial-google-pixel-family-support-t3639262
write which rom/kernel are you using, and also provide full logcat in between when su is working and it stops working

goodwin:
Not needed anymore - Pixel has official support

GabrielTK:
@goodwin Official Support?

PR #293, opened 2017-07-18
Explicitely add path to magisk-binary if called in recovery.

jenslody:
Needed for new module-template, tha fixes access-denied when running execv from below /data.


jenslody:
See also https://github.com/topjohnwu/magisk-module-template/pull/22.

PR #364, opened 2017-12-28
Code style improvements to build.py

NeverMine17:
- Option `-h` or `--help` now brings up help menu.
  - Moved all imports to the top.
  - Renamed function options that shadowed other things.
  - Removed unused arguments.
  - Other stuff.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,405 +1,431 @@
 #!/usr/bin/env python3
-import sys
+import argparse
+import base64
+import datetime
+import errno
+import lzma
+import multiprocessing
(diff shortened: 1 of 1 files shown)


topjohnwu:
Thanks, but the imports are split in purpose, because I want things to fail quickly before importing tons of other modules (which takes time to load).
Also, I personally prefer tabs.
Some other stuffs are nice, but since there are too many changes, I'll close it for now.
If you want to update your PR, it's welcome to update and reopen :)

PR #368, opened 2018-01-17
Superuser request prompt not saving

uvera:
So recently I've encountered a bug where Magisk is unable to save superuser request prompts even if set on "Forever". 
This results in apps spamming SU requests. This seems to be a database issue. 
Here are the logs: 
https://haste.blamealexcruz.com/mT4dXgC6mg.scala

Video showcase here:
https://drive.google.com/open?id=1G0m6y8bWE1C6_zmB0Vo5YLQx7Rg9sGPs

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -23,7 +23,7 @@
 
 
 ## License
-
+ 
 ```
 Magisk, including all git submodules are free software:
 you can redistribute it and/or modify it under the terms of the 


pogo1975:
from my experience when i installl 15.3 clean it will not create database, when i updating from 14 then it will see old permisions but do not add new...

PR #370, opened 2018-02-05
utils/file.c: NULL terminate all files read into memory

chenxiaolong:
Some functions, like `patch_init_rc()`, treat buffers read into memory
as a string instead of a byte buffer. Since the buffers weren't
NULL-terminated, this resulted in out-of-bounds reads and caused crashes
in certain conditions.

THis commit updates fd_full_read() to always NULL-terminate the buffers
so that they can be treated as strings when working with text files.

### DIFF
--- a/native/jni/utils/file.c
+++ b/native/jni/utils/file.c
@@ -401,8 +401,9 @@ int mmap_rw(const char *filename, void **buf, size_t *size) {
 void fd_full_read(int fd, void **buf, size_t *size) {
 	*size = lseek(fd, 0, SEEK_END);
 	lseek(fd, 0, SEEK_SET);
-	*buf = xmalloc(*size);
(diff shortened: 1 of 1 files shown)


chenxiaolong:
Example of where this caused a crash on my device before:

https://github.com/topjohnwu/Magisk/blob/599ae952519138a697b96c369da218f2cf876524/native/jni/core/magiskinit.c#L178-L179

and

https://github.com/topjohnwu/Magisk/blob/599ae952519138a697b96c369da218f2cf876524/native/jni/utils/pattern.c#L32-L38

TheComputerGuy96:
@chenxiaolong How about Magisk + DualBootPatcher compatibility? :thinking:

chenxiaolong:
@TheComputerGuy96 Coming soon :)

PR #373, opened 2018-02-12
Use more generic way to mount vendor

TheScarastic:


TheScarastic:
i agree, but i dont think there is better way to handle it

auanasgheps:
Guys, I don't know if I am exact, but this can/should be addressed in the TWRP.
Actually this has been done for Xiaomi Mi5, an [updated TWRP](https://forum.xda-developers.com/showpost.php?p=70813439&postcount=2) has been released to use /cust as /vendor.

Probably Magisk should not mess around with this.

TheScarastic:
uhh? that is not even related with full_treble enabled selinux is split and while trying to patch it init crashes

Jerry981028:
Perhaps we should verify the phone model first.

feherneoh:
Just an idea, but cannot magiskinit just have a look at the fstab from the original ramdisk, and set up the blockdevice for vendor based on that?

TheScarastic:
@Dinolek pretty sure it wont work that way atleast for magisk since it doesn't looks in /dev/block instead it gets partition number from /sys/dev/block  https://github.com/topjohnwu/Magisk/blob/master/native/jni/core/magiskinit.c#L118

Dinolek:
@TheScarastic /sys/dev/block are by by block numbers, not by dev name, at least in my device as magisk search.
I tried it and flashing magisk work with this symlink.

feherneoh:
@Dinolek and what about booting?

Dinolek:
@feherneoh @TheScarastic Yes, sorry, it doesn't boot with that symlink, I first thought that this is for flashing , not for booting

native/jni/core/magiskinit.c
```diff
@@ -428,11 +428,24 @@ int main(int argc, char *argv[]) {
...
 		mkdir("/sys", 0755);
 		xmount("sysfs", "/sys", "sysfs", 0, NULL);
 
+		// Check if device uses early-mount
+		if (access("/sys/firmware/devicetree/base/firmware/android/fstab/vendor/dev", R_OK) == 0) {
+			fd = open("/sys/firmware/devicetree/base/firmware/android/fstab/vendor/dev", O_RDONLY | O_CLOEXEC);
+			vendorpath[read(fd, vendorpath, sizeof(vendorpath))] = '\0';
+			strcpy(vendor, strrchr(vendorpath, '/') + 1);
+			close(fd);
+		} else
```
TheComputerGuy96: Where's the `}`?

native/jni/core/magiskinit.c
```diff
@@ -428,11 +428,24 @@ int main(int argc, char *argv[]) {
...
+			strcpy(vendor, "vendor");
+
```
TheComputerGuy96: And, where's the `{`?

dark-1: My bad , I mean "Single line Code" for `else`

native/jni/core/magiskinit.c
dark-1: @TheScarastic ,
So that's the reason you said LOL ?????

spidey:
@Jerry981028, unfortunately the binary you released is not working for me neither, on a Xiaomi Redmi Note 4X Snapdragon I'm using RedWolf recovery.

Jerry981028:
@spidey Need detailed information, please open an issue in the my repo.

topjohnwu:
I will find a way to configure a custom vendor partition without hardcoding for specific devices.
But before it's implemented, I'll still leave this PR open.

TheScarastic:
@topjohnwu can you explain what do you mean by  "hardcoding* since current code dynamically looks up for vendor partition through device-tree early mount sysfs and mounts it, if no sysfs is present it fallback to traditional "vendor" partition

sjayminsgithub:
Any update regarding this issue ??

topjohnwu:
Yes :)

PR #374, opened 2018-02-15
Fix detecting su in Qiwi

umerov1999:


skaforey:
Wow this looks like a simple fix, does it work? I suspect if it does, this would fix the same naive check that Microsoft InTune does.  For now I have been using tasker to run a script to rename /sbin/su to /sbin/mu when I need to open Outlook.

auanasgheps:
I am surprised that this path was not already unmounted... Well done.

skaforey:
I confirmed this worked by building magisk with this change, but once it's triggered, /sbin/su never gets relinked.....I think we need an additional change to relink

skaforey:
Furthermore, not sure how to actually mount it back since the binary points to /root/<randomname> that doesn't seem to be stored anywhere.  I tried symlinking back to /sbin/magisk but that doesn't work I guess.

skelletR:
Hi I have problem to detecting root the app hex Editor say.. Can not change shell, is your phone rooted? Stream closed

topjohnwu:
This is not a proper way to hide Magisk, the changes are irreversible

PR #375, opened 2018-02-22
Fixup support for Nook HD, add support for Acclaim

worstperson:
This fixes and updates support for the Nook HD (hummingbird) and HD+ (ovation), and adds support for the Nook Tablet (acclaim).

### DIFF
--- a/native/jni/magiskboot/bootimg.c
+++ b/native/jni/magiskboot/bootimg.c
@@ -124,11 +124,16 @@ int parse_img(const char *image, boot_img *boot) {
 				fprintf(stderr, "PXA_BOOT_HDR\n");
 				boot->hdr = malloc(sizeof(pxa_boot_img_hdr));
 				memcpy(boot->hdr, head, sizeof(pxa_boot_img_hdr));
-			} else if (memcmp(((boot_img_hdr*) head)->cmdline, NOOK_MAGIC, 12) == 0
-					|| memcmp(((boot_img_hdr*) head)->cmdline, NOOK_NEW_MAGIC, 26) == 0) {
-				boot->flags |= NOOK_FLAG;
-				fprintf(stderr, "NOOK_GREEN_LOADER\n");
-				head += NOOK_PRE_HEADER_SZ - 1;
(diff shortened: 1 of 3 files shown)


osm0sis:
I have a question regarding the Nook HD "new magic" and the Acclaim magic - are there different versions for recovery images? I know with the old "Green Loader" version there was also "Green Recovery".

worstperson:
Acclaim magic is uniform, Nook HD uses "Red Loader" and Nook HD+ uses "Green Loader". TWRP3 built by amaces uses "eMMC recovery.img+secondloader for hummingbird" and "eMMC recovery.img+secondloader for ovation", just like the ROMs.

osm0sis:
Thanks! I'll fix up support for AIK and AK2. :+1:

worstperson:
It may be used in some old roms, I'm not certain, it does not apply to Magisk though. hashcode/amaces Lollipop ROMs use "Green Loader" for both HD and HD+, while Marshmallow+ ROMs from amaces/Unlegacy-Android use the new magic.

airend:
> Does "Red Loader" need to be added to this PR then? I don't see any mentioning of it.

I think that it was only used in old recovery images, which need a slightly different loader.

worstperson:
"Red Loader" is currently being used in official TWRP for ovation

airend:
Yup, basically anything that uses the old LP device trees (our official TWRP images are stuck on those).

osm0sis:
Last question, you say Accliam magic is uniform, and I see it's "BauwksBoot", so does that mean that's the same even for recovery? Or is there "BauwksRecovery" ? Thanks!

osm0sis:
Looks like no, since I found TWRP still has "BauwksBoot" - so I think we're good: https://github.com/osm0sis/AnyKernel2/commit/c81b3d565843516661eaf54a74745f72ef3dfe3b 👍

worstperson:
Yes, all acclaim boot and recovery images I've found(that are still up) use "BauwksBoot" @ 0x30.

PR #376, opened 2018-02-23
Fix error parsing MTK boot.img

shakalaca:
Both should be copy & paste error:
1. boot->r_fmt should be re-checked instead of boot->k_fmt once MTK header was found in ramdisk.
2. ramdisk_size should be restored instead of kernel_size when uncompressed ramdisk was found.


Lobesitdoll:
Will give this a go, thank you

PR #379, opened 2018-03-11
SDCARDFS Workaround for Permission 13 Errors

Permission13:
This allows mounting and writing on /dev/block/loop for devices which suffer from the Permission 13 error which is NOT due to F2FS.
Reason is probably kernel or DM blocking accesses to /dev which ends up with I/O errors which results in mount errors or the kernel switching the filesystem of (loopback) mounted files to read-only.

One drawback:
After having rebooted to complete the installation of a module,
you have to reboot one more time before being able to install the next one (device busy when unlinking)

On my phone this workaround results in the need for the following SELinux rules:

> sepol_allow("kernel", "adb_data_file", "file", ALL); 
sepol_allow("kernel", "adb_data_file", "dir", ALL); 
sepol_allow("kernel", "system_data_file", "file", ALL); 
sepol_allow("kernel", "system_data_file", "dir", ALL);

Almost everything should be adaptable except SOURCE_TMP and TARGET_TMP definitions.


dark-1:
@Permission13 
What's with the commit by @Devying ? 😲 

Best regards 👍

Permission13:
The content is actually almost the same - just joined and cleaned and in one commit. Also had some problems locally with updating/merging submodules and did a clean "restart". You can compare with the previous one in the closed pull requests.

dark-1:
@Permission13 ,
my advice , Kindly use the same email as of @Permission13  in your git client & not @Devying 's email.
   
I can see commit made only by @Devying  & Not by @Permission13 .
So I am guessing Both are same.

Best regards 👍

Permission13:
Next time i definitely will - i just didnt expect anybody to use the same senseless non existing email. I am not using github normally. Sorry for the confusion

PR #381, opened 2018-03-21
boot->tail_size calc bugs

idhyt:
if head != boot->map_addr, boot->tail_size calc error!


topjohnwu:
Thanks for reporting the bug!

PR #384, opened 2018-05-05
Fix early mount on S9/S9+

msdx321:
Unlike other common OEMs, Samsung use uppper case partition name.

e.g. /dev/block/platform/11120000.ufs/by-name/SYSTEM

This will cause setup_block() fails to find a match partition.
Thus, we should use strcasecmp instead of strcmp.

Signed-off-by: msdx321 <msdx321@gmail.com>


topjohnwu:
Cool, thank you very much 💯

PR #386, opened 2018-05-15
Fix build.py issue when attempting to build all

altr0n:
When building all, no target member of the args object is ever instantiated, causing the script to fail when it enters the ```build_binary``` function. 

A check was added to add the ```target``` attribute and properly instantiate it.

Tested on Ubuntu 18.04


topjohnwu:
Already fixed, thanks!

PR #388, opened 2018-05-28
Added a check to ensure the patched boot image can fit in the boot partition

rocco8620:
Added a check in the install script to ensure the patched boot image can fit in the boot partition before attempting to write the image. If the size is too big the installation fails with a descriptive error message.

This missing check may result in broken boot images (due to image truncation when writing on boot partition).

On the device Honor 7 (PLK-L01 16GB EU version) this problem prevented correct installation.

### DIFF
--- a/scripts/flash_script.sh
+++ b/scripts/flash_script.sh
@@ -124,6 +124,17 @@ cd $MAGISKBIN
 # Source the boot patcher
 . $COMMONDIR/boot_patch.sh "$BOOTIMAGE"
 
+ui_print "- Checking patched image size against partition size..."
+
+exact_file_size new-boot.img
+exact_partition_size "$BOOTIMAGE"
+ui_print "- New image size: $EXACT_FILESIZE, boot partition size: $PARTITION_SIZE"
+
+if [ "$EXACT_FILESIZE" -gt "$PARTITION_SIZE" ]; then
+  ui_print "The patched image is too big!"
+  abort "You need to patch the boot image from the MagiskManager apk, reduce it size manually and flash it manually." 
+fi  
+
 flash_boot_image new-boot.img "$BOOTIMAGE"
 rm -f new-boot.img
 
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -342,6 +342,13 @@ mktouch() {
   chmod 644 $1
 }
 
+exact_file_size() {
+  EXACT_FILESIZE=$(stat -c%s "$1")
+}
+exact_partition_size() {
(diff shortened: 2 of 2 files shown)


sjayminsgithub:
@rocco8620, what does "You need to patch the boot image from the MagiskManager apk, reduce it size manually and flash it manually." mean??

rocco8620:
It's a descriptive error message. After the installation fails because of insufficient space in the boot partitions, this error message tells the user what to do install magisk successfully (patch a valid boot image with the magiskManagre application, remove some not necessary content (eg. not used boot logos) from the image, repack the boot image and flash it.

sjayminsgithub:
@rocco8620, OK I understand it now. And I'd suggest putting some extra information on how to do that, and maybe links to some possible tools for that.

rocco8620:
It's an idea but there isn't a lot of space where the error message is shown (in a terminal-like window, on TWRP recovery mod or equivalent) so no copy/paste of informations is possible. Maybe we can add a link to a wiki page (maybe here on GitHub) where the problem/solution is explained in details.

sjayminsgithub:
Well, that'd feel good, too. Wiki in this repo doesn't seem to be enabled, is it? It can be a one for all troubleshooting guide for Magisk. @topjohnwu

rocco8620:
Yes I think it's a good idea

rocco8620:
I have tested version 16 of magisk installation script. During the installation using TWRP mod, I saw the messages about the highcomp mode being enabled but this didn't prevent the corruption of the boot image. 

I don't actually know how the highcomp mode works, so maybe you require more detailed informations about that. I can try to attach the installation log if it's needed. My device is an Honor 7 (PLK-L01 16GB EU version)

scripts/flash_script.sh
```diff
@@ -124,6 +124,17 @@ cd $MAGISKBIN
...
 
+ui_print "- Checking patched image size against partition size..."
+
+exact_file_size new-boot.img
+exact_partition_size "$BOOTIMAGE"
+ui_print "- New image size: $EXACT_FILESIZE, boot partition size: $PARTITION_SIZE"
+
+if [ "$EXACT_FILESIZE" -gt "$PARTITION_SIZE" ]; then
+  ui_print "The patched image is too big!"
+  abort "You need to patch the boot image from the MagiskManager apk, reduce it size manually and flash it manually."
```
TheComputerGuy96: s/it/its/

zaaraameera2014:
Orait

zaaraameera2014:
Tukarlah sial

PR #391, opened 2018-06-16
Prevent setting zero over than bound in magiskinit.c

npes87184:
Hi, @topjohnwu 

The &cmd will return a pointer which point to a pointer of cmdline.
It is a memory address which is usually 8 bytes in 64 bits machine.

However, the struct cmdline is 4 bytes. This will cause setting zero
beyond the bound.

Below is a simple example to show the differentiation:
```
struct cmdline {
        char skip_initramfs;
        char slot[3];
};

static void parse_cmdline(struct cmdline *cmd)
{
        printf("%lu\n", sizeof(*cmd)); /* 4 */
        printf("%lu\n", sizeof(&cmd)); /* 8 */
}

int main()
{
        struct cmdline cmd;
        parse_cmdline(&cmd);
        return 0;
}
```
This patch prevents this.

Thanks.

Signed-off-by: npes87184 <npes87184@gmail.com>

### DIFF
--- a/native/jni/core/magiskinit.c
+++ b/native/jni/core/magiskinit.c
@@ -75,7 +75,7 @@ struct device {
 
 static void parse_cmdline(struct cmdline *cmd) {
 	// cleanup
-	memset(cmd, 0, sizeof(&cmd));
+	memset(cmd, 0, sizeof(*cmd));
 
 	char cmdline[4096];
 	mkdir("/proc", 0555);


sjayminsgithub:
It seems like this patch is already included in the main sources, could you check that??

npes87184:
Hi, @sjayminsgithub 

As I check the main line, the [magiskinit.c](https://github.com/topjohnwu/Magisk/blob/master/native/jni/core/magiskinit.c) still uses `&` instead of `*`.

Thank you.

sjayminsgithub:
My bad, I mistook your repo as the topjohnwu's.

npes87184:
Hi, @sjayminsgithub 

That's ok.

Thank you for your prompt reply.

topjohnwu:
Thanks, stupid typo lol

PR #392, opened 2018-06-19
Fix survival script for AVB 1.0 signed boot image

xingrz:
The survival script didn't check if the boot.img was signed. This will cause failing boot on devices required signed boot images.

This patch checks AVB 1.0 in the survival script with `$BOOTSIGNER`, and assign the `$BOOTSIGNED`, in the same way the flash script does.

### DIFF
--- a/scripts/addon.d.sh
+++ b/scripts/addon.d.sh
@@ -13,7 +13,9 @@
 main() {
   # Magisk binaries
   MAGISKBIN=/data/adb/magisk
-  APK=/data/magisk.apk
+  APK=/data/adb/magisk.apk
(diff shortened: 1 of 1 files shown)


topjohnwu:
I added some additional checks to the path of the APK.
Thanks for the suggestion 👍

xingrz:
Why? In https://github.com/topjohnwu/Magisk/blob/master/native/jni/core/bootstages.c#L843, during the first boot after flashing Magisk, `/data/adb/magisk/magisk.apk` was moved to `/data/magisk.apk`. So I think the APK will always be there.

Did you verified wildcard works here?

PR #395, opened 2018-06-24
magiskhide: close fd when error out

npes87184:
Hi, @topjohnwu 

In original implementation, the file descriptor will remain open when
error out. In addition, this patch also shrinks the exit point to single
exit point to prevent some leaking happens again.

Thanks.

Signed-off-by: npes87184 <npes87184@gmail.com>


npes87184:
Hi, @topjohnwu 

Thats right. However, Closing file descriptors when you are done using them makes your code more reusable and easier to extend. Conceptually, when the process dies, it is your responsibility to free allocated memory, close file descriptors, etc. Of course, every sane OS (and even Windows) will clean up after your process, but that's not something to expect.

Thanks

PR #405, opened 2018-07-20
Respect filesystem type when mounting mirrors

nonnymoose:
This change reads the filesystem type from /proc/mounts as well and passes that to mount when mounting the mirrors.

This _fixes_ the issue in which devices with ubifs filesystems throw the error
```
mount /dev/ubi0_0->/sbin/.core/mirror/system failed with 15: Block device required
```
about which I think I complained at you a few months ago.

It should also improve compatibility with other strange filesystem types, but I have no way of testing that.

### DIFF
--- a/native/jni/core/bootstages.c
+++ b/native/jni/core/bootstages.c
@@ -22,7 +22,7 @@
 #include "resetprop.h"
 #include "magiskpolicy.h"
 
-static char *buf, *buf2;
+static char *buf, *buf2, *buf3;
 static struct vector module_list;
 
 extern char **environ;
@@ -500,6 +500,7 @@ void startup() {
 		// Allocate buffer
 		buf = xmalloc(PATH_MAX);
 		buf2 = xmalloc(PATH_MAX);
+		buf3 = xmalloc(PATH_MAX); // used for detecting fs type
 
 		simple_mount("/system");
 		simple_mount("/vendor");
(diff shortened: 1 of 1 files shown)


native/jni/core/bootstages.c
```diff
@@ -802,6 +804,7 @@ void late_start(int client) {
 	// Allocate buffer
 	if (buf == NULL) buf = xmalloc(PATH_MAX);
 	if (buf2 == NULL) buf2 = xmalloc(PATH_MAX);
+	if (buf3 == NULL) buf3 = xmalloc(PATH_MAX);
```
nonnymoose: There should probably be a saner default for allocating the filesystem buffer, but I'm not sure what it is.

npes87184:
Hi, @nonnymoose 

buf3 will memory leak.
Remember to free buf3 like buf and buf2 do in [Line 841](https://github.com/topjohnwu/Magisk/blob/master/native/jni/core/bootstages.c#L841).

Thanks.

nonnymoose:
@npes87184 Thanks for your suggestions! I have fixed the memory leak, but bricked my device in the process :(, so it should be an hour (or three) before I can test it and commit.

nonnymoose:
@npes87184 OK, I have done that. Do you have an idea of what the maximum fs_type size should be?
The longest filesystem supported by my Ubuntu machine is 11 characters.

nonnymoose:
@npes87184 I agree with your suggestions. The maximum length filesystem type I could find was still 11, but it can't hurt to be _really, really_ careful, so 64 seems good to me.

npes87184:
Hi, @nonnymoose 

Well, the static allocated I meant is `static char buf_fs_type[64] = {0}`.
But, I don't know whether use static allocated is a good idea or not. This question is left to @topjohnwu.

Regards!

nonnymoose:
Sorry about that. I'm not used to c and tend to get terminology wrong.

nonnymoose:
@topjohnwu Thanks so much!

PR #447, opened 2018-08-05
Avoid crashes when opening links on device without Internet browser

tonymanou:
If there is no available application to handle the intent, `context.startActivity(intent)` will crash... (It's the case when trying to open a link while on an Android device where there is no Internet browser, for instance.)
Here we try to resolve the intent first before showing a toast to warn the user the link cannot be opened.


app/src/full/res/values/strings.xml
```diff
@@ -96,6 +96,7 @@
     <string name="hide_manager_toast">Hiding Magisk Manager…</string>
     <string name="hide_manager_toast2">This might take a while…</string>
     <string name="hide_manager_fail_toast">Hide Magisk Manager failed…</string>
+    <string name="open_link_failed_toast">No application found to open the link…</string>
```
tonymanou: @topjohnwu fell free to change the string's key if it does not suit you

PR #448, opened 2018-08-05
Fix crash on boot on Android pre-O

tonymanou:
When using JobIntentService on pre-O, we must have `android.permission.WAKE_LOCK` permission because a wake-lock is used internally.


topjohnwu:
I moved the permission to the full variant, because the stub does not need it

PR #469, opened 2018-08-12
Handle time-consuming boot scrips to prevent breaking other modules

sjayminsgithub:
Hello,
Currently, if some boot script is time-consuming, then it will break other Modules to be properly Magic Mounted. So, if it's the post-fs-data, don't let the script run for more than a sec, otherwise we go the blocking way. Also, for module scripts, we kill the process to prevent it from conflicting when Magic Mounting is happening, but for general scripts, we ignore it.
Regards.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,7 +1,8 @@
 apply plugin: 'com.android.application'
 
 def configProps = new Properties()
-configProps.load(new FileInputStream(rootProject.file('config.prop')))
+def configPath = project.hasProperty('configPath') ? project.configPath : rootProject.file('config.prop')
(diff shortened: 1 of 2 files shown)


sjayminsgithub:
@topjohnwu, please increase the time from 1 second to 2 seconds in wait_for_pid().

sjayminsgithub:
Thanks for the consideration!
I totally agree with what you said. I preferred the per script idea because many modules do some post-fs-data stuff which is limited to them only. I'm currently imagining what your specific idea should be! I strongly wish Magisk handled all the modules stuff at an advanced level to make sure a little script kiddie can not destroy users' devices.
Best regards.

PR #487, opened 2018-08-20
Disable defex_lsm on Oreo A530F

breakersvd:
Added hexpatch to disable defex security

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -150,6 +150,9 @@ if [ -f kernel ]; then
   ./magiskboot --hexpatch kernel \
   736B69705F696E697472616D6673 \
   77616E745F696E697472616D6673
+  
+  # disable defex_lsm security
+  ./magiskboot --hexpatch kernel b91f040071802f005460de b91f040071812f005460de
 fi
 
 ##########################################################################################


topjohnwu:
This is very cool! May you elaborate how you get the patch?

breakersvd:
@topjohnwu Sure,
First i downloaded the original A530F sources from SAMSUNG OPEN SOURCE RELEASE CENTER
and after extracting the source from `kernel.tar.gz` we use a toolchain (i prefer the one specified in README_kernel txt) so i used android toolchain from google git server.

now we set needed variables (explained in README_kernel txt) then run make command.
after first success build i edited this file [scripts/mkcompile_h] to prevent update the time and date that printed in the kernel (to make the differences comparing smaller as much as possible)
then rebuild the sources.

now we will go to the needed sources to patch and study them extensively to know what we need to patch (in out situation [security/defex_lsm])

and here the problem we face the device is `killed process if pid was 0 || 1`
so i opened `defex_procs.c` and edited the line `if (!p || p->pid == 1 || !p->mm)` to `if (!p || p->pid != 1 || !p->mm)` (very important: the edits must not be delete or add; only replace).

now we run make command to build the sources and after finish we compare the both stock built image with new patched built image (because of disabling updating kernel build info [data & time] the result of comparing will be small; only what we edited).

and in our case the result was b91f040071802f005460de b91f040071812f005460de.
we can see the difference only (80 > 81) but we must add the right and left characters to prevent hexpatcher changing other lines.

that is it!

topjohnwu:
I'm going through the same route as you do, but missed the date/time stuff so it is very difficult for me to diff the binaries 😅
Thanks for doing the work!

breakersvd:
@topjohnwu 
i'm glad to do this, anyway for the date, time & info stuff
during compiling process you will see lines started with `UPD`
these lines is responsible for updating the info so do the next to disable them:

it is simple in the file `mkcompile_h`:

> if [ -r $TARGET ] && \
>       grep -v 'UTS_VERSION' $TARGET > .tmpver.1 && \
>       grep -v 'UTS_VERSION' .tmpcompile > .tmpver.2 && \
>       cmp -s .tmpver.1 .tmpver.2; then
>    rm -f .tmpcompile
> else
>    vecho "  UPD     $TARGET"
>    mv -f .tmpcompile $TARGET
> fi
> rm -f .tmpver.1 .tmpver.2

make it like this to disable the updating info:

> if [ -r $TARGET ] && \
>       grep -v 'UTS_VERSION' $TARGET > .tmpver.1 && \
>       grep -v 'UTS_VERSION' .tmpcompile > .tmpver.2 && \
>       cmp -s .tmpver.1 .tmpver.2; then
>    rm -f .tmpcompile
> #else
>    #vecho "  UPD     $TARGET"
>    #mv -f .tmpcompile $TARGET
> fi
> rm -f .tmpver.1 .tmpver.2

PR #488, opened 2018-08-20
fix build script

ahxxm:
- resource: `util_functions.sh` is needed to build apk
- filename: `zip_main` needs `app-{buildType}.apk`


topjohnwu:
Fixed: 89a501a3af4fef4c70fc861ab0888e01d5394ccf

PR #589, opened 2018-09-09
Rollback values-de translation Update -> Aktualisierung 

GuepardoApps:
As we use "Aktualisierung/en" for nearly all translations of "Update/s", we should keep them also in those 2  strings


SuperSandro2000:
Why not change it everywhere to update? Aktualisierung is very long and sounds quite weird...

SuperSandro2000:
It is consistent with the play store. So probably the best choice.

GuepardoApps:
Play store itself is not consistent using "Update" or "Aktualisierung". I would go with the german translation. In the end topjohnwu decides 😉

SuperSandro2000:
Bunmer.
He can't really decide if he does not know German

GuepardoApps:
Well, he can follow the discussion and decide then. We used "Aktualisierung" for a long time and for every translation of "Update". It is the german translation and there is nothing weird about it. It is not much longer then Update though...
I had a quick look at the Play Store, too. They use "Update" alongside with "Aktualisierung", but that's also not consistent. It's a mix and most used is the german translation "Aktualisierung".

SuperSandro2000:
Well he can't join the discussion cause it is about a opinion on the German language.
Well then it doesn't really matter

PR #673, opened 2018-10-12
scripts: ensure $APK gets set + support signing when Manager is hidden

osm0sis:


sjayminsgithub:
The default shell in Android is mksh, which does not perform filename generation for variable assignments. The asterisks in `APK=` will be literal in that case and this $APK will not be what's expected. One possible workaround is to use command substitution along with echo command.

osm0sis:
Updated with new `magisk --sqlite` command @topjohnwu added and informed me about. :+1:

Kadawe512:
c5758b3f2dc92d4dfa0129e8267c6d901240852c

PR #718, opened 2018-10-22
Fix crash when using other su

vvb2060:
#713


cobrafast:
When will there be a new build with this fix?
I believe I'm affected, too, and cannot install Magisk after a LineageOS update right now.
Using an older version of the manager app doesn't let me install Magisk without updating to the broken 6.0.1 version first.

PR #735, opened 2018-10-28
mostly maintenance

TacoTheDank:
- updated bouncycastle dependencies in utils from 1.59 to 1.60, and gms safetynet in snet from 15.0.1 to 16.0.0
- reformatted code, xml files, and pretty much everything
- completed a few lambdas that you missed (like [this](https://github.com/topjohnwu/Magisk/commit/a843ae4a3c92bbf78fd34e1254112ec4271d2c5c#diff-92b04e12b10b1785f06b5ab74e603302L53), for example)
- encoding set to UTF-8, just because
- in the sh files, put brackets around things, just because
- fixed some english in various places that i noticed


topjohnwu:
Some of the changes are done, but I won't take all of your changes

TacoTheDank:
@topjohnwu thats alright, all that matters is that you got some use out of it :)

PR #751, opened 2018-11-01
Add include stopped packages flag for broadcast

vvb2060:
FLAG_INCLUDE_STOPPED_PACKAGES = 0x00000020
If set, this intent will always match any components in packages that are currently stopped.


topjohnwu:
@vvb2060 massive thanks! Never know this flag exists LOL

PR #763, opened 2018-11-07
Fix grammatical errors, unnatural-sounding English and bad punctuation.

ianmacd:
This fixes some errors, uses English that a native speaker would use, and clarifies a couple of things.

### DIFF
--- a/app/src/full/res/values/strings.xml
+++ b/app/src/full/res/values/strings.xml
@@ -9,15 +9,15 @@
     <string name="install">Install</string>
 
     <!--Status Fragment-->
-    <string name="magisk_version_error">Magisk is not installed</string>
+    <string name="magisk_version_error">Magisk is not installed.</string>
     <string name="checking_for_updates">Checking for updates…</string>
     <string name="magisk_update_available">Magisk v%1$s is available!</string>
     <string name="invalid_update_channel">Invalid Update Channel</string>
-    <string name="safetyNet_check_text">Tap to start SafetyNet check</string>
+    <string name="safetyNet_check_text">Tap to start SafetyNet check.</string>
     <string name="checking_safetyNet_status">Checking SafetyNet status…</string>
     <string name="safetyNet_check_success">SafetyNet Check Success</string>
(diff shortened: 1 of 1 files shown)


ianmacd:
Yes, it's actually correct. The verbal compound 'will be' carries forward from the first clause to the second, where it is implied.

It's not strictly incorrect to add a second, identical verbal compound, but it's superfluous and therefore sounds clumsy to native ears. Non-native speakers are, indeed, prone to double-up in this way.

Regarding the possessive pronoun, English has historically used 'his' where the gender is unknown.

Opinions are divided on the use of 'their' in the singular as a modern, politically correct contrivance to work around this.

In the US, in particular, this has become widespread in recent years, particularly at Internet companies, leading to stilted, comical language when the user's name is obviously gender-specific, but he/she has elected not to make his/her name known to the underlying platform, e.g. 'Follow Susan and their postings.'

I personally don't care to pander to such trends, but whatever. Anything beats the "its" in the actual case in point from the file.

Didgeridoohan:
Just because something has historically been true, doesn't mean it is today...

I'm not trying to pander to any trend to eradicate gender differences in the language, it's just that "his" no longer is as much of a neutral pronoun as it has been in the past.

ianmacd:
> Just because something has historically been true, doesn't mean it is today...

No, it doesn't, but then you descend into the quagmire of defining truth and having to demonstrate what is "true" today. That, in turn, requires designating an authority on the matter and that's not trivial in the English-speaking world.

This is how I still speak today and I am far from the only one, so for some set of native English speakers (a significantly large one, I would argue), the practice is current.

> I'm not trying to pander to any trend to eradicate gender differences in the language, it's just that "his" no longer is as much of a neutral pronoun as it has been in the past.

That all depends on your perception. Those who see inequality in everything and wish to be offended by such trivial matters will seize upon them to do so. Those for whom it is nothing more than a linguistic convenience will continue to innocuously employ it as such.

ianmacd:
'His' has been changed to 'his/her' in response to @Didgeridoohan's comment.

One further change made to another string.

PR #826, opened 2018-12-05
Fix possible error block counts

Pzqqt:
Add "-k" parameter to force the unit to 1024 bytes.


topjohnwu:
Although the script is meant to only run under busybox environment (and `df` in busybox defaults to 1024 blocks), but adding a new flag doesn't hurt, right? :)

Pzqqt:
> Although the script is meant to only run under busybox environment (and `df` in busybox defaults to 1024 blocks), but adding a new flag doesn't hurt, right? :)

In my tests, I found that after calling `recovery_actions`, the unit of `df` is not necessarily 1024 bytes(512 bytes for me).

PR #904, opened 2018-12-23
Update tutorials.md

vichaos:
If "Logger buffer sizes" under "Developer Options -> Debugging" is set to "Off" magisk hide will never be triggerd.


topjohnwu:
This is no longer relevant in new MagiskHide implementations :)

PR #939, opened 2019-01-01
A full update to the French translation strings

DevDef:
Lots of mistakes fixed.
Cheers,

Davy


xerta555:
@DevDef Hello, i have take a look to your fixed strings, there are lots of errors:
- don't remove the backslash before each ' symbol bcz it's an escape carachter.
- never use conjugated verbs, always use them at the infinitive
- for the 'safetyNet_res_invalid' translation, i would have left my translation which seems more appropriate in view of the context of use
- concerning the symbol of the two points:, personally I prefer without a space between the word and the symbol, I find it clearer, but it is only a detail
- for the 'no_info_provided' and few other translations, i have respect the syntax, if in the initial translation there is a capital letter, i write a capital letter
- for the 'magisk_update_title' and the next one, i would have left my one bcz it's useless to take more space in a notification, especially when a shortcut can be used
- there are few another translations who don't begin with capital letter, it's a problem

Overall, apart from all the mistakes I've mentioned, there are some good ideas, but the translations can't be merged directly as they stand, it will generate a crash on John's side at the next compilation.

I think a global review is necessary, @topjohnwu thanks to not merging it, need a fix.

papjul:
@xerta555 I'm not the author of the pull request but I can give my point of view.
- The backslash is not needed if you use the typographic apostrophe ’ (which is the case in this pull request). The backslash is only needed against the simple quote '. So it will probably not crash the compilation.
- In French, typographic rules says that you have to put an unbreakable space between two-points symbol, so the word is not separated from its : symbol if there is a new line.
I will leave the other points to @DevDef

papjul:
> > In French, typographic rules says that you have to put an unbreakable space between two-points symbol, so the word is not separated from its : symbol if there is a new line.
> 
> On the second point, I do not remember a rule like this, if that is the case then indeed the corrections will have to be taken into account. From a visual point of view I prefer when there is no space, I find it more readable.

In France, typographic rules are published through "Lexique des règles typographiques en usage à l’Imprimerie nationale". Since you probably don't have this book, you can have a look at Wikipedia conventions which are based on this book: https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Conventions_typographiques#Espaces
« Sur Wikipédia en français, les signes de ponctuation doubles (« ; », « : », « ? » et « ! ») doivent être précédés d’une espace insécable et suivis d’une autre espace (à l’exception du deux-points quand il est utilisé pour exprimer une heure), ainsi que l’exigent les normes typographiques françaises, avec des nuances quant à la largeur de l’espace insécable placée avant, selon le signe de ponctuation double[21] (règles usuelles pour les publications imprimées) »
[21] Le Lexique, p. 149, préconise l’espace fine insécable pour : le point-virgule, le point d’exclamation et le point d’interrogation. Pour le deux-points il précise : « espace mots insécable ».

You can also find the same rule in Québec, Canada: http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?Th=2&t1=&id=2039

While it's best to use "espace fine insécable" in some cases, I don't recommend it in apps because they are not the easiest to find on a keyboard + if it is not supported by the font used in the phone running the Android system, it may actually result in a worse display.

DevDef:
All what @papjul wrote is right, and I’m glad that someone else took the time to detail all these French typographic rules. :-)
My proposed revised translation is respecting these rules, including the thin spaces before « ; », « ? » and « ! ». I’m sure that every Android devices since version 4.4 can handle thin spaces correctly (provided the used font is not too “exotic”).
And, yes, you need a non-breaking space before « : », no space is considered to be a mistake (as well as after a French left guillemet and before the right one).

> While it's best to use "espace fine insécable" in some cases, I don't recommend it in apps because they are not the easiest to find on a keyboard

Well, whether it’s hard to type it in Android apps is not a problem, because we don’t ask users to type it, these characters are just displayed in the apps’ strings. Concerning the app developer, it depends on its OS. On GNU/Linux, you’ll have to choose your own combination. Personally, I use `Alt Gr`+`<space>` for non-breaking space and `Alt Gr`+`<shift>`+`<space>` for thin space (note: `Alt Gr` is the right `Alt` on French AZERTY keyboard layout).

xerta555:
Okay, after reviewing all the points, the correction seems ready to me.
This will also allow me to review the different rules concerning the French language :)

DevDef:
Ooops again, there’s a typo in my last commit title in the word… typo! :$

papjul:
> > Well, when the strings are multiple choices in a select box that is part of a sentence, I usually don’t use a capital letter for these substrings. But, this is not very important. ;-) In your particular example, you used a capital letter after a left parenthesis, that’s wrong in French.
> 
> I didn't realize the context in which this sentence is used

There are typographic rules for this too:
https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Conventions_typographiques#%C3%89num%C3%A9rations
https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Conventions_typographiques#LISTES

TL;DR:
- Different sentences enumerated: capital letter.
- List introduced by « : »: no capital letter.
- List introduced by « : » but listed with 1) 2) 3) or A) B) C): capital letter.

For parenthesis: http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=3357
TL;DR:
- If the sentence is full, use a capital letter
- If it is part of another sentence, no capital letter

DevDef:
@papjul, do you see anything that deserves a change? A full review would be nice, it’s always easier to see someone else’s mistakes than our own ones.

papjul:
@DevDef 
Here are my two cents:
- Maybe remove capital V from version in install_magisk_title
- "sort_by_update" probably takes too much space in the context it is used. It may be simplified by "trier par date" or "trier par date récente", or maybe better, remove the "trier par" part which is redundant with "Mode de tri" (but you will need to remove it too from "sort_by_name")
- Is it possible to change "Manager" by "Gestionnaire" in the app_name? Because it is translated "Gestionnaire" everywhere but the app name is "Manager" so we need some sort of consistency here.
- "dl_one_module" may be translated "Un seul module peut être téléchargé à la fois" (in the context, it is more understandable than the infinitive "Télécharger un seul module à la fois", since it is a toast error when you try to download two modules at a time).
- I liked the original translation "settings_hide_manager_summary", "Réempaqueter" doesn't sound very French.
- "su_warning" is not consistant. "Autoriser" then "Refusez".

Other than that, it probably needs to be tested in context and be careful about correct wrapping of texts. I currently run Canary build for another issue so I may help there.

DevDef:
>     * Maybe remove capital V from version in install_magisk_title
Not maybe, but surely! I missed it.

>     * "sort_by_update" probably takes too much space in the context it is used. It may be simplified by "trier par date" or "trier par date récente", or maybe better, remove the "trier par" part which is redundant with "Mode de tri" (but you will need to remove it too from "sort_by_name")
Agreed, except that “trier par date récente” doesn’t sound very good French. I replaced the two strings by “par nom” and “par date décroissante”.

>     * Is it possible to change "Manager" by "Gestionnaire" in the app_name? Because it is translated "Gestionnaire" everywhere but the app name is "Manager" so we need some sort of consistency here.
You’re right this is not consistent. I don’t know if it’s easy to localize the app name… So, until we know more, I’ve replaced all the _Gestionnaire Magisk_ occurrences by “Magisk Manager”.

>     * "dl_one_module" may be translated "Un seul module peut être téléchargé à la fois" (in the context, it is more understandable than the infinitive "Télécharger un seul module à la fois", since it is a toast error when you try to download two modules at a time).
I missed the context. Given your explanation, I agree.

>     * I liked the original translation "settings_hide_manager_summary", "Réempaqueter" doesn't sound very French.
Though, it is a proper French verb: https://fr.wiktionary.org/wiki/r%C3%A9empaqueter
And _empaqueter_ is the official French term in every GNU/Linux distro.
BTW, I wonder if the English string _settings_restore_manager_summary_ is missing the word “name” in the end… So its translation is “Restaurer Magisk Manager avec son nom de paquet d’origine”.

>     * "su_warning" is not consistant. "Autoriser" then "Refusez".
Don’t agree. The first is the description of an action. The second is an injunction (hence the exclamation mark).

> Other than that, it probably needs to be tested in context and be careful about correct wrapping of texts. I currently run Canary build for another issue so I may help there.
Sure, that would be great! :-)
Thank you for your suggestions.

papjul:
Fine by me, I believe this can be integrated in the next Canary build for testing.

PR #978, opened 2019-01-19
Fix ways to detect Magisk through MagiskHide

Fox2Code:
Two changes are made in this repo.
The first is a fix of a bug that made the Magisk Unblock not deleted after Magisk had finished his work.
I have replaced the **0** permission when creating Magisk by **0770** to allow file deletion
The seconds is hiding the **/sbin** and **/root** modified date since only Magisk modify their content it can be used to detect Magisk. I have reverted it to the default kernel file time to make it like an unrooted phone

I have compiled with FrankeNDK and tested on my phone
(My phone is under Android Oreo 8.1.0)
I actually have my fork installed on my phone and it works fine

### DIFF
--- a/native/jni/daemon/bootstages.cpp
+++ b/native/jni/daemon/bootstages.cpp
@@ -31,6 +31,7 @@ char *system_block, *vendor_block, *magiskloop;
 
 static int bind_mount(const char *from, const char *to);
 extern void auto_start_magiskhide();
+extern void hide_mod(const char *file);
 
 /***************
  * Magic Mount *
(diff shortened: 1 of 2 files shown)


Fox2Code:
And the **hide_mod("/sbin")** need to be done just before **tmpfs** mount because too soon the modification time will be modified after and because if it's done after it will affect the **/sbin** overlay and not the real **/sbin**

vvb2060:
Google:1230796800, other OEM:0. Not everyone is 0.
You didn't modify ctime(change time).

Fox2Code:
Closed New pull request will be sent when changes are made (For kernel timestamp I will take **/res** I think google phone have this folder if not,  send me a message)

PR #979, opened 2019-01-19
Fix ways to detect Magisk through MagiskHide fixed (Thanks to vvb2060)

Fox2Code:
Previous pull request #978 (Please read if you want a complete overview of the changes)
This pull request manage to hide magisk from time stamp file detection and fix a bug that makes the Magisk "UNBLOCKFILE" not deleted 
Change the ChangeTime, ModifiedTime, and AccessTime timestamp of **/root** and **/sbin** to the kernel default timestamp when the app is under MagiskHide
/!\ Application not under MagiskHide see the Magisk's **/sbin** overlay but that's not the real **/sbin** folder seen by the applications under MagiskHide /!\

Thanks for @vvb2060 for reporting 2 bugs (see #978)


Fox2Code:
I have updated for the latest version of Magisk
But please instead of put **0** in value when you create "UNBLOCK FILE" something that allow their deletion because it seem on some phone the "rm UBLOCKFILE" in **magisk.rc** are not working if the file was created with a 0 permission (I have this issue on my phone)
Put 0770 fixed this issue on my phone
Thank for Magisk a I hope my help was helpful

Fox2Code:
The pull request is ready for the latest version of Magisk I had resolved all conflict and tested it on my phone if a modification need to be done on my side please inform me

Fox2Code:
I have a better idea to fix Magisk hide detection issue I will split them in two different pull request
since the last update, the **/sbin** timestamp is not modified but **/root** still have an issue

PR #987, opened 2019-01-22
Fix to return the proper path for pre v18 Magisk versions

### LINKED ISSUE #981: Magisk Manager stopped working on startup
Earlier I've installed a debug build to help solve #966, and that started to notify me that my device requires additional setup. First I've ignored that because I wanted to revert to a release build soon, but today I wanted to use an updated root app so Magisk took the permission from it, and for some reason it wasn't able to acquire it again. Then I went to the Superuser tab of the Manager, and the app list was empty. Then I decided to run that additional setup recommended by the Manager, and from now on it instantly crashes on startup. even after I updated it today. It seems that the problem is actually 2 problems:

The problem itself that causes the crash is in libsu, but I've taken a look on the [crash log](https://gist.github.com/MPeti1/31b4f5aca7aada48e31b025939151e2a) and that does not make sense to me because you don't try to get any array's length at that line, not even in previous versions, however I remember that in the past it did return a null array if there were no results in listFiles().

But the main problem is unknown to me. Is it a proper thing to just install the canary build over the release?

Please note that at the moment I've the 6.1.0-1d9c4410, 175 canary build installed but still Magisk 17.1 to be able to test Direct Install, and I've rebooted the phone.

MPeti1:
From Magisk v18 the new Magisk path is `/sbin/.magisk/img`, but before that it was `/sbin/.core/img`. It made Magisk Manager confused if used with an older Magisk version, for example v17.1
This commit fixes #981

### DIFF
--- a/core/src/main/java/com/topjohnwu/core/Const.java
+++ b/core/src/main/java/com/topjohnwu/core/Const.java
@@ -17,7 +17,7 @@ public class Const {
     public static final String SU_KEYSTORE_KEY = "su_key";
 
     // Paths
-    public static final String MAGISK_PATH = "/sbin/.magisk/img";
+    public static final String MAGISK_PATH = Config.magiskVersionCode < 18000 ? "/sbin/.core/img" : "/sbin/.magisk/img";
     public static final File EXTERNAL_PATH;
     public static File MAGISK_DISABLE_FILE;
 


topjohnwu:
This is intentionally removed. Magisk Manager will no longer support anything lower than v18 (as mentioned in the release notes).

PR #1020, opened 2019-02-02
Updated French translation

DevDef:
- added three missing strings (unsupport_magisk_title, unsupport_magisk_message and show_system_app)
- add “pid” string even if it is marked as not translatable, just to ease comparison
- better meaning for core_only_enabled string translation
- indent XML comment


topjohnwu:
Another contributor has already updated the strings. If you want to improve the translations, please open a new PR, thank you!

Issue #1018, opened 2019-02-02
[Feature Request]: Support Direct Boot


bychlz39104:
Some apps use su before unlock, but this time magisk manager is not available, unable to log and toast. When unlocked, the manager can start processing the backlog of events, the system is busy and the broadcast may have been lost.

Jackktheslayer95:
Jshshabs

topjohnwu:
@bychlz39104 the backlogs are not caused by the fact that Magisk Manager is not boot aware (fun fact, it is!). It is the way Android works: no broadcasts seems to work before `android.intent.action.BOOT_COMPLETED` is sent, which is even later after you unlocked the device

vvb2060:
@topjohnwu Manager really does not support direct boot( even you added `android:directBootAware="true"`, system ignored it). App use credential encrypted storage, so it cann't work.

official ver:
![official](https://user-images.githubusercontent.com/26996262/52168001-32d97800-275f-11e9-89a9-7b9dc9807b26.PNG)
dispatchTime=0 but finishTime=now, this broadcast was discarded without being sent.

unofficial ver:
![unofficial ](https://user-images.githubusercontent.com/26996262/52168059-d591f680-275f-11e9-8982-b56f3a618baa.PNG)
(use `reboot` instead of `boot` may make it work on chinese rom)
![test](https://user-images.githubusercontent.com/26996262/52168106-96b07080-2760-11e9-8955-64c11294d55e.PNG)


test version: 
official  6.1.0-56d2fb9a & 18.1-56d2fb9a
unofficial 6.1.0-56d2fb9a3 & 18.1-56d2fb9a3 (based on 56d2fb9a )
test method : use adb shell before unlock the device,  broadcasts history: `dumpsys activity broadcasts` .

PR #1032, opened 2019-02-04
Support Direct Boot

vvb2060:
#1018 
only system app can set android:directBootAware="true" at the application level.
for support direct boot, app need use de storage.

now, manager can log su and start request activity  before unlock


topjohnwu:
Huge thanks, I've done some modifications (repo database does not need to be in the de storage), but most of your implementation is preserved.

PR #1072, opened 2019-02-12
Completely rework MagiskHide

arter97:
Previous MagiskHide detects new app launches via listening through logcat
and filtering launch info messages.

This is extremely inefficient and prone to cause multiple issues both
theoratically and practically.

Rework this by using inotify to detect open() syscalls to target APKs.

This also solves issues related to Zygote-forked caching mechanisms such as
OnePlus OxygenOS' embryo.

Signed-off-by: Park Ju Hyung <qkrwngud825@gmail.com>


arter97:
I think this is worth bumping up the major version number to v19.0 or something imho :)

I tested this in OnePlus 6 OxygenOS, Mi Pad 4 PA, Mi Pad 4 PixelExperience and OnePlus 5 LineageOS.

There might be some corner cases that I might have missed, so please go ahead and review my code.

Thanks.

native/jni/magiskhide/proc_monitor.cpp
```diff
@@ -106,6 +146,209 @@ static void hide_daemon(int pid) {
...
+		pid = fast_atoi(dp->d_name);
+
+		// We're only interested in PIDs > 1000
+		if (pid <= 1000)
+			continue;
+
+		uid = get_uid(pid) % 100000; // Handle multiuser
+		if (hide_uid.find(uid) != hide_uid.end()) {
+			// Make sure our target is alive
+			if ((ppid = parse_ppid(pid)) < 0 || read_ns(ppid, &pns) || read_ns(pid, &ns))
```
vvb2060: app processes are not checked whether the separation is done or not, causing MagiskHide unmount stuffs in the namespace of zygote, all apps will no longer have root access.
btw, If user uses appops to ignore an app `READ_EXTERNAL_STORAGE` op, mount namespace of its process will not be separated from zygote.

native/jni/magiskhide/proc_monitor.cpp
```diff
@@ -106,6 +146,209 @@ static void hide_daemon(int pid) {
...
+				continue;
+
+			// Check if it's a process we haven't already hijacked
+			hide = false;
+			pos = pid_ns_map.find(pid);
+			if (pos == pid_ns_map.end()) {
+				hide = true;
+				pid_ns_map.insert(pair<int, uint64_t>(pid, ns.st_ino));
+			} else if (pos->second != ns.st_ino) {
+				hide = true;
```
vvb2060: Why true? We don't have to repeat the unmount again.

native/jni/magiskhide/proc_monitor.cpp
arter97: Is that what usleep() loop with st_ino was before @topjohnwu ?
It didn't work with the rework, but I'll check the corner case you've mentioned.
Thanks.

native/jni/magiskhide/proc_monitor.cpp
arter97: This is intentional.

inotify will report events even when an app didn't actually launch, so to prevent repetitive unmounts on an already hijacked process, the list of hijacked processes will be recorded in pid_ns_map.
In this case(pos->second != ns.st_ino), the pid was previously hijacked, but is now a different one because the previous one is terminated. Linux reuses pid numbers after all. So re-unmounting is necessary and the list is updated accordingly.

vvb2060: Memory usage will continue to increase

arter97: Theoretic maximum in absolute worst case is 8*32K as pid number is limited to 32K.
I chose to leave it that way without GC as traversing through the map on every detect_new_processes() is very inefficient.

But on a second thought, GC'ing the map upon update_apk_list() seems like a good idea and it'll even solve the map getting to the theoretical max size on extremely long uptime.

Will send another patch soon.

PR #1077, opened 2019-02-14
early boot magisk hide extension for sony ric daemon

j4nn:
This might be a bit specific to sony xperia xz1c/xz1/xzp phones with oreo fw, but maybe you can use at least the "extend magiskhide with exec option" commit and possibly implement a more generalized solution for the "detect sony ric daemon and hide magisk from it on boot" commit.
These two commits allow to do FOTA system update with unlocked bootloader on sony xperia xz1c/xz1/xzp phones with oreo fw if running a patched kernel which masks unlocked state in response from Trust Zone. The kernel is not flashed but booted from usb via 'fastboot boot' instead, having magisk root before and after fota.
For more details please see here:
https://forum.xda-developers.com/xperia-xz1-compact/development/rooted-kernel-hiding-bootloader-unlock-t3898711
Thank you very much for your excellent magisk tools.

### DIFF
--- a/native/jni/core/init.cpp
+++ b/native/jni/core/init.cpp
@@ -373,6 +373,10 @@ static const char wrapper[] =
 "unset LD_PRELOAD\n"
 "exec /sbin/magisk.bin \"${0##*/}\" \"$@\"\n";
 
+static const char ric_wrapper[] =
+"#!/system/bin/sh\n"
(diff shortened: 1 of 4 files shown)


j4nn:
Closed this pull request in favour of 
https://github.com/topjohnwu/Magisk/pull/1454
which is sufficient to accomplish the original purpose.

Issue #980, opened 2019-01-19
Incompatible With System-As-Root Without Ramdisk


topjohnwu:
Many Chinese devices uses system as root without ramdisk block. A fix shall come soon (please check #1102)

xiaowyyyyya:
Great to see your reply, really appreciate that! And by the way, is not only Chinese MIUI not working, all of the 3rd party AOSP, MIUI are not working too. Hopefully you can find the way to make it work! Thank you very much!

DeliriousMetro:
That is a bit strange.

I own an Mi 6X which runs the custom rom - Xiaomi.eu (which is based off the China base rom) and Magisk 18.0 works fine on it and other versions too.

It does lag out if you install Magisk straight after the rom, I would have to install the rom first and wipe cache then boot up and then flash Magisk. If you install Magisk straight after the rom installation including a wipe without boot, the device will lag out a lot.

xiaowyyyyya:
> That is a bit strange.
> 
> I own an Mi 6X which runs the custom rom - Xiaomi.eu (which is based off the China base rom) and Magisk 18.0 works fine on it and other versions too. Infact, I have been running Magisk 17.X since purchasing the phone when it came out.
> 
> It does lag out if you install Magisk straight after the rom, I would have to install the rom first and wipe cache then boot up and then flash Magisk. If you install Magisk straight after the rom installation including a wipe without boot, the device will lag out a lot.
> 
> Seeing the reports about Magisk not working on Mi 8 SE on the thread over at Xiaomi.eu makes me think this might be a device specific matter?

It is a device specific issue, it's just for Mi 8 SE

xiaowyyyyya:
@topjohnwu this is a device specific issue related to Mi 8 SE, not for all Xiaomi devices.

solidv:
Same problem with Mi 8 SE, last magisk to work is version 16.6

kissunyeason:
just xiaomi 8se，not all xiaomi device

xiaowyyyyya:
@isaac chen

xiaowyyyyya:
@tingyichen

xiaowyyyyya:
@topjohnwu I personally did some research based on this issue, and it seems that most of the ### **Snapdragon 710** devices (Xiaomi M8 SE, Smastisan Pro 2S, etc.) have the same issue. Therefore, in conclusion, I think Magisk is somehow incompatible with **Snapdragon 710**, not only Xiaomi devices.

[...]

PR #1102, opened 2019-02-19
Support Aonly&System-as-root devices

vvb2060:
Partition layout: https://source.android.com/devices/bootloader/system-as-root#about-system-as-root

In `boot.img`, we patch kernel to make it want_initramfs , and add stub ramdisk. There are only `init` in the stub ramdisk. 
magiskinit already supports system-as-root devices.

#1064

PR #1115, opened 2019-02-22
Trim dev name

vvb2060:
#980 mi8se


vvb2060:
Xiaomi8SE and Smastisan Pro 2S need this, it's different from the case of #1102.

PR #1136, opened 2019-02-27
Use REBOOT foreground broadcast

vvb2060:
Many Chinese ROMs block BOOT_COMPLETED  broadcast. Some ROMs require users to add app to system whitelist, but some ROMs only system app can auto start. Users have to install Manager as system app and give up hide Manager.
This PR use REBOOT broadcast. Send broadcast to component instead of package will skip intent filter and permission check. REBOOT broadcast only root and system can sent.


vvb2060:
Xiaomi users feedback this PR can bypass the MIUI limit

topjohnwu:
Seems like a good idea! Thanks

Issue #1149, opened 2019-03-01
UNBLOCKFILE doesn't get deleted


Fox2Code:
The magisk UNBLOCKFILE (**/dev/.magisk_unblock**) doesn't get deleted

I provide the **logs.zip** witch contain three file
magisk.log > The logs of magisk
logcat.log > the boot logcat of my phone
cmdresult.txt > the result of the command **ls -la** on **/dev** folder

[logs.zip](https://github.com/topjohnwu/Magisk/files/2919526/logs.zip)

This is a big hiding issue because magisk can be detected just by listing the content of the **/dev** folder

topjohnwu:
What version are you using? It is not present on any of my devices.

Fox2Code:
@topjohnwu I found why this happen it due to the post-fs-data duration is more than 10 second
I have enable the core only mode and the file was gone

PR #1154, opened 2019-03-02
Do not create UNBLOCKFILE if it taken too long

Fox2Code:
This is a fix of #1149 

The magisk UNBLOCKFILE doesn't get deleted because the module loading take more than 10 seconds (since it module dependent we can't make it faster) and the file was created after the **rm UNBLOCKFILE** was called in **init.rc** 

so to fix that I create a file **POSTFSDATA** (/dev/.magisk_post_fs_data) that is created before **/sbin/magisk --post-fs-data** and get deleted after the **wait UNBLOCKFILE 10**
And magisk create the **UNBLOCKFILE** only if the **POSTFSDATA** exists

### DIFF
--- a/native/jni/core/bootstages.cpp
+++ b/native/jni/core/bootstages.cpp
@@ -571,7 +571,9 @@ static void dump_logs() {
  ****************/
 
 [[noreturn]] static void unblock_boot_process() {
-	close(xopen(UNBLOCKFILE, O_RDONLY | O_CREAT, 0));
+	if (access( POSTFSDATA, F_OK ) != -1 ) {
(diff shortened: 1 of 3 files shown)


Fox2Code:
[Sleep11sec.zip](https://github.com/topjohnwu/Magisk/files/3333783/Sleep11sec.zip)
I'm back!!! I provide the following module to reproduce the bug (The module just wait 11 seconds on post-fs-data) 

I still have the bug with the latest version of the magisk cannary 

I will make a post on Issues with better idea to fix this bug since I think the way I done the fix is maybe not the best way to fix the bug even if it can be a good temporary solution

Fox2Code:
Issue I created about fixing the bug => #1641

PR #1165, opened 2019-03-04
Updated Russian translation

zertyuiop:
Added translation for missing strings. Fixed some lines for better language.

### DIFF
--- a/app/src/full/res/values-ru/strings.xml
+++ b/app/src/full/res/values-ru/strings.xml
@@ -7,6 +7,8 @@
     <string name="log">Логи</string>
     <string name="settings">Настройки</string>
     <string name="install">Установка</string>
+    <string name="unsupport_magisk_title">Неподдерживаемая версия Magisk</string>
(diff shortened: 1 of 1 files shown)


igor-dyatlov:
@zertyuiop  #1157 two days ago sent a request.

zertyuiop:
@igor-dyatlov Yes, I know, but you have made some errors in your translation. For example "(Core only mode enabled)" is not "(Включен только Core режим)", it stated that the Magisk Core mode enabled, so you need to remove "только" from this line. Another examle: "Скачать один модуль за раз." This message appeares when you try to download two modules at ones so the translation should be "Возможно скачивание лишь одного модуля за раз". You also add dots in the end of some strings, but I think they look ugly. On the other hand, I fixed some translation errors, for example "namespace" is not "именное пространство" but "пространство имён". I also changed translation of "hide" from "скрыть" to "замаскировать", It is more suitable for how Magisk Hide works.

igor-dyatlov:
@zertyuiop точки добавил, так как они есть в оригинальных стрингах. Внес ваши изменения. https://github.com/topjohnwu/Magisk/pull/1157/commits/13d4d4a8608053532e8f94b28f057a357fc40ac4

zertyuiop:
I see you added all my fixes so I can close this request.

PR #1182, opened 2019-03-08
fix parse_int

ahxxm:
errors from compiling:

```
In file included from jni/magiskhide/proc_monitor.cpp:24:
jni/utils/include/utils.h:209:22: error: member reference base type 'char *' is not a structure or union
        return parse_int(__s.data());
                         ~~~^~~~~
jni/magiskhide/proc_monitor.cpp:107:14: note: in instantiation of function template specialization 'parse_int<char *>' requested here
                        int uid = parse_int(value);
                                  ^
1 error generated.
```

our parse_int only takes `.data()` from `string_view`, it does not have to be a template

### DIFF
--- a/native/jni/core/db.cpp
+++ b/native/jni/core/db.cpp
@@ -43,7 +43,7 @@ int db_settings::getKeyIdx(string_view key) const {
 }
 
 static int ver_cb(void *ver, int, char **data, char **) {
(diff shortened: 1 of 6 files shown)


topjohnwu:
I use templates because I want to also support potential `std::string`. Stupid C++ standard make these 2 things different, what can I do :)

PR #1209, opened 2019-03-12
New MagiskHide UI

vvb2060:
Screenshot_20190312202208


DenyDarko:
Hey @vvb2060 that's lovely :D

One question out of curiosity though. 
Why Afwall gets a second line while it has only one sub-package on it?

PR #1258, opened 2019-03-28
Translation done

Gozzwip:
Please change the name of the language to Azərbaycanca


topjohnwu:
The name of the language is auto generated, not specified by me

PR #1338, opened 2019-04-14
Replace drawer menu with bottom navigation

paphonb:
This has been my biggest annoyance when developing and having to navigate around the Magisk Manager app quite often. 

I find that it is very slow to click the button on the top left (or dragging the menu from the left edge) and then reaching up to click the menu I want to be in. It's much easier and faster to just tap the target on a bottom bar.


topjohnwu:
@paphonb, the UI for Magisk Manager will no longer be maintained by me, so you'd want to discuss with @diareuse regarding designing decisions. I already pinged him regarding this issue.

diareuse:
Contribution is appreciated however bottom bar will be included with a new design, hence there is no point of introducing it at this time.

These changes will be overwritten by bunch of others that will be coming in.

PR #1342, opened 2019-04-16
[Arch] Rewritten homepage

diareuse:
_Not final redesign._

Cleaned up homepage homepage a bit xP
In addition new navigation is introduced in later commits.

Stable; *with dark theme*.


topjohnwu:
Check branch WIP

PR #1355, opened 2019-04-21
Removing android app data backup

termux0:
This commit will remove/turn off backup functionality for ADB to backup Magisk manager's private directory `/data/data/com.topjohnwu.magisk/<any file and folder>`.

That private directory is only accessible to that app only but setting `android:allowBackup="true"` in manifest will allow ADB(Android debug bridge) to take backup of private directory it means whole directory will get dumped.

This might affect security of user's device because that backup can also be restored any malicious code/file can be injected or any file can be manipulated or altered.

You can also allow backup of only some or specific file/folders.

Reference/Support material:-
https://developer.android.com/guide/topics/data/autobackup.html#EnablingAutoBackup
https://blog.devknox.io/what-is-android-allowbackup-how-it-affects-your-app/

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,7 +9,7 @@
     <uses-permission android:name="android.permission.USE_FINGERPRINT" />
 
     <application
-        android:allowBackup="true"
+        android:allowBackup="false"
         android:name="a.e"
         android:theme="@style/AppTheme"
         android:usesCleartextTraffic="true"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -12,7 +12,7 @@
         android:allowBackup="false"
         android:name="a.e"
         android:theme="@style/AppTheme"
-        android:usesCleartextTraffic="true"
+        android:usesCleartextTraffic="false"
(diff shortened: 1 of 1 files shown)


termux0:
It is a debugging feature of android and this should not be allowed on a stable version

termux0:
Setting `android:usesCleartextTraffic` to false

Indicates whether the app intends to use cleartext network traffic, such as cleartext HTTP. The default value is "true".
When the attribute is set to "false", platform components (for example, HTTP and FTP stacks, DownloadManager, MediaPlayer) will refuse the app's requests to use cleartext traffic. Third-party libraries are strongly encouraged to honor this setting as well. The key reason for avoiding cleartext traffic is the lack of confidentiality, authenticity, and protections against tampering: a network attacker can eavesdrop on transmitted data and also modify it without being detected.

Read more here --> https://developer.android.com/guide/topics/manifest/application-element

topjohnwu:
`android:allowBackup="true"` is explicitly set to true because non of the sensitive data is stored in the app's internal data. All sensitive settings (root access, root related settings etc.) are stored externally in `/data/adb/magisk.db`.

`android:usesCleartextTraffic` is also explicitly set to false because many links in module's `README.md` contains `http` links, and I do not want to block them.

Issue #1226, opened 2019-03-19
RootBeerFresh detects Magisk 18.2


Ingan121:
It's "MAGISK UDS AND STAT" process detects Magisk even if MagiskHide is being used.
Its source is available here: https://github.com/kimchangyoun/rootbeerFresh
Also, sample app is available here: https://play.google.com/store/apps/details?id=com.kimchangyoun.rootbeerFresh.sample
Tested on Samsung Galaxy S7 (SM-G930L) with 8.0 Oreo
![Screenshot_20190320-000445_RootbeerFresh](https://user-images.githubusercontent.com/34885971/54616941-cfec3800-4aa3-11e9-9d94-6fb99800076a.jpg)

Ingan121:
UDS is Unix Domain Socket, and it detects magisk by checking /proc/net/unix (It may have more ways to detect Magisk.)
NDK C++ Source of UDS detecting part ([Full](https://github.com/KimChangYoun/rootbeerFresh/blob/master/LibRootbeerFresh/src/main/jni/toolChecker.cpp)): 
```/*****************************************************************************

* Description: Check the Unix Domain Socket used by Magisk

*

* Parameters: none

*

* Return value: 0 - non-existant / not visible, 1 or more - exists

*

*****************************************************************************/

int Java_com_kimchangyoun_rootbeerFresh_RootBeerNative_checkForMagiskUDS( JNIEnv* env, jobject thiz )

{

    int uds_detect_count = 0;

    int magisk_file_detect_count = 0;

    int result = 0;


    // Magisk UDS(Unix Domain Socket) Detection Method.

    // The unix domain socket is typically used for local communications, ie IPC.

    // At least Android 8.0 can look up unix domain sockets.

    // You need to be sure that you can query the unix domain socket on Android 9.0 or later.

    FILE *fh = fopen("/proc/net/unix", "r");

    if (fh) {

        for (;;) {

            char filename[BUFSIZE] = {0};

            uint32_t a, b, c, d, e, f, g;

            int count = fscanf(fh, "%x: %u %u %u %u %u %u ",

                               &a, &b, &c, &d, &e, &f, &g);

            if (count == 0) {

                if (!fgets(filename, BUFSIZE, fh)) {

                    break;

                }

                continue;

            } else if (count == -1) {

                break;

            } else if (!fgets(filename, BUFSIZE, fh)) {

                break;

            }


            LOGD("%s", filename);


[...]

PR #1359, opened 2019-04-22
Anti-detection

vvb2060:


Ingan121:
Does this solve #1226?

Ingan121:
@vvb2060 I think the max digit should be less than 32, to remove possibility of generating 32-digit socket name and being detected by existing UDS-based detection methods, which currently check for socket names longer than 32 chars.

PR #1362, opened 2019-04-25
Fix installation for A-only system-as-root devices

DD3Boh:


DD3Boh:
DO NOT MERGE (yet)
Still needs a review for Berkeley that right now is mounting /system weirdly for A-only system-as-root

topjohnwu:
That function is fully tested on a lot of devices, I don't see any issues with that so far.
What are you trying to fix here?

DD3Boh:
On a device with a correctly configured A-only system-as-root (which is present officially on something like 5 devices) the system partition is mounted as /system_root, which means that magisk can't mount it just using "mount /system"
The real bug was that the zip couldn't be flashed because not able to mount system, unless /system was already previously mounted. And of course every /system mount failed together with it.

topjohnwu:
That just isn't true. As long as there is a block named `system`, it will be mounted

DD3Boh:
Actually no, because there is no system folder
You would need to put an mkdir there for /system in case it isn't present. 
In my opinion it's better to mount it the way the recovery would mount it normally instead of remounting the partition manually to /system_root unconditionally
Anyway if you don't want to merge these commits at least add a mkdir for /system in case it isn't existing

DD3Boh:
Let me correct myself about the above statement.
Mkdir system won't work, because /system is there and isn't there at the same time
Indeed with a proper system as root implementation (A-only) /system is a symlink of /system_root/system
So `ls /system` gives `No such files or directory` but `mkdir /system` gives `File exists`

DD3Boh:
Re-done on https://github.com/topjohnwu/Magisk/pull/1376

PR #1376, opened 2019-04-27
Properly mount system for A-only system-as-root devices

DD3Boh:
* A properly configured A-only system-as-root device in recovery should have system mounted to /system_root and /system being a symlink of /system_root/system
* A-only system-as-root devices don't need to be manually mounted to /system_root doing it as default

Signed-off-by: Davide Garberi <dade.garberi@gmail.com>

### DIFF
--- a/scripts/flash_script.sh
+++ b/scripts/flash_script.sh
@@ -78,7 +78,7 @@ chmod -R 755 $MAGISKBIN
 # addon.d
 if [ -d /system/addon.d ]; then
   ui_print "- Adding addon.d survival script"
-  mount -o rw,remount /system
+  mount -o rw,remount $SYSTEM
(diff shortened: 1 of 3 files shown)


DD3Boh:
Tested and working on:
A-only non system-as-root
A-only system-as-root properly configured (system mounted to /system_root in recovery)
A-only system-as-root poorly configured (system mounted to /system in recovery)
A/B

DD3Boh:
@topjohnwu I think that this patch is ready to go and could be merged, no issues shown on any type of devices.

WardZhouA:
> @topjohnwu I think that this patch is ready to go and could be merged, no issues shown on any type of devices.
Is this the issue that caused failure when flashing pixel 2xl on Android Q beta3?

PR #1399, opened 2019-05-02
exclude useless files

vvb2060:


topjohnwu:
Are you sure those files are absolutely useless? I know that the `*.version` in `META-INF` is definitely useless, but I'm not so sure for those in `/org` and `/kotlin`

vvb2060:
https://discuss.kotlinlang.org/t/kotlin-1-2-50-increases-apk-size-by-97-kb-versus-1-2-41/8228

PR #1452, opened 2019-05-13
Fix boot failure caused by boot unsigned

muicoder:


topjohnwu:
We only sign boot images that were originally signed. That is intended behavior.

PR #1454, opened 2019-05-13
extend magiskhide with exec option

j4nn:
this shall be used to start a process in "stock" environment
where magisk components are not visible

this is needed in very early boot phase to start special root
detection processes like sony xperia ric daemon

when ric daemon thinks bootloader is still locked and detects
presence of /sbin/su or other utilities, kernel is rebooted
in it's very early boot phase

this patch adds new --exec option which will create private
mount namespace and change all mount points sharing to MS_PRIVATE
so hide_daemon() can be used to unmount magisk mounts just before
exec() to target process

### DIFF
--- a/native/jni/magiskhide/magiskhide.cpp
+++ b/native/jni/magiskhide/magiskhide.cpp
@@ -6,10 +6,12 @@
 #include <string.h>
 #include <sys/wait.h>
 #include <sys/types.h>
+#include <sys/mount.h>
 
 #include <magisk.h>
 #include <daemon.h>
 #include <flags.h>
+#include <utils.h>
 
 #include "magiskhide.h"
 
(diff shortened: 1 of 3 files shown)


j4nn:
The purpose of this change is to hide magisk from sony ric daemon which is started very early on boot with sony xperia xz1c/xz1/xzp oreo fw.
If ric daemon can find magisk components (like su binary) while detecting phone still as bootloader locked, a reboot is triggered.

With this change and a kernel patch masking bootloader unlocked state in response from TrustZone we can do full FOTA system update of sony stock fw with unlocked bootloader having magisk root before and after the update.
For more details please see here:
https://forum.xda-developers.com/xperia-xz1-compact/development/rooted-kernel-hiding-bootloader-unlock-t3898711
Thank you very much for your excellent magisk tools.

topjohnwu:
@j4nn do you mind giving joining Magisk's internal slack group for further discussion? If you are willing to, just send me an email 😃

j4nn:
@topjohnwu, to be honest, never heard of "slack groups" before. I can try it if you think it's needed, but would prefer simple comments here if possible.

j4nn:
@topjohnwu, I just send you an email as you've requested to discuss further via that slack thing...

PR #1488, opened 2019-05-21
Use the standard Android Reboot API to perform reboot

Trumeet:
# Purpose
Magisk currently uses `reboot` command to perform hard-reboots which is not good for devices and developers, that's because:

* Android has its own shutdown/reboot procedure
* Some developers would like to receive the shutdown broadcasts
* Inside Android's own procedure, SystemServer shutdowns services gently so it's safe to perform the hard shutdown at last. Performing it without shutting down the services may cause errors.

# Is this PR tested?
I'm not sure if that the `svc power reboot` method works perfectly from Android L to P but I've taken a look about this part of the source of L and P and they look similar. The reboot API may be changed by OEM and may won't work normally, but on my Android Pie (AospExtended) device it works perfectly in debug mode. 

Another thing is that I've tested rebooting to normal, recovery and bootloader successfully but I cannot reboot into `download` and I have no idea about it. 

# How does it work?
I use `svc power reboot <reason>` instead, it should be more stable than the pervious commit which uses `service call`.

# Acknowledgement
Thanks to RikkaW and Kr328

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/utils/RootUtils.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/utils/RootUtils.kt
@@ -163,16 +163,7 @@ class RootUtils : Shell.Initializer() {
 
         @JvmStatic
         fun reboot(reason: String) {
-            Shell.su("service call power ${getRebootTransactionCode()} i32 0 s16 $reason i32 0").submit()
-        } 
-
-        private fun getRebootTransactionCode(): Int {
-            // Get the transaction code for the reboot method in PowerManager.
-            // Warning: Hidden-api usage.
-            val clz = java.lang.Class.forName("android.os.IPowerManager\$Stub")
-            val field = clz.getDeclaredField("TRANSACTION_reboot")
-            field.setAccessible(true)
-            return field.getInt(null)
+            Shell.su("svc power reboot $reason").submit()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/topjohnwu/magisk/utils/RootUtils.kt
```diff
@@ -158,7 +158,21 @@ class RootUtils : Shell.Initializer() {
...
+        @JvmStatic
+        fun reboot(reason: String) {
+            Shell.su("service call power ${getRebootTransactionCode()} i32 0 s16 $reason i32 0").submit()
+        } 
+
+        private fun getRebootTransactionCode(): Int {
+            // Get the transaction code for the reboot method in PowerManager.
+            // Warning: Hidden-api usage.
+            val clz = java.lang.Class.forName("android.os.IPowerManager\$Stub")
+            val field = clz.getDeclaredField("TRANSACTION_reboot")
```
RikkaW: Don't assume the field of `reboot` transaction code is `TRANSACTION_reboot`, the format may be `TRANSACTION_reboot_XX` (XX is number).

Trumeet: Could you give me an example of how that XX works? Thanks.

RikkaW: I only know some examples, such as `TRANSACTION_getInstalledApplications_49 = 50` appears on API 25.

Trumeet: > I only know some examples, such as `TRANSACTION_getInstalledApplications_49 = 50` appears on API 25.

Thanks.

Trumeet:
The latest commit did not work on my device... After debugging I will re-open it.

zareii:
J530fxxu4android8.01 root magiske

PR #1531, opened 2019-05-31
Update samsung.md

ianmacd:
MTP is now known to sometimes corrupt the AP file on transfer to the PC, so we should warn users to prefer `adb`.

Furthermore, quite a few users are reporting a shrunken `/data` file-system after flashing with Odin. This has been traced to the flashing of only an AP file, which causes some versions of Odin to shrink `/data`. The phenomenon is reproducable.


topjohnwu:
@ianmacd `samsung.md` is no longer used (I just didn't removed it yet). If you want to add, pleas add it directly to `install.md`

PR #1577, opened 2019-06-11
Redirect system su to MagiskSU

vvb2060:
Chinese MIUI ROM have built-in su, and the system partition is protected by avb.


ewing03:
Cool thanks

topjohnwu:
I don't necessarily want to mix system su with Magisk su.
`/sbin/su` will almost certainly always have precedence over `/system/(x)bin/su`, so calling `su` will actually call Magisk

PR #1583, opened 2019-06-11
Add missing strings for verity and force encrypt removal

Zackptg5:
Add missing strings for verity and forceencrypt.
fileencryption is used on oneplus devices (and probably others) (I've verified this on my OP5T)
support_scfs is used on some samsungs (see [https://twitter.com/arter97/status/614118117719932928](url) and [https://forum.xda-developers.com/note5/orig-development/kernel-discussion-root-recompiling-t3219990](url)


Zackptg5:
So if there's multiple matches together like `support_scfs,verify`, then the comma will only get removed for the first match which would leave an extra comma on the line. Still looking into why that occurs

Zackptg5:
closed till I figure this out

PR #1601, opened 2019-06-16
Fix error using Magisk with bootup pin code enabled

shakalaca:
Magisk daemon creates new magiskdb before /data decryption and uses it afterwards, the status of hide and superuser requests will be out of sync because of using different magiskdb after /data decryption in Manager.


shakalaca:
ah, missing one line of code ..

topjohnwu:
@shakalaca `/data/adb/magisk.db` should not be credential encrypted; it should be device encrypted, which should already be readable when Magisk started up. What setup is your device using?

shakalaca:
For the record: ROG Phone with FDE (by default)

PR #1603, opened 2019-06-16
Make /sepolicy unreadable by apps

arter97:
Some apps tries to read /sepolicy and see if it's patched for su.

Signed-off-by: Park Ju Hyung <qkrwngud825@gmail.com>


vvb2060:
What app can access the u:object_r:rootfs:s0 context?

arter97:
Permissive kernel.

topjohnwu:
Permissive kernels already broke all advanced sandboxing on modern Android, which magiskhide heavily depends on.

PR #1606, opened 2019-06-16
Add lzop kernel and ramdisk.cpio support to Magisk using busybox fallback

osm0sis:
scripts: boot_patch: support recognized but unsupported compressions
- magiskboot can recognize but does not support decompressing lzop, but the installer also includes busybox, which does, so test for recognized compressed kernel or ramdisk.cpio after unpack and use busybox to do so
- scripted generically in case more formats fit this criteria in the future

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -84,6 +84,17 @@ case $? in
     ;;
 esac
 
+for part in kernel ramdisk.cpio; do
+  var=$(basename $part .cpio)comp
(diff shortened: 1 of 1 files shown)


osm0sis:
Not sure if you'll care for the script changes required, but they're fairly simple and execute quickly, so I wanted to at least show how easy this would be to add. 😉

osm0sis:
Let me know if there is some other way you'd prefer this done and I'll revise. 👍

topjohnwu:
Sorry bro, but I do not want to support formats that are not supported by `magiskboot` natively. This will piss off the 3 people out there using LZOP, but I really don't care :)

PR #1624, opened 2019-06-22
Delete unused variable

JoanVC100:
I don't know if this variable can be deleted, but build.py says that is unused, so... (I'm not an expert with code, please don't kill me if I'm wrong) 
:')


topjohnwu:
Thanks for the PR, but the code is almost a direct copy from AOSP, and I want to keep it in sync :)

JoanVC100:
At least I tried :')

PR #1630, opened 2019-06-24
Manager: fix ModulesFragment reboot menu

osm0sis:
- correct 'booloader' typo breaking bootloader entry
- remove extra bootloader entry Shell.su line which is unnecessary since it's covered by reboot()
- revert to using `reboot recovery` for recovery entry since `svc power reboot recovery` triggers a very disconcerting "Factory data reset" reboot dialog on many devices
- add Reboot to EDL mode option for good measure


DenyDarko:
Just in the mood for commenting.

This reboot to recovery thing makes kinda sense. 

I guess the reason for non-twrp, stock recovery handling that way the "official" svc reboot way is for end users to be easier to wipe their devices. 
Less steps. 

Meaning that a "normal" end user, will only be doing that (rebooting into recovery) for reset purposes.

PR #1655, opened 2019-07-01
update camera access

Prakyy:
cam access to scan a qr code to root your phone automatically... pm for additional info


topjohnwu:
Nope :)

PR #1681, opened 2019-07-09
Add keyStorePass to config.prop

mohammad92:


mohammad92:
For some reason it wasn't there and I done a clean merge and cherry-picked this commit again and ...

PR #1685, opened 2019-07-10
Add "sepolicy_custom"

RikkaW:
Huawei kernel only accepts sepolicy for once. So if we need to add custom SELinux rules, the only choice is to compile Magisk with new rules.

There are already multiple projects/apps require different rules, so I choose to create a custom mechanism. And the reason for not choose to open pull requests to add rules is it's inappropriate to add rules related to untrusted_app.

The change is simple, read sepolicy_custom in `early_mount` and use sepolicy_custom in `patch_sepolicy` if exists.

The user needs to use the ["custom sepolicy installer"](https://github.com/RikkaApps/magisk-custom-sepolicy-installer). It's modified from Magisk installer. On install, it generates sepolicy_custom using the `magiskpolicy` tool and adds sepolicy_custom to boot img. So devs only need to provide zips with their own rules.


SuperSandro2000:
Can you please rebase instead of merge master into this PR? This is really really noisy and break git graph.

XspeedPL:
Are there any official plans regarding this?
A few people would love to see this included.

gyj1109:
Are there any new progress?

topjohnwu:
I am working on a complete new implementation, this PR will be kept open until it is added officially.

Evengard:
Waiting for it, because newer versions of EdXposed migrated already to this new system, it just won't boot on the official Magisk ending up in a bootloop and I don't want to install an older Magisk with the patches... So I'm stuck on an older version.

PR #1688, opened 2019-07-11
Some fixes

Gozzwip:


Gozzwip:
Some scripts are translated but it shows in English
That is an example

PR #1694, opened 2019-07-13
MagiskInstaller: check Download for matching file

osm0sis:
- save bandwidth if latest installer zip already exists from previous Download Only
- existing Download zip must match naming scheme for latest zip and of course match MD5


topjohnwu:
Already handled by changes in @diareuse's new implementationn

PR #1699, opened 2019-07-15
WIP: Magisk: Add system mode

mohammad92:
Locked bootloader devices can't install magisk patched boot.img
will leave this for future


topjohnwu:
Sorry, but these kind of huge changes without discussion will not be accepted.

The so-called "system" mode will *NOT* work on non system-as-root devices due to SELinux. 

"System installed Magisk" is only possible on system-as-root devices, and is not even remotely close to what you are proposing here.

mohammad92:
This was meant for discussion, and not yet ready.
And this for 7.0 and older versions of android which has permissive kernel. 
If our problem with SAR we can force boot patching and skip system mode.
I can do unofficial builds for certain devices only.

PR #1829, opened 2019-09-08
new strings added

Gozzwip:


Gozzwip:
Please add new strings so it can be easy to find and translate, they are not in the same order as in the original file.

Issue #1286, opened 2019-04-01
Cannot disable modules without recovery 


simymo:
I have a huawei Mate 20 (HMA-L29) which doesn't have a compatible twrp recovery, as a result, I installed Magisk using the patched .img file downloaded from this post and everything works fine.

Today I installed a bad Magisk module and got a bootloop after reboot, I flashed back my stock recovery file and it revived with Magisk disabled. But now my problem ls every time I enabled Magisk by flashing the patched img, it will enter bootloop mode again, even the Magisk Manager is uninstalled completely, and I have no way to disable the installed modules without root access or magisk enabled. (Neither could create a ".disable_magisk" file in /cache via adb shell with no root access)

Rather than putting a .disable_magisk in cache folder, I believe it will be more helpful if it can be temporary disabled by placing a file in some specified path in sdcard.

A2HI2H:
Try with flashing stock boot image

VR-25:
Due to file based encryption, /data/media/0 is scrambled by the time Magisk starts. Placing a file there is unreliable, unless you don't use FBE (unlikely when unrooted). A potential solution would be forcing core only mode with a key hold/press.

topjohnwu:
I did try to implement bootloop prevention but it had caused some issue. I will try to make it more stable in the next release

jancm:
Untested but a possible temporary hack:

Place a script in /data/adb/post-fs-data.d
Have the script check for a flag file in /sdcard/
i.e. [ -e /sdcard/disable-magisk ]&&touch /cache/.disable_magisk

If you get a boot loop you can use adb to write the flag file to /sdcard

post-fs-data.d runs before any modules are loaded so should run before the boot loop kicks in. The next attempted boot after that should act on the .disable_magisk file.

Like I said untested, definitely not the cleanest method and a total hack - but I think it should work until something better comes along. Better than having to wipe data.

RaymondSchnyder:
What if you could place the .disable_magisk file to /system?
Yea yea, i know. The system gets tampered.
But at this point, this could help:
1. Detect if /system/.disable_magisk exists
2. if yes, enable core only mod

After that, the user is able to remove the troublemakers and can restore the systempartition from his backup, so its systemless again.

Is that an ugly workaround or an idea?
because wiping my /data is definietly not something i would rather do. 

Also, if i understood the documentation correctly, on A/B Devices, the file should be placed at /data/cache/.disable_magisk.
Is that the only place?

[...]

PR #1842, opened 2019-09-14
Can force enable core mode when patching image

vvb2060:
#1286


topjohnwu:

PR #1876, opened 2019-09-27
dtb: support patching qcom dtbtool generated dt images

YoDevil:
Old qcom bootloaders for chips like msm8916, msm8974 load dt's from a dt.img appended to the end of the boot.img. This is already handled in the boot parsing section of magisk as what you call the "extra" section.
However, the new dtb patching to support A-only 2SI SAR did not take that into account.

The image has special formatting, consisting of
* A common header, which holds information like the header version and the number of dts in the image.
* A table of infos for each included dtb, which holds some chipset information and the size and offset in the image of the dtb
* Finally, each dtb is appended (with padding to align to page_size)

This patch makes sure to keep the appropriate dt.img format, with the necessary edits to account for size changes

References:
https://github.com/LineageOS/android_system_tools_dtbtool

### DIFF
--- a/native/jni/magiskboot/bootimg.cpp
+++ b/native/jni/magiskboot/bootimg.cpp
@@ -289,7 +289,7 @@ int unpack(const char *image, bool hdr) {
 
 	if (hdr) {
 		FILE *fp = xfopen(HEADER_FILE, "w");
-		fprintf(fp, "pagesize=%u\n", boot.hdr->page_size());
+		fprintf(fp, "page_size=%u\n", boot.hdr->page_size());
 		fprintf(fp, "name=%s\n", boot.hdr->name());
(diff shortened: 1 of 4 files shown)


topjohnwu:
I appreciate your work, but does the old devices store fstab in DTBs? If not, then the support to understand this weird format is not needed.

The only reason why DTB patching is needed in the first place is that some modern devices have their early mount partitions stored in the fstab in the device tree.

YoDevil:
Yes, I did this cause we early mount /system by having it in dts fstab. And so every device does, it was a requirement from Android 8.0, since we store the text fstab in /vendor (/system/vendor) instead of in the ramdisk

YoDevil:
Bump

Jprimero15:
Up

PR #1960, opened 2019-10-20
Added custom dialog for setting app's name after repackaging

diareuse:


app/src/main/java/com/topjohnwu/magisk/ui/settings/SettingsFragment.kt
```diff
@@ -224,26 +226,26 @@ class SettingsFragment : BasePreferenceFragment() {
...
+                values.add(locale.toLangTag())
             }
+
+            Pair(names.toTypedArray(), values.toTypedArray())
+        }.subscribeK { (names, values) ->
+            lp.isEnabled = true
+            lp.entries = names
+            lp.entryValues = values
+            lp.summary = currentLocale.getDisplayName(currentLocale)
+        }
```
diareuse: These are just formatting collisions. Make sure you're using the new code style, not the obsolete one :) 

image

PR #1964, opened 2019-10-21
magiskboot: add SPRD dt support

osm0sis:
- per https://github.com/USA-RedDragon/sprd-mkbootimg-tools/blob/master/dtbtool.c
- touch up hdr and table naming to be more uniform


osm0sis:
I think I've got everything needed here, please correct me if I missed something. 🙂

PR #1988, opened 2019-10-25
Add missing install.sh in dev guide module tree

ViRb3:


Didgeridoohan:
But the install.sh file is only part of the module installer template, not of the module itself. So, it shouldn't be included in the docs describing what the module folder structure looks like. Because it won't be there...

ViRb3:
Makes sense, I didn't realize that 😅

PR #1989, opened 2019-10-25
Manager/Boot/Init: Create Core-Only Boot Images for self-rescue of devices lacking non-OEM recovery options

adpoliak:
Work on this was started independently of https://github.com/topjohnwu/Magisk/pull/1842
There is vast similarity between these two requests, with this one differing by:

* If creating a Core-Only image, offer only to patch a boot image, not flash in place or download the ZIP file, to prevent a user from accidentally making it their primary boot image or use the zip file in a Recovery that will not accept it.
* Have MagiskInit log the fact that a Core-Only image is booted, to assist troubleshooting
* Expose Core-Only Mode for utility functions


topjohnwu:

PR #1999, opened 2019-10-27
RU translation & general strings improvements

Displax:


topjohnwu:
Please re-open a PR with the conflict fixed. The resources are changed from `stub/src/main/res/values-ru/strings.xml` to `stub/res/values-ru/strings.xml`

PR #2000, opened 2019-10-27
Update ca-strings

JoanVC100:


topjohnwu:
Please re-open a PR with the conflict fixed. The resources are changed from `stub/src/main/res/values-ca/strings.xml` to `stub/res/values-ca/strings.xml`

PR #2042, opened 2019-11-05
Very little French translation update

xerta555:


topjohnwu:
This is not needed....

Issue #2073, opened 2019-11-13
magiskboot 20+ is mangling the DTB when it patches it.


ianmacd:
Versions of `magiskboot` since early October have a fatal flaw affecting some devices:

```
[ianmacd@lagom ~]$ ls -l device/samsung/winnerx/prebuilt/recovery_dtbo
-rw-rw-r--. 1 ianmacd ianmacd 8388608 Nov  4 13:29 device/samsung/winnerx/prebuilt/recovery_dtbo
[ianmacd@lagom ~]$ ./magiskboot dtb device/samsung/winnerx/prebuilt/recovery_dtbo patch recovery_dtbo.new 2>/dev/null 
[ianmacd@lagom ~]$ ls -l recovery_dtbo.new                            
-rw-r--r--. 1 ianmacd ianmacd 2895216 Nov 13 11:44 recovery_dtbo.new
```

As you can see, two thirds of the file is being eaten.

One immediate effect of this is that Samsung Qualcomm devices will not boot after their stock recovery image is patched by Magisk. This probably applies to devices from other manufacturers, too.

One workaround is to unpack the bad image and replace the mangled `recovery_dtbo` with a copy previously patched by Magisk 19.4. When the image is repacked and subsequently flashed, it will boot as expected. That's what I'm doing until the bug in `magiskboot` is fixed.

Cheers,

Ian

cawilliamson:
I opened a bug about this a while ago but had nowhere near this level of detail.

Basically this is the exact issue I've been facing.

I found a similar workaround:

- Patch with 19.3
- Upgrade in-situ to 20.1

It would be greatly appreciated if this could be resolved though because it's a real pain in the arse for everyone on Sammy QC devices.

osm0sis:
Image for @topjohnwu to reproduce the issue with:
https://gitlab.com/osm0sis/boot-img-collection/blob/master/aosp-hdr1-avb2-recoverydtbo-seandroid-512sig-samsung_gfold-smf907b-recovery.img :+1:

ianmacd:
After further investigation, I found the bad commit to be 4d91e50d6dec76a4bdcec737f11d7547b0da34cb: `Update dtb patch to not use in-place modification`.

This adds the ability to the `magiskboot dtb patch` command to write to a new output file, rather than edit the existing file in-place.

If the file is edited in-place, there is in fact still no error. The file is correctly rewritten. It's only when the new option to write the output to a new file is used that the resulting file is invalid.

[...]

PR #2096, opened 2019-11-20
Exclude recovery DTBO from patching.

ianmacd:
See the DTBO mangling issue at https://github.com/topjohnwu/Magisk/issues/2073 for the background to this PR.

I'm submitting this just in case it will work for the generic case, i.e. not only Qualcomm-based Samsung devices. It does, at least, work for those devices, but I have no idea if all hell will break loose on other makes and models.


topjohnwu:
DTBO partition format patching is introduced in this commit (8d2198865697b3fe55474e523c473297df5bbd72) and fixes this issue

PR #2097, opened 2019-11-20
magisk: add option to reset /data/adb when there are boot issues

osm0sis:
- where --remove-modules may not be sufficient, --remove-magisk-data may, for example with the current issue introduced in 20105 where the broken magisk.db will soft-brick a device without Magisk Manager installed until /data/adb/* is removed


topjohnwu:

PR #2116, opened 2019-11-29
Fix Magisk detection by parsing ro.bootmode

fox8091:
This prevents apps from detecting Magisk when Magisk is installed to the recovery partition. In this configuration, ro.bootmode is set to recovery, meanwhile booting normally produces "unknown" in this field.


fox8091:
I should note, this primarily applies to Samsung and EMUI 9 devices, though it applies to any device that lacks a ramdisk for the boot partition.

sudoxd:
xd

fox8091:
On my Pixel 2 it shows as "unknown" by default, but I can test if "normal" works as well.

osm0sis:
@fox8091, and did "normal" work? Logically "normal" seems like it would be the more correct, standard value across devices. :wink:

Please amend your pull request commit with normal instead of unknown if it doesn't cause any issues on your Pixel.

osm0sis:
@fox8091 any follow-up?

PR #2159, opened 2019-12-16
Fix bug:magiskinit cause a crash - not closedir handle

vforkliu:
Use FrankeNDK compile magiskinit does not cause a crash,but use Stock(origin) ndk(ndk-r20)comile 
will crash。"not closedir handle"

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>Magisk</name>
+	<comment>Project Magisk created by Buildship.</comment>
+	<projects>
+	</projects>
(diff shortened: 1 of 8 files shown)


vforkliu:
[    2.271260] magiskinit: Inject magisk services: [IMymYhrsgh] [3nCntffbn3] [D3l41Bha3bCR]
[    3.117071] init: Starting service 'IMymYhrsgh'...
[   13.121353] init: Service 'IMymYhrsgh' (pid 158) killed by signal 35
whole log:[dmesg_error1.txt](https://github.com/topjohnwu/Magisk/files/3967594/dmesg_error1.txt)

PR #2184, opened 2019-12-28
Let customize.sh control the entire installation process

RikkaW:
The new module installation process uncompresses all files first. This is inappropriate in these situations (uncompress all files first is meaningless and some kind of a waste).
1. The module wants to extract specific files depending on the device (like [these](https://github.com/Magisk-Modules-Repo?utf8=%E2%9C%93&q=riru&type=&language=))
2. The module wants customized uncompress process (like [this](https://github.com/Magisk-Modules-Repo/riru_storage_redirect))

This pr changes the behavior for `customize.sh`.
* If it exists, the installer script only extracts necessary files (`module.prop` `customize.sh` `uninstall.sh`) and let `customize.sh` to do the left jobs.
* If not, the installer script extracts all files.

This does not contrary to the goal (reduces complexity) of the new installation process. For most modules which do not require `customize.sh`, nothing changes. For others, since the module dev already needs to write the script, it's okay to let them extract files themselves.

And for now, no module switched to the new format, it's safe to change it.

### DIFF
--- a/scripts/module_installer.sh
+++ b/scripts/module_installer.sh
@@ -28,6 +28,11 @@ is_legacy_script() {
   return $?
 }
 
+is_customize_script() {
+  unzip -l "$ZIPFILE" customize.sh | grep -q customize.sh
+  return $?
+}
+
 print_modname() {
   local len
   len=`echo -n $MODNAME | wc -c`
@@ -89,6 +94,8 @@ mkdir -p $MODPATH
 # Install
 ##########
 
+print_modname
+
 if is_legacy_script; then
   unzip -oj "$ZIPFILE" module.prop install.sh uninstall.sh 'common/*' -d $TMPDIR >&2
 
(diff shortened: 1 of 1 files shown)


topjohnwu:
@RikkaW one reason why I extract everything from the zip is because I wanted to force apply default file/folder permissions/selabels. At least by doing so, developers cannot "forget" to do the proper thing.

What do you think about it?

Zackptg5:
I don't think there's a need for this. I was on the same page as @RikkaW at first, but after reworking some of Unity installer, you can just as easily manipulate the extracted files by just changing paths from $TMPDIR to $MODPATH.
In the usage cases listed above, I found it easiest to just put everything into a folder, (such as $MODPATH/common), copy the appropriate files out into $MODPATH, then delete common folder.

@topjohnwu If you want to keep this in, you'll need to change `'^SKIPUNZIP='`  to `'^SKIPUNZIP=true'`

Farmernode420:
https://github.com/topjohnwu/Magisk/blob/master/scripts/module_installer.sh#L1-L2[]()

PR #2248, opened 2020-01-11
Script fixes for Android 10 + Tegra, magiskboot improvement

osm0sis:
* fix Tegra-named "APP" block /system partition mounting
* fix lzop detection use-case of `magiskboot decompress` which was broken in the generic stream refactor
* adjust mount scripts to support SOS, APP and CAC Tegra partition naming (vendor is still vendor, oddly)
* -Xnodex2oat is removed on Android 10 in AOSP (despite it still erroneously showing in `dalvikvm --help`), older devices will still run safely without it
* Android 10 dynamically linked binaries need APEX mounts and variables so add this to recovery_actions/cleanup (thanks @Zackptg5)

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -94,7 +94,7 @@ ensure_bb() {
 }
 
 recovery_actions() {
-  # Make sure random don't get blocked
+  # Make sure random won't get blocked
   mount -o bind /dev/urandom /dev/random
   # Unset library paths
   OLD_LD_LIB=$LD_LIBRARY_PATH
@@ -110,10 +110,12 @@ recovery_actions() {
 
 recovery_cleanup() {
   ui_print "- Unmounting partitions"
-  umount -l /system 2>/dev/null
-  umount -l /system_root 2>/dev/null
-  umount -l /vendor 2>/dev/null
-  umount -l /dev/random 2>/dev/null
+  (umount_apex
+  umount -l /system
+  umount -l /system_root
+  umount -l /vendor
+  umount -l /persist
+  umount -l /dev/random) 2>/dev/null
   export PATH=$OLD_PATH
   [ -z $OLD_LD_LIB ] || export LD_LIBRARY_PATH=$OLD_LD_LIB
   [ -z $OLD_LD_PRE ] || export LD_PRELOAD=$OLD_LD_PRE
@@ -166,9 +168,9 @@ mount_name() {
 mount_ro_ensure() {
   # We handle ro partitions only in recovery
   $BOOTMODE && return
-  local PART=$1$SLOT
-  local POINT=/$1
-  mount_name $PART $POINT '-o ro'
+  local PART=$1
+  local POINT=$2
+  mount_name "$PART" $POINT '-o ro'
   is_mounted $POINT || abort "! Cannot mount $POINT"
 }
 
@@ -182,7 +184,7 @@ mount_partitions() {
   [ -z $SLOT ] || ui_print "- Current boot slot: $SLOT"
 
   # Mount ro partitions
-  mount_ro_ensure system
+  mount_ro_ensure "system$SLOT app$SLOT" /system
   if [ -f /system/init.rc ]; then
     SYSTEM_ROOT=true
     [ -L /system_root ] && rm -f /system_root
@@ -193,8 +195,11 @@ mount_partitions() {
     grep ' / ' /proc/mounts | grep -qv 'rootfs' || grep -q ' /system_root ' /proc/mounts \
     && SYSTEM_ROOT=true || SYSTEM_ROOT=false
   fi
-  [ -L /system/vendor ] && mount_ro_ensure vendor
+  [ -L /system/vendor ] && mount_ro_ensure vendor$SLOT /vendor
   $SYSTEM_ROOT && ui_print "- Device is system-as-root"
+
+  # Allow /system/bin commands (dalvikvm) on Android 10+ in recovery
+  $BOOTMODE || mount_apex
 
   # Mount persist partition in recovery
   if ! $BOOTMODE && [ ! -z $PERSISTDIR ]; then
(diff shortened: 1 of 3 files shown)


vvb2060:
https://github.com/topjohnwu/Magisk/blob/984997e73b18f9fc4067e1cde1b6bfc0e639c7e4/native/jni/magiskpolicy/rules.cpp#L173 can't work?

osm0sis:
Ah, thanks. We were talking about it in the Slack like it wasn't in there, but it must have just been failing because of the magiskinit crashing bug. I'll rebase the commit out. :+1:

dianlujitao:
@osm0sis Does this work for you? On 10 `/system/bin/dalvikvm` is part of apex:
```
OnePlus3:/ # ls -al /system/bin/dalvikvm
lrwxr-xr-x 1 root shell 38 2009-01-01 00:00 /system/bin/dalvikvm -> /apex/com.android.runtime/bin/dalvikvm
```
Obviously `/apex` is not available from recovery environment, notice that `/system/bin/linker64` is part of apex too, which means all system executables are not usable too even if they're not part of apex.

The detection just failed to run:
https://github.com/topjohnwu/Magisk/blob/984997e73b18f9fc4067e1cde1b6bfc0e639c7e4/scripts/util_functions.sh#L304

osm0sis:
It works while booted with my changes. From recovery is still an issue. I plan to bring over the /apex fixes from AK3 but I didn't want to complicate this PR any more than it already is.

scripts/util_functions.sh
```diff
@@ -203,12 +206,66 @@ mount_partitions() {
...
     if ! is_mounted /persist; then
       # Fallback to cache
-      mount_name cache /cache
+      mount_name "cache cac" /cache
       is_mounted /cache && PERSISTDIR=/cache || PERSISTDIR=
     fi
   fi
 }
 
+mount_apex() {
```
dianlujitao: Skip mount if apex is already available

scripts/util_functions.sh
```diff
@@ -203,12 +206,66 @@ mount_partitions() {
...
+      ;;
+      *) mount -o bind $APEX $DEST;;
+    esac
+  done
+  export ANDROID_RUNTIME_ROOT=/apex/com.android.runtime
+  export ANDROID_TZDATA_ROOT=/apex/com.android.tzdata
+  export BOOTCLASSPATH=/apex/com.android.runtime/javalib/core-oj.jar:/apex/com.android.runtime/javalib/core-libart.jar:/apex/com.android.runtime/javalib/okhttp.jar:/apex/com.android.runtime/javalib/bouncycastle.jar:/apex/com.android.runtime/javalib/apache-xml.jar:/system/framework/framework.jar:/system/framework/ext.jar:/system/framework/telephony-common.jar:/system/framework/voip-common.jar:/system/framework/ims-common.jar:/system/framework/android.test.base.jar:/apex/com.android.conscrypt/javalib/conscrypt.jar:/apex/com.android.media/javalib/updatable-media.jar
+}
+
+umount_apex() {
```
dianlujitao: Skip umount if not mounted by us

scripts/util_functions.sh
```diff
@@ -203,12 +206,66 @@ mount_partitions() {
...
+            losetup -d $LOOP 2>/dev/null
+          fi
+        fi
+      ;;
+      *) mount -o bind $APEX $DEST;;
+    esac
+  done
+  export ANDROID_RUNTIME_ROOT=/apex/com.android.runtime
+  export ANDROID_TZDATA_ROOT=/apex/com.android.tzdata
+  export BOOTCLASSPATH=/apex/com.android.runtime/javalib/core-oj.jar:/apex/com.android.runtime/javalib/core-libart.jar:/apex/com.android.runtime/javalib/okhttp.jar:/apex/com.android.runtime/javalib/bouncycastle.jar:/apex/com.android.runtime/javalib/apache-xml.jar:/system/framework/framework.jar:/system/framework/ext.jar:/system/framework/telephony-common.jar:/system/framework/voip-common.jar:/system/framework/ims-common.jar:/system/framework/android.test.base.jar:/apex/com.android.conscrypt/javalib/conscrypt.jar:/apex/com.android.media/javalib/updatable-media.jar
```
dianlujitao: com.android.runtime and com.android.conscrypt jars are enough, the rest can be removed

scripts/util_functions.sh
osm0sis: It only runs in recovery, where apex is not available, We are what makes it available.

scripts/util_functions.sh
osm0sis: See above. Nothing but us knows how to mount it.

scripts/util_functions.sh
osm0sis: You're wrong actually, tzdata and the TZDATA variable are absolutely required. I've tested this myself, directly, on a real device. Have you?

scripts/util_functions.sh
dianlujitao: It's available with my changes on twrp gerrit, I've tested on my device

scripts/util_functions.sh
dianlujitao: I'm only talking on BOOTCLASSPATH, ANDROID_RUNTIME_ROOT and ANDROID_TZDATA_ROOT are needed. I didn't highlight them but github...

osm0sis:
> blank line between title and body of commit "scripts: fixes for Tegra partition naming + bootsigner on Android 10" is missing

Wow. Stop being ridiculous.

dianlujitao:
> > blank line between title and body of commit "scripts: fixes for Tegra partition naming + bootsigner on Android 10" is missing
> 
> Wow. Stop being ridiculous.

Seriously?
image
image

scripts/util_functions.sh
osm0sis: It's just a path variable, and it's the one a booted ROM would have. No harm in leaving them in to leave the use-case of the script broad.

osm0sis:
> > > blank line between title and body of commit "scripts: fixes for Tegra partition naming + bootsigner on Android 10" is missing
> > 
> > 
> > Wow. Stop being ridiculous.
> 
> Seriously?
> image
> image

Not sure what you or your device is smoking, look at either of these - there is no line:
https://github.com/topjohnwu/Magisk/commit/ed8faa5a7adf3d5a41016d76aefbf31f8def7a1b
https://github.com/topjohnwu/Magisk/commit/ed8faa5a7adf3d5a41016d76aefbf31f8def7a1b.patch

dianlujitao:
> > > > blank line between title and body of commit "scripts: fixes for Tegra partition naming + bootsigner on Android 10" is missing
> > > 
> > > 
> > > Wow. Stop being ridiculous.
> > 
> > 
> > Seriously?
> > image
> > image
> 
> Not sure what you or your device is smoking, look at either of these - there is no line: [ed8faa5](https://github.com/topjohnwu/Magisk/commit/ed8faa5a7adf3d5a41016d76aefbf31f8def7a1b)
> https://github.com/topjohnwu/Magisk/commit/ed8faa5a7adf3d5a41016d76aefbf31f8def7a1b.patch

take a look at [a random sane commit](https://github.com/torvalds/linux/commit/f5ae2ea6347a308cfe91f53b53682ce635497d0d.patch), notice the blank line below `Subject` line.
If you use editors such as vim it'll highlight for you.

osm0sis:
> > > > > blank line between title and body of commit "scripts: fixes for Tegra partition naming + bootsigner on Android 10" is missing
> > > > 
> > > > 
> > > > Wow. Stop being ridiculous.
> take a look at [a random sane commit](https://github.com/torvalds/linux/commit/f5ae2ea6347a308cfe91f53b53682ce635497d0d.patch), notice the blank line below `Subject` line.
> If you use editors such as vim it'll highlight for you.

Cool. Neat that it does that. That's not a requirement for a git commit.

scripts/util_functions.sh
osm0sis: Do you already have them in production on a device's TWRP or are they just for the next release if they get merged?

dianlujitao: I'm not bigbiff or Dees-Troy so I can't tell if they'll get merged into TWRP, but it's possible. Anyway my implementation actually works with your fix before you push mount_apex stuff, and adding such exception is trivial and harmless.

osm0sis: I just checked your implementation and it seems to rely on symlinks, which won't work for APEX on many ROMs on most devices, where the APEX files are within ext4-dedup apex_payload.img files within .apex files (as you can see my script also accommodates).

I'll add a check for your apex symlinks so we can still set things up properly if they're found.

dianlujitao: Yea you're right, you remind me of upgradable apex is a thing, it's hard to keep track of it on twrp side in this case because system can be mount/umounted anytime thus mount apex on twrp start is not reliable.

dianlujitao:
@osm0sis mount_apex is not executed from addon.d on my device thus
```
/data/adb/magisk/addon.d.sh: /system/bin/dalvikvm: not found
```
can you verify?

osm0sis:
This hasn't been merged yet, but once it is mount_apex will be executed as part of both normal zip install and addon.d since it's added under the recovery_actions function.

dianlujitao:
I guess I know what's wrong here.

On nornal installation `recovery_actions` is executed after `mount_partitions`
https://github.com/topjohnwu/Magisk/blob/984997e73b18f9fc4067e1cde1b6bfc0e639c7e4/scripts/flash_script.sh#L45
https://github.com/topjohnwu/Magisk/blob/984997e73b18f9fc4067e1cde1b6bfc0e639c7e4/scripts/flash_script.sh#L105

but on addon.d it's not:
https://github.com/topjohnwu/Magisk/blob/984997e73b18f9fc4067e1cde1b6bfc0e639c7e4/scripts/addon.d.sh#L51-L57

thus mount_apex failed to detect /system/apex

osm0sis:
Good catch! I'll revise that in addon.d and add to this PR. Thanks! :+1:

osm0sis:
I think the idea was that with the zip installer there's already a self-contained environment commands are being run within, but with addon.d this isn't the case until recovery_actions is run, so we run it first so that our busybox is used, etc.

So I just added `mount_apex` only in recovery directly to mount_partitions and it should work for both cases from there.

dianlujitao:
The new impl works fine, thanks

thekiefs:
Any chance this can be merged so I can test it?

osm0sis:
Waiting on @topjohnwu

Sazpaimon:
Building with this PR allows Magisk to work on the port of Pie to the Nintendo Switch, as that also uses Tegra partition names.

osm0sis:
@thekiefs, it's in 20304 Canary. :+1:

thekiefs:
@topjohnwu @osm0sis looks like it worked! Awesome job, and thank you! 👍 

Now I just need to try patching and flashing 8.0.2 once they release the images. 👀

PR #2279, opened 2020-01-17
Add a function to get API level for Realme devices

mauronofrio:
Realme devices has not a "ro.build.version.sdk" in build.prop 
but it is in build_default.prop, so if the normal function fails
there will be another one that will try to get the sdk version.


mauronofrio:
Nothing?

PR #2284, opened 2020-01-18
Restore /system symlink on clean up

dianlujitao:
Don't break existing environment.


osm0sis:
Just occurred to me `rm -rf /system` is kind of a scary thing to point at /system which may or may-not be fully unmounted due to `umount -l` (lazy unmount), so I'd suggest replacing that with `rmdir /system` to be safe. :+1:

dianlujitao:
https://github.com/topjohnwu/Magisk/pull/2402

PR #2297, opened 2020-01-22
Added clarification to sample config.props

ajayyy:
This might save people like me some headaches.


osm0sis:
Documentation was fully reworked.

PR #2304, opened 2020-01-23
Add ianmacd theme to Magisk Manager for better contrast.

ianmacd:
This theme provides higher contrast caption text, as well as a soothing green primary colour.


ianmacd:
Merged into a new pull request, #2316.

PR #2329, opened 2020-01-26
Translation Correction

dark-basic:
For a mistake of mine, forget to translate these lines, also correct a stress. And another change.


dark-basic:
--- **VIEW ME** 
What I also want to add, to see with the design, is that some devices where the screen is less wide happen that the whole title does not fit in the grid.

I attach an image
https://fotos.subefotos.com/c4e55a34c5b290b8ea4d79c2f1ff8e39o.png

Another important notice. Some translations are not being applied, even if the translation exists in the code. 
I have verified my translations, and they exist, they are translated, but these are not being applied. (It is in the part of the waiting time to deny and the super user toast)

PR #2349, opened 2020-01-28
Updated Flash / SuRequest activities with app themes

topjohnwu:
CompatActivity/Fragment logic has been moved to respective BaseUI. Some deprecated and unused styles have been removed in favor or newer themes.


awtaaar4123:
i have proplem in safetynet

PR #2354, opened 2020-01-29
The long press to unlock is a recovery menu

RandomReaper:
Tested on my s10e


topjohnwu:
No it's not. You need to boot into download mode to unlock bootloader

PR #2433, opened 2020-02-10
Update UK strings

tarasyyyk:
Refactored resource strings according to English source.


tarasyyyk:
@topjohnwu is it make sense to wait when new design will be ready and after that to translate strings?

PR #2436, opened 2020-02-10
Fix various prop hide issues in MagiskHide

Fox2Code:
By checking props on @Fernixx Samsung phone I noticed that 
`ro.vendor.boot.warranty_bit` has the value `1`
So I made this patch to improve MagiskHide on some Samsung phones
(Including my friend phone)


osm0sis:
Can you add the prop and value I mentioned from #2116 too, so both can be closed?

Fox2Code:
Ok I will merge #2116 into my branch

Fox2Code:
Done!
I let `unknown` as prop value because it seem more widespread than `normal`
On my phone _(LineageOS Pie)_ and on @Fernixx phone _(Stock OneUI Pie)_ the props are set to `unknown`

osm0sis:
Hmm I guess we'll see if setting them to unknown has any effect on others

Fox2Code:
Maybe changing the beaviour of props patching may be a solution

topjohnwu:
@osm0sis on Pixel, it seems to be unknown

Fox2Code:
Maybee changing the behavior of the prop patching may be a solution.
If for example `normal` or `unknown` it does nothing but if something else it detected it set the property to `unknown`

Fox2Code:
Made a patch to handle the case where `ro.bootmode` or `ro.boot.mode` is set to `normal`
Since it should not affect the detection if the prop is `normal` rather than `unknown`
But I rather prefer setting the prop to `unknown` in other cases since it appear to be more common across many devices

osm0sis:
Looks like maybe a better way to write the check for unknown and normal would be as they do in this PR?

https://github.com/topjohnwu/Magisk/pull/2470/files

Fox2Code:
Yup I will change that and put value to `unknown` if `recovery` is detected

Fox2Code:
Since we are no longer talking about `.warranty_bit` props I renamed the discussion

osm0sis:
Looks good! Can you squash it all to one commit and add "Closes #2116" to your commit message to clean it up for @topjohnwu to merge it? :+1:

Fox2Code:
Ok I mess up my local branch by trying to squash all commits together and made everything worse (First time I try to squash commits)
I will try again tomorrow but I promise nothing.

Fox2Code:
:tada: Yeah I finally succeed to squash it all :tada: 
At least now I know how to make it
The problem was when I tried to squash It would take #2116 branch instead of master Magisk branch
I successfully find a workaround for the issue so it's fine now

osm0sis:
Great! My only nitpick is the main commit name of "Squash commits" doesn't really explain what it does anymore. It should ideally be something more like your PR name, like "MagiskHide prop additions".

Otherwise, good to go for merge, @topjohnwu!

Fox2Code:
Renamed the commit (The name it a bit longer because I put everything in title instead of description.)

PR #2454, opened 2020-02-12
Add specific strings for Home + minor changes for Modules

sn0ooow:
In previous versions of Manager, we had specific strings for "Installed version" and "Latest version". Unfortunately, not only they were removed, but in Home section we re-use "Installed" and "Remote" from Modules section. It's confusing, but not only confusing, there are languages that don't work like English, some languages need singular and plural forms. For instance in Romanian case, "Instalate" in Modules section and "Instalat" in Home section ("Instalate" is the plural form and the other is the singular form). Other languages like French might need this as well. And now it's more easier to understand the context ("current_installed_version", "current_remote_version", "module_section_installed", "module_section_updates", "module_section_action_update_all").


topjohnwu:
@sn0ooow thanks for the info, the home screen text are not final yet, but will leave this issue open so I can keep aware of this issue

PR #2457, opened 2020-02-13
Disable Volumes keys in FlashActivity

Fox2Code:
Sometimes modules ask to push volumes keys but the problem is that by pushing the volume keys during a flash it also changes the volume of the phone.

And I found a way to block volume keys so the volume doesn't change when flashing a module and pressing volume keys.


Fox2Code:
Will fix conflict later and try a better way to implement it
Until then do not merge

osm0sis:
It's been a month, so we'll close this for now, until you can rework. Ping me to reopen when you're ready. :+1:

PR #2462, opened 2020-02-15
Add a Chinese traslation.

zjt2005:
Add a link to Chinese translation of the "samsung-system-as-root" part.

增加了一个中文翻译版的链接。(文章在我的博客里)


zjt2005:
Eh, did I tap something just now?

osm0sis:
Good idea, and thanks, but I don't think linking externally will be the way to do it.

Issue #1905, opened 2019-10-08
Support Q recovery


YoDevil:
In Q aosp recovery uses /system to store recovery binaries and configs (/system/bin, /system/etc).
It expects system to be mounted to /mnt/system instead, and likewise vendor to /mnt/vendor.

Magisk install script is incompatible since it tries mounting to /system

topjohnwu:
What is incompatible about that though? Magisk will directly mount the raw device block to where it should be (`/system`, `/vendor`), and it does not need anything from the recovery as all tools required is all bundled with the installer.

YoDevil:
Well you can't mount on /system when /system contains recovery itself

topjohnwu:
Why? It's just a folder, and any folder can be used as a mountpoint.

YoDevil:
I believe it is caused by the mounting, but i don't know the reason:
```
[   46.924898] Archive:  /sideload/package.zip
[   46.924943]   inflating: META-INF/com/google/android/update-binary
[   46.924958]   inflating: META-INF/com/google/android/updater-script
[   46.924973]   inflating: arm/magiskinit
[   46.924988]   inflating: arm/magiskinit64
[   46.925002]   inflating: arm/magiskboot
[   46.925016]   inflating: x86/magiskinit
[   46.925030]   inflating: x86/magiskinit64
[   46.925044]   inflating: x86/magiskboot
[   46.925058]   inflating: common/magisk.apk
[   46.925072]   inflating: common/boot_patch.sh
[   46.925086]   inflating: common/util_functions.sh
[   46.925100]   inflating: common/addon.d.sh
[   46.925114]   inflating: chromeos/futility
[   46.925139]   inflating: chromeos/kernel_data_key.vbprivk
[   46.925153]   inflating: chromeos/kernel.keyblock
[   47.079760] Segmentation fault
[   47.232974] mount: mounting /system on /system_root failed: Invalid argument
[   47.236984] mount: mounting /system_root/system on /system failed: No such file or directory
[   50.386796] mount: can't read '/etc/fstab': Too many symbolic links encountered
[   50.691702] E:Error in /sideload/package.zip (status 1)
```
Notice the Segmentation fault.

I can try providing a snapshot of the Lineage 17.0 recovery (based on aosp Q recovery) for HTC M8 if you're willing to look into it.

CaptainThrowback:
@YoDevil

[...]

PR #2496, opened 2020-02-29
Fixup mounting system on Lineage Recovery

aleasto:
* Lineage Recovery 17.1, like AOSP Q recovery, has '/' as a shared
   mount point, causing `mount --move` to fail.
   If it fails, directly mount system to /system_root via
   the /dev/block/bootdevice symlink, like AnyKernel and OpenGapps

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -209,6 +209,11 @@ mount_partitions() {
     SYSTEM_ROOT=true
     setup_mntpoint /system_root
     mount --move /system /system_root
+    if [ $? != 0 ]; then
+      umount /system
+      umount -l /system 2>/dev/null
+      mount_ro_ensure "system$SLOT app$SLOT" /system_root
+    fi
     mount -o bind /system_root/system /system
   else
     grep ' / ' /proc/mounts | grep -qv 'rootfs' || grep -q ' /system_root ' /proc/mounts \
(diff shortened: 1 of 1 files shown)


YoDevil:
I finally got around to debug #1905 .
Weirdly your busybox just segfaults rather than gracefully exiting with error, but ultimately this was the cause of failed installation.

scripts/util_functions.sh
```diff
@@ -209,6 +209,11 @@ mount_partitions() {
     SYSTEM_ROOT=true
     setup_mntpoint /system_root
     mount --move /system /system_root
+    if [ $? != 0 ]; then
+      umount /system
+      umount -l /system 2>/dev/null
+      mount -o ro -t auto /dev/block/bootdevice/by-name/system$SLOT /system_root
```
PeterCxy: This will fail on devices with dynamic partitions (it failed on my Redmi K30 5G). Should use `mount_ro_ensure "system$SLOT app$SLOT" /system` just like how it mounts `/system` in the first place.

osm0sis:
You're right, but to follow the logic of the script it would need to be `mount_ro_ensure "system$SLOT app$SLOT" /system_root`

@YoDevil can you fix that and force push the commit again?

YoDevil:
That is quite a neat helper!
Verified that it still works on my device.

I'm a tad confused though. Was the installer working on Lineage Recovery before this patch for you @PeterCxy ?

PeterCxy:
@YoDevil Nope, it was not working, due to the same reason why you wrote this patch. Your patch technically fixed the problem, but the only problem was that the block device path is wrong (on dynamic partition devices, they will be `/dev/block/dm-*`), and that's why I posted that comment.

YoDevil:
Ah ok! Perfect then.

PR #2509, opened 2020-03-02
Add linebrake notice for module.prop

GaryOderNichts:
This made some trouble when creating a module.prop on Windows. The file could not be read properly by magisk manager and my module folder had an \r at the end which made it unremovable through Magisk Manager.


docs/guides.md
```diff
@@ -70,6 +70,7 @@ ex: ✓ `a_module`, ✓ `a.module`, ✓ `module-101`, ✗ `a module`, ✗ `1_mod
 This is the **unique identifier** of your module. You should not change it once published.
 - `versionCode` has to be an **integer**. This is used to compare versions
 - Others that weren't mentioned above can be any **single line** string.
+- Make sure to use the `UNIX (LF)` line brake type and not the `Windows (CR+LF)` or `Macintosh (CR)` one.
```
MediumOne: I think this is a typo - please correct "line brake" to "line break".

GaryOderNichts: Yup that's a typo. Fixed it

Issue #2512, opened 2020-03-03
Add a hint to the Update README that you have to "unroot" before you can see the OTA update


RAYs3T:
On my Pixel 4 XL i have to do the "uninstall" via Magisk first, in order to see that a new OTA is available.

Otherwise it keeps saying "System is up-to-date".

We should add this as a small hint in the upgrade guide

a447948814:
must be Brush the whole package.

PR #2513, opened 2020-03-03
Add hint: OTA updates not shown until magisk is uninstalled

RAYs3T:
On some devices the OTA update check is reporting "System is up-to-date" even when a new OTA is available. You have to uninstall Magisk first to see the new update.

#2512

### DIFF
--- a/docs/tutorials.md
+++ b/docs/tutorials.md
@@ -3,6 +3,8 @@
 ## OTA Installation
 Magisk does modifications systemless-ly, which means applying official OTAs is much simpler. Here I provide a few tutorials for several different kind of devices to apply OTAs and preserve Magisk after the installation if possible.
 
+**NOTE: On some devices a new OTA update might not be shown as "available" until you uninstall Magisk and the updater says "System is up-to-date"**
+
 **NOTE: In order to apply OTAs, you HAVE to make sure you haven't modified `/system` (and `/vendor` if available) in any way. Even remounting the partition to rw will tamper block verification!!**
 
 ### Prerequisites


osm0sis:
You would be better off looking at what package needs to be added to the MagiskHide list by default to avoid the issue you describe.

linuxandria:
@osm0sis I'd recommend against that as it may in fact brick your device if it gives you an incremental update. Standard OTA is generally blocked when rooted for good reason. I know on OnePlus devices it'll download the full OTA instead but others restoring stock then updating is probably safest

RAYs3T:
> I suggest maybe saying "On some devices if the proper service isn't magisk hidden then you may have to temporarily unroot

My understanding is that the OTA update dialog is embedded in the system and therefore not listed in hidelist (since it is not a standalone application). If this is the case it would be the only way to unroot first in order to update.

If it is available in the hidelist, should we gather some "examples" of the update dialog package names and suggest to add them in the installation guide?

linuxandria:
@RAYs3T pixels it's done through Google play services, Motorola and OnePlus devices all have a dedicated application (OnePlus is called OPBackup)

But see https://github.com/topjohnwu/Magisk/pull/2513#issuecomment-594398687

scottismyname:
I'm a bit confused....on my Pixel 2XL and Pixel 3 I can't update.  Google play services is already hidden....do additional things need to be hidden?  I tried doing an uninstall (restore images) and re-checking, but it still says my system is up to date.

bell18-web:
@bell18-web

PR #2517, opened 2020-03-04
addon.d: fix passing args to the script in /data

aleasto:
Fixing up on https://github.com/topjohnwu/Magisk/commit/b3bc1a3907be0ba3b67872bb508ea64ac8bdc90b

### DIFF
--- a/scripts/addon.d.sh
+++ b/scripts/addon.d.sh
@@ -7,10 +7,12 @@
 #
 ########################################################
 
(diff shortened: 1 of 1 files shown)


aleasto:
Could call `trampoline "$@"` instead, but i think this should be better future-proof

osm0sis:
Already covered in #2456, please close.

PR #2549, opened 2020-03-15
fix standalone shell exec

aleasto:
The checks for standalone were not quite working for me.
`[ -o option_name ]` seems to not be a thing in ash at all, and `set -o standalone 2>/dev/null && return` returned on my mksh too. So in https://github.com/topjohnwu/ndk-busybox/pull/3 i added a safe way to detect the standalone mode.

Also the exec was being called with no arguments cause "$@" was used inside a function.
So i routed "$@" through the function calls. I hope I did not leave any place behind.

Please check app/src/main/res/raw/util_functions.sh as i'm not sure what calls install_module() there, but that needs to pass "$@" through too

### DIFF
--- a/scripts/addon.d.sh
+++ b/scripts/addon.d.sh
@@ -60,7 +60,7 @@ initialize() {
     ui_print() { log -t Magisk -- "$1"; }
   fi
   OUTFD=
-  setup_flashable
+  setup_flashable "$@"
 }
 
 main() {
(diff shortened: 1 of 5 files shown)


osm0sis:
Any more elegant way to do this? Passing "$@" everywhere seems like a lot.

I'm thinking rewriting https://github.com/topjohnwu/Magisk/commit/db9a4b31f9b204c2951acfe99064bbe90096c2ca will be the way to go, since that's what broke addon.d, etc.

aleasto:
I certainly agree.

PR #2566, opened 2020-03-21
Add Python3.5 support to build.py

AleixDev:
I know that f-Strings are great, but in this case I think that it's more beneficial to use the "old" style formatting, since it could help to solve this [issue](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/4401?diff_id=33206225&start_sha=0b212b195570cf820442fe29300d4b0813467a8c#note_139227353) (about the inclusion of Magisk to the F-Droid repo).

It would be awesome to have this app on the F-Droid repo, so I hope that it is okay for you to accept this request :)

### DIFF
--- a/build.py
+++ b/build.py
@@ -24,8 +24,8 @@ def vprint(str):
 
 
 # Environment checks
-if not sys.version_info >= (3, 6):
-    error('Requires Python 3.6+')
+if not sys.version_info >= (3, 5):
+    error('Requires Python 3.5+')
 
 if 'ANDROID_HOME' not in os.environ:
     error('Please add Android SDK path to ANDROID_HOME environment variable!')
@@ -67,7 +67,7 @@ STDOUT = None
 def mv(source, target):
     try:
         shutil.move(source, target)
-        vprint(f'mv: {source} -> {target}')
+        vprint('mv: {} -> {}'.format(source, target))
     except:
         pass
 
@@ -75,7 +75,7 @@ def mv(source, target):
 def cp(source, target):
     try:
         shutil.copyfile(source, target)
-        vprint(f'cp: {source} -> {target}')
+        vprint('cp: {} -> {}'.format(source, target))
     except:
         pass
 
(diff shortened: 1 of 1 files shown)


topjohnwu:
There are currently no plans for releasing Magisk Manager on F-droid as the app has its own way to distribute updates. The app and native zips work toe to toe, so tight control over the update flow is required.

And also, Python 3.8 is almost 0.5 years old, I don't think requiring Python 3.6 is asking too much.
But anyways, thanks for the PR!

AleixDev:
>  But anyways, thanks for the PR!

Thank you for taking time to respond it :)

> And also, Python 3.8 is almost 0.5 years old, I don't think requiring Python 3.6 is asking too much.

I know and it's very understandable that you prefer a recent Python version, they are having problems upgrading Python, but I hope they will find the way to solve them soon.

> There are currently no plans for releasing Magisk Manager on F-droid as the app has its own way to distribute updates. The app and native zips work toe to toe, so tight control over the update flow is required.

Since F-Droid developers and you are against the inclusion of your app to the official F-Droid repo, makes me think that I can be wrong, but I just want to explain my point of including it:
Publishing Magisk Manager to F-Droid doesn't limit your control over the update flow, the app can continue to have its own way to distribute updates and F-Droid can be used just to get the app from another source. This way all people that trust F-Droid official repo can install this app knowing that it doesn't have any binary blob (without having to manually check the source code or trust you directly). And this can be beneficial for you since it can make your app a little bit more popular.

Iey4iej3:
It looks like a versioning issue to keep the flashed zip and the manager compatible. Note that [Xposed Installer](https://f-droid.org/en/packages/de.robv.android.xposed.installer/) also only gave an installer to download the manager and the flashing zip.

rootkea:
I second @AleixDev.

The only reason I have not installed Magisk yet is because it's not available in F-droid. I only use FLOSS apps on my device.

I dearly want Magisk in F-droid. Hoping to start the inclusion process again https://gitlab.com/fdroid/fdroiddata/-/merge_requests/4401#note_473732270 :)

PR #2579, opened 2020-03-23
Feature/log as primary

dodo887:


osm0sis:
@diareuse will open a PR himself when he's ready.

PR #2582, opened 2020-03-23
Added clarification to sample config.props

ajayyy:
Reworked version of https://github.com/topjohnwu/Magisk/pull/2297


topjohnwu:
Should be addressed in this commit: f6a2b1c88255b1027ae3ff1b72db889d39cb8fd7

PR #2597, opened 2020-03-28
Manager: remove v from download names

osm0sis:
- now that Canaries are only commit hashes for the version string, the v is unnecessary/confusing, so the simplest solution is to just remove the v from the filename for all Manager-based downloads of Magisk and Manager


osm0sis:
Requesting @diareuse also, just to double check there aren't any other locations I missed or might need adjustment after these changes. :+1:

diareuse:
Oh I see, let me see whether it's hardcoded somewhere else

PR #2625, opened 2020-03-30
Add secure folder creation in boot_complete

sabpprook:
On Oneplus 7T HD1900 OOS 10.0.8.HD65AA after userdata wiped, device will get bootloop because secure folder not created in late_start section, do it again under boot_complete will fix it.

### DIFF
--- a/native/jni/core/bootstages.cpp
+++ b/native/jni/core/bootstages.cpp
@@ -722,6 +722,9 @@ void late_start(int client) {
 		// It's safe to create the folder at this point if the system didn't create it
 		if (access(SECURE_DIR, F_OK) != 0)
 			xmkdir(SECURE_DIR, 0700);
+		// Return if folder not created and do it again in boot_complete section
+		if (access(SECURE_DIR, F_OK) != 0)
+			return;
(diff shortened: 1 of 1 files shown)


osm0sis:
@topjohnwu, will the rewrite for Android 11 address this?

osm0sis:
@sabpprook is this resolved?

sabpprook:
@osm0sis OP7T is my daily drive phone, I'll make the test with latest commit once I get free.

osm0sis:
Great, I'm going to open this again.

Please close again if it is, in fact, resolved, but otherwise it's good to leave it open as a reminder of the issue. :+1:

topjohnwu:
@osm0sis this is not a proper solution. Boot completed would happen AFTER the user unlocked the device, which mean if we create the folder at that point and reboot, it will be an awkward user experience. A potential fix would be allow Magisk Manager to detect this situation and prompt the user to reboot manually.

Will keep this PR open until a proper solution is created

topjohnwu:
Addressed in this commit: fc1844b4dff14d589626dd770c64d5da892e1e0c

PR #2653, opened 2020-04-05
Hungarian translation

gidano:
Hi,
I've made the Hungarian translation.
Unfortunately I couldn't find any other way to submit, I hope you will be in the right place here...?!

Sincerelly,
gidano

[MM.Hungarian.translation.strings.zip](https://github.com/topjohnwu/Magisk/files/4434055/MM.Hungarian.translation.strings.zip)


topjohnwu:
Please create a proper PR, not uploading a zip file

PR #2669, opened 2020-04-10
Added Hungarian Translation

gidano:
Added Hungarian Translation


gidano:
Added Hungarian Translation
I really don't understand how to handle GitHub, so I hope it's okay ?!

PR #2685, opened 2020-04-13
fix Magisk install on ODROID-N2/C4

voodik:
This prevent install on some devices, in this case its Odroid dev boards
First it's a dtb block device node name, in Odroids it's a  /dev/block/dtbs
Next patched dtb does not flash with cat (insufficient space)
So i think dd is better option to write binary data.

Regards.


osm0sis:
The error from cat is just from /dev/zero, so it actually works fine. Don't change that.

osm0sis:
Thanks, looks good now!

PR #2709, opened 2020-04-19
RECOVERYMODE should not always be overridden on legacy SaR devices.

ianmacd:
When the device wasn't booted from recovery, it's presumably already rooted in the boot partition (e.g. Samsung Fold and Tab S6), so we should not force RECOVERYMODE to true.

Without this, Magisk Manager will erroneously install Magisk in the recovery partition of boot-rooted legacy SaR devices.

This sacrifices blanket detection of legacy SaR devices in favour of allowing boot-rooted devices to have Magisk updated by MM, rather than having to flash them in TWRP.

### DIFF
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -129,7 +129,7 @@ check_boot_ramdisk() {
   $ISAB && return 0
 
   # If we are using legacy SAR, but not AB, we do not have ramdisk in boot
-  if grep ' / ' /proc/mounts | grep -q '/dev/root'; then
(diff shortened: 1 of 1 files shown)


fox8091:
Fair warning, we spoof the bootmode now (#2436) because it was a possible detection method before.

marign25:
Root

tsynik:
I'm too vote for this. On Magisk-enabled Android TV-box (9.0 SaR) Magisk detect stock kernel and tries install to recovery and broke it.

tsynik (approved):
+1

topjohnwu:
@ianmacd I addressed this issue in ca9f9fee9a9197b6a56f4bf5ed3d4ae26a7a6dd7

PR #2711, opened 2020-04-20
scripts: module and installation fixes

### LINKED ISSUE #2720: Latest Magisk canary fails to install on Fire HD 8 2018
The issue that the Magisk Installer detects the Boot Image on /dev/log/kernel and it should be saying /dev/block/mmcblk0p11 as a target boot image.
If you need me to attach more logs or files please let me know. The TWRP Version is 3.3.1-0
[recovery.log](https://github.com/topjohnwu/Magisk/files/4506902/recovery.log)
the recovery fstab is 
[recovery.fstab.txt](https://github.com/topjohnwu/Magisk/files/4506916/recovery.fstab.txt)

osm0sis:
**scripts: ensure system is able to be mounted rw before attempting**
- this is needed for installations on Lineage 17.1 Recovery (AOSP Q) for logical partition devices, which uses /dev/block/mapper to stage the partitions

Thanks LuK1337 & erfanoabdi @ Lineage

**scripts: add author name back to module install banner print**
- print_title can now do up to 2 lines

**scripts: fix find_block false positives /dev/log/kernel and /dev/BOOT**
- try /dev/block first with full depth to catch all platform/soc variations to the by-name directory, and the new dynamic partition /dev/block/mapper
- next try uevent for block devices as before
- lastly try /dev with maxdepth 1 (immediate directory) to find /dev/bootimg, /dev/recovery, etc. while avoiding /dev/log/kernel
- move bootimg higher in the list than boot so /dev/bootimg gets found first and avoids /dev/BOOT
- recovery_a/_b now also exists
- minor touch-ups for readability and consistency

Fixes #2720

### DIFF
--- a/scripts/flash_script.sh
+++ b/scripts/flash_script.sh
@@ -78,6 +78,7 @@ chmod -R 755 $MAGISKBIN
 # addon.d
 if [ -d /system/addon.d ]; then
   ui_print "- Adding addon.d survival script"
+  blockdev --setrw /dev/block/mapper/system$SLOT 2>/dev/null
   mount -o rw,remount /system
   ADDOND=/system/addon.d/99-magisk.sh
   cp -af $COMMONDIR/addon.d.sh $ADDOND
--- a/scripts/module_installer.sh
+++ b/scripts/module_installer.sh
@@ -48,12 +48,15 @@ is_legacy_script() {
 }
 
 print_modname() {
-  local len
-  len=`echo -n $MODNAME | wc -c`
+  local authlen len namelen pounds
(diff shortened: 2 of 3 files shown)


fox8091:
Maybe gate that with a check to make sure the file actually exists? blockdev may fail gracefully, but it seems cleaner to gate it behind a file check.

osm0sis:
> Maybe gate that with a check to make sure the file actually exists? blockdev may fail gracefully, but it seems cleaner to gate it behind a file check.

I masked the stderr, so it's fine.

PR #2757, opened 2020-05-01
scripts: try to preserve /etc/fstab on AOSP Q-based recovery

osm0sis:
- busybox mount using /etc/fstab would break on Lineage 17.1 since /etc is a symlink to /system/etc and Magisk uses /system as a mountpoint
- work around it for now by temporarily copying /system/etc/* to /etc/* in the rootfs, and once TWRP has finalized their Android 10+ recovery setup the mount scripts will need to be gutted and mostly rewritten

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -140,6 +140,11 @@ recovery_actions() {
   unset LD_LIBRARY_PATH
   unset LD_PRELOAD
   unset LD_CONFIG_FILE
+  # Try to preserve /etc/fstab on AOSP Q-based recovery
+  if [ -L /etc ]; then
+    setup_mntpoint /etc
+    cp -af /etc_link/* /etc
+  fi
 }
 
 recovery_cleanup() {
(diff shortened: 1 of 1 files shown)


topjohnwu:
The only thing we actually depend on the recovery for mounting is `/data` because decryption has to be involved. For consistent recovery installation, my opinion would be **DON'T** try to mount `/data` ourselves, and only access `/data` if the recovery already had it mounted for us.

For all other partitions (e.g. system/vendor/cache etc), we should just directly mount things ourselves and should not depend on any fstab files **AT ALL**.

This mean our end goal is to eliminate **ALL** `mount <mount_point>` everywhere in our scripts and always do `mount <dev> <dir>` to mount partitions to our own path. The inconsistency of custom recoveries is driving me crazy at this point, I do not want my installation script have ANYTHING to do with custom recovery at all, so we should roll our own environment ourselves from scratch.

osm0sis:
Problem that was brought to my attention is that lots of devices are using partitions named other things as /vendor, such as oem, cust or even cache, so preserving fstab would fix this.

@aleasto and others are discussing it in Lineage Slack currently.

osm0sis:
The reason for this change was /vendor, and I guess mounted /vendor is only needed for running dalvikvm, so I'll close this, since likely the weird devices using other names for /vendor don't rely on it for dalvikvm.

PR #2758, opened 2020-05-01
scripts: don't abort if /vendor fails to mount

osm0sis:
- /vendor is used only on some older devices for recovery AVBv1 signing so is not critical if fails
- this fixes installation in Lineage 17 Recovery on some older devices where /vendor is actually by-name partitions like oem, cust (or even cache), which likely also don't require the AVBv1 signing

Related: #2757


nicknitewolf:
Works fine now, thanks

PR #2784, opened 2020-05-09
magiskhide: add vendor.* props

osm0sis:


Fox2Code:
If everyone add prefix `vendor.` maybe having only props without `vendor.` prefix and make magisk automacily check for both variant with/without `vendor.` prefix is an easier solution to maintain props hide up to date.

PR #2809, opened 2020-05-17
Hide certain props in boot_complete

### LINKED ISSUE #2803: Magisk canary 122b4d6 will break OP7T + OP8/8Pro fingerprint unlock
OnePlus 7T HD1900 OOS India 10.3.2.HD65AA

It only happened on https://github.com/topjohnwu/Magisk/commit/122b4d66b62db2ff6d5c6aa02933c33f121217d2, previous build https://github.com/topjohnwu/Magisk/commit/a62bdc58cb2e06128b82d332ba1f4e72f5700269 works fine.
I'm not able to get logcat right now, sorry for this.

sabpprook:
Resolves broken fingerprint since Canary 20411 on OnePlus 7T, 8, and 8 Pro.

Fixes #2803

### DIFF
--- a/native/jni/core/bootstages.cpp
+++ b/native/jni/core/bootstages.cpp
@@ -326,6 +326,8 @@ void boot_complete(int client) {
 
 	auto_start_magiskhide();
 
+	hide_sensitive_props_late();
+
 	if (access(MANAGERAPK, F_OK) == 0) {
 		// Install Magisk Manager if exists
 		rename(MANAGERAPK, "/data/magisk.apk");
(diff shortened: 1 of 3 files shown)


topjohnwu:
Thanks for the PR!
The fix isn't as trivial as you and @osm0sis might have think 😜, namely MagiskHide state race conditions, post-fs-data state race conditions, and this also has to be enabled when MagiskHide is not auto started but the user later enabled MagiskHide.

Check my implementation for more details: 97db49a57bab9fc573eee861b59bed95dfd5812a

PR #2821, opened 2020-05-23
Allow consecutive points

vvb2060:
AOSP allows it, why do we not allow it?


vvb2060:
Example: WeChat. `com.tencent.mm:hotpot..`

PR #2824, opened 2020-05-23
Add DoH

vvb2060:


app/src/main/java/com/topjohnwu/magisk/di/NetworkingModule.kt
```diff
@@ -43,6 +45,10 @@ fun createOkHttpClient(context: Context): OkHttpClient {
         builder.sslSocketFactory(NoSSLv3SocketFactory())
     }
 
+    builder.dns(DnsOverHttps.Builder().client(builder.build())
+            .url(HttpUrl.get("https://162.159.36.1/dns-query")) //Cloudflare
```
topjohnwu, line 49: Why are you using raw IP addresses instead of https://cloudflare-dns.com/dns-query?

vvb2060: In order to remove the initial DNS query, this domain name is also polluted.

vvb2060:
Pre-resolved IP added from certificate.

PR #2829, opened 2020-05-24
Fix RU strings

Displax:


Displax:
@topjohnwu why is this closed? "Режим" much better than "Статус".

topjohnwu:
@Displax I accidentally pressed the wrong button :p

PR #2830, opened 2020-05-25
magiskboot: add support for lz4 compressed dt (extra)

osm0sis:
- legacy devices brought up to Android 10 may now use a compressed dt in a hdr_v0 AOSP dt variant extra section, so detect, decompress and recompress this
- so far these have only been done using lz4 compression (latest format revision magic), e.g. LOS 17.1 victara (Moto X)


osm0sis:
Reference file for testing purposes: https://gitlab.com/osm0sis/boot-img-collection/-/blob/master/aosp-osver-lz4qcdt1-lzmard-moto_x-victara-los17-recovery.img :+1:

npjohnson:
It works on victara!

npjohnson:
As always, thanks John!

PR #2840, opened 2020-05-29
Fix Direct Install on NAND devices

### LINKED ISSUE #2740: Direct Installation method of Magisk (20408) using MM (284) Fails on NAND/MTD based devices
This is branch of #2720 , originating from #1526 .

Finding bootimage should be after closing issues above fixed, and while flashing from recovery it is actually working well.
When trying to update Magisk using "direct Installation" option in Magisk manager (which is recommended method), the flashing process supposedly finds bootimage, but fails after dumping.

Output of Magisk Manager:
```
- Target image: /dev/BOOT
- Device platform: armeabi-v7a
! Unsupported/Unknown image format
! Installation failed
```
I think I will need to examine what actually is contained in /dev/BOOT and whether it dumps correct image.
I'll try to update on the issue when possible.

osm0sis:
- install functions are reimplemented in Java so the nanddump workarounds in util_functions.sh (from 47e50e8511b2b8990d72d8dde02e69d22e2e0c76) must be mirrored here as well

Fixes #2740

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstallImpl.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstallImpl.kt
@@ -41,6 +41,7 @@ abstract class MagiskInstallImpl : FlashResultListener {
 
     private val console: MutableList<String>
     private val logs: MutableList<String>
+    private var srcNand: String = ""
     private var tarOut: TarOutputStream? = null
 
     private val service: GithubRawServices by inject()
@@ -239,6 +240,11 @@ abstract class MagiskInstallImpl : FlashResultListener {
     }
 
     private fun patchBoot(): Boolean {
+        if ("[ -c $srcBoot ] && nanddump -f boot.img $srcBoot".sh().isSuccess) {
+            srcNand = srcBoot
+            srcBoot = File(installDir, "boot.img").path
+        }
+
         var isSigned = false
         try {
             SuFileInputStream(srcBoot).use {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstallImpl.kt
```diff
@@ -258,6 +264,10 @@ abstract class MagiskInstallImpl : FlashResultListener {
             return false
         }
 
+        if (!srcNand.isEmpty()) {
```
diareuse, line 267: This line will cause a crash in case the first if is not a success. You cannot access lateinit property when it's uninitialised. I'd suggest changing
`private lateinit var srcNand: String`
to
`private var srcNand: String = ""`

osm0sis: Thanks! Fixed up per your suggestion in my latest force push. :raised_hands:

osm0sis:
Ah, yup, SuFileInputStream must need the full path, so "boot.img" for our local file isn't enough.

Resolved in my latest force push by setting srcBoot with the full path instead. :+1:

488315:
@osm0sis is there any updates for supporting amonet patched boot image through direct installation yet?

osm0sis:
@488315 you're off-topic for this PR. amonet is more complex than simple magic-matching, so no, it likely won't be directly supported here any time soon. I *have* added support to it to my Android Image Kitchen in my recent v3.7 release, howerver. If you'd like to open your own issue for amonet support to keep it on the radar, feel free, but please don't discuss it further here.

PR #2845, opened 2020-05-30
Fix: log spamming on NAND devices

Msprg:
Appended ` >&2` at the end of the `flash_eraseall` and `nandwrite` to stop spamming progress lines on the screen while direct install in Magisk Manager.


osm0sis:
Handled in #2840 since it was my solution. :+1:

Issue #2027, opened 2019-11-03
Android TV Support


YouROK:
I cannot select modules from the repository using the remote control. Also this problem is in the exposed

YouROK:
Add to item layout:
 android:focusable="true"
 android:clickable="true"

topjohnwu:
@diareuse I have no plans to work on this, but you might want to check if this is feasible in your redesign

topjohnwu:
Use this issue to track Android TV support

topjohnwu:
#2175

PR #2861, opened 2020-06-03
Disable Volume keys in flash activity

Fox2Code:
Recap of old pull #2457: Sometimes modules ask to push volumes keys but the problem is that by pushing the volume keys during a flash it also changes the volume of the phone.

This time the code is much cleaner and should help more the project

I add a new API `BaseUIFragment.onKeyEvent` that should help #2027  
(Currently used for blocking volumes keys)

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/ui/base/BaseUIActivity.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/ui/base/BaseUIActivity.kt
@@ -2,6 +2,7 @@ package com.topjohnwu.magisk.ui.base
 
 import android.content.Intent
 import android.os.Bundle
+import android.view.KeyEvent
 import android.view.View
 import android.view.ViewGroup
 import androidx.appcompat.app.AppCompatDelegate
@@ -94,6 +95,10 @@ abstract class BaseUIActivity<ViewModel : BaseViewModel, Binding : ViewDataBindi
     override fun onResume() {
         super.onResume()
         delegate.onResume()
+    }
+
+    override fun dispatchKeyEvent(event: KeyEvent): Boolean {
+        return currentFragment?.onKeyEvent(event) == true || super.dispatchKeyEvent(event)
     }
 
     override fun onEventDispatched(event: ViewEvent) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/topjohnwu/magisk/ui/base/BaseUIActivity.kt
```diff
@@ -96,6 +97,10 @@ abstract class BaseUIActivity<ViewModel : BaseViewModel, Binding : ViewDataBindi
         delegate.onResume()
     }
 
+    override fun dispatchKeyEvent(event: KeyEvent): Boolean {
+        return currentFragment?.onKeyEvent(event) as Boolean || super.dispatchKeyEvent(event)
```
diareuse, line 101: This will crash the application when `currentFragment == null`. I suggest changing the the forced cast to optional like so:
```kotlin
return currentFragment?.onKeyEvent(event) == true || super.dispatchKeyEvent(event)
```

app/src/main/java/com/topjohnwu/magisk/ui/flash/FlashFragment.kt
```diff
@@ -62,6 +63,10 @@ class FlashFragment : BaseUIFragment<FlashViewModel, FragmentFlashMd2Binding>()
         super.onDestroyView()
     }
 
+    override fun onKeyEvent(event: KeyEvent): Boolean {
+        return event.keyCode == KeyEvent.KEYCODE_VOLUME_UP || event.keyCode == KeyEvent.KEYCODE_VOLUME_DOWN
```
diareuse, line 67: This doesn't look really nice and additional keys will result in ugly commits.

This would be way nicer:
```kotlin
return when(event.keyCode) {
    KeyEvent.KEYCODE_VOLUME_UP,
    KeyEvent.KEYCODE_VOLUME_DOWN -> true
    else -> false
}
```

Fox2Code:
Thank @diareuse for reviewing my code
I am new to Kotlin even if I am a long time java developers so thank
I will apply theses changes and squash commit immediately

Fox2Code:
Done!

Fox2Code:
@topjohnwu Why this pull request is not merged yet?
I am curious about what you are waiting to merge this pull request since t has not been touch.
If you haven't merged this pull request sooner I think there a reason behind it.

Fox2Code:
Just to know if I need to change anything in this pull.

PR #2899, opened 2020-06-20
Fix bytecode compatibility when compiling with Java9 ( Fix #2898 )

### LINKED ISSUE #2898: Manager crashing when performing "Direct Install"
Latest canary 0850bca9 
Log is here.

[log.txt](https://github.com/topjohnwu/Magisk/files/4807884/log.txt)

Fox2Code:
This commit is a bit technical and may need bytecode knownedge to fully understand but I will try to explains it anyways
Fix #2898

This bugs is due to the fact that the bytecode generated in java9 make `buf.position(off)` with a java8 runtime  
This is due to a addition in the `ByteBuffer` / `CharBuffer` / `***Buffer` in java9 that add  
`public ByteBuffer position(int i)` this overide the supermethod in  `Buffer`
`public Buffer position(int i)` this lead to generate  
`INVOKEVIRTUAL java/lang/ByteBuffer position (I)Ljava/lang/ByteBuffer;`  instead of  
`INVOKEVIRTUAL java/lang/ByteBuffer position (I)Ljava/lang/Buffer;`  and because the descriptor  
`position (I)Ljava/lang/ByteBuffer;` for method `position` doesn't exists in Java8 this cause to create a `NoSuchMethodError`
Even if we don't need the return value java still throw this error because the return value asked doesn't exist  

I english the literal traduction is that when compiled in java9 it's ask `ByteBuffer` but since in java8 only the method that return `Buffer` exists it just return that the method doesn't exist  

To workaround this issue we just cast to `Buffer` forcing java9 to always ask for `Buffer` as return value instead of `ByteBuffer` for its method calls opcodes

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/utils/PatchAPK.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/utils/PatchAPK.kt
@@ -20,6 +20,7 @@ import io.reactivex.Single
 import timber.log.Timber
 import java.io.File
 import java.io.FileOutputStream
+import java.nio.Buffer
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
 import java.security.SecureRandom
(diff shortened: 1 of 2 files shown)


topjohnwu:
Thank you for bringing awareness to this issue! I used a more "proper" fix for JDK building incompatible bytecode

Fox2Code:
I hope this works.
If this really work I might use this method instead of for other projects that need it.

PR #2914, opened 2020-06-26
Let isolation namespace base on app namespace

vvb2060:
Let isolation namespace base on app namespace instead of the global namespace


topjohnwu:
This would be better to introduce a new option in settings as it is different than it used to be

vvb2060:
Isolation namespace should be the most restrictive option, it isn't the default.  In most cases it's no different than before.  
If the app in need will use `su -mm`, so I don't think it's necessary to add another option, too many options will cause confusion.

PR #2916, opened 2020-06-26
Remove force_pm_install

vvb2060:
gms package verifier only checks the `notBefore` value of the certificate.


osm0sis:
Won't this trigger the Play Protect issue again since you removed the workaround in force_pm_install?

vvb2060:
Tested on gms version 20.24.13 (150400-316577029)

vvb2060:
gms will warn apps signed with the recently generated certificate, which is why patched manager could not be installed before.

PR #2919, opened 2020-06-26
Cast ByteBuffer to Buffer before calling position()

shakalaca:
On some devices (e.g ROG Phone) there will be error while patching boot.img:

Caused by: java.lang.NoSuchMethodError: No virtual method position(I)Ljava/nio/ByteBuffer; in class Ljava/nio/ByteBuffer; or its super classes (declaration of 'java.nio.ByteBuffer' appears in /system/framework/core-oj.jar)

### DIFF
--- a/signing/src/main/java/com/topjohnwu/signing/SignBoot.java
+++ b/signing/src/main/java/com/topjohnwu/signing/SignBoot.java
@@ -18,6 +18,7 @@ import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.security.PrivateKey;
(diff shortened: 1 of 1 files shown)


Displax:
Already fixed by https://github.com/topjohnwu/Magisk/commit/099e703834e8ace6eb4cc1b3e447677c5219a278

Fox2Code:
#2899 already made a similar pull request and explained the bug
It's because Java treats methods with same parameters, same name but different return value as completely different methods.
and the third cast to `Buffer` "**((Buffer)** image **)**.position() != headerSize" is uncessesary since the return value still `int` before and after cast
And also found the same problem with a `CharBuffer` in `app/src/main/java/com/topjohnwu/magisk/core/utils/PatchAPK.kt`
But topjohnwu fix is better because there no risk to forget a cast somewhere else in the code

shakalaca:
Ah didn't notice the gradle hack, but that one won't work, I'll send another patch.

Issue #3014, opened 2020-07-21
SAR and A/B interactions break internal install on Nintendo Switch


classic-gentleman:
This is quite peculiar but I believe is a legitimate case nevertheless.

As you may be aware, some hardware revisions of the Nintendo Switch are amenable to modding and Android ports have been created for it. Usually the port is installed to the SD card and it's straightforward. My particular setup, though, has Switchroot Pie installed to the eMMC, side by side with HorizonOS (and Ubuntu Linux, for the matter).

The HOS partitioning scheme is regular GPT, but it does reserve some partition names, of note:

```
PRODINFO -> /dev/block/mmcblk0p1
PRODINFOF -> /dev/block/mmcblk0p2
BCPKG2-1-Normal-Main -> /dev/block/mmcblk0p3
BCPKG2-2-Normal-Sub -> /dev/block/mmcblk0p4
BCPKG2-3-SafeMode-Main -> /dev/block/mmcblk0p5
BCPKG2-4-SafeMode-Sub -> /dev/block/mmcblk0p6
BCPKG2-5-Repair-Main -> /dev/block/mmcblk0p7
BCPKG2-6-Repair-Sub -> /dev/block/mmcblk0p8
SAFE -> /dev/block/mmcblk0p9
SYSTEM -> /dev/block/mmcblk0p10
USER -> /dev/block/mmcblk0p11
```

Herein lies the problem. `mmcblk0p10`, that is, `SYSTEM`, is being confused for an A/B system partition because of the case insensitive compare here https://github.com/topjohnwu/Magisk/blob/f1fb7404c20dbdcb32517e2f870f7b5a5d8a7386/native/jni/init/mount.cpp#L67 and the inexistent slot here https://github.com/topjohnwu/Magisk/blob/f1fb7404c20dbdcb32517e2f870f7b5a5d8a7386/native/jni/init/mount.cpp#L246

A solution would be to _invert_ the tested scenario  in `SARBase::mount_system_root` and look for the NVIDIA naming scheme _first_, then with `system[+slot]`, and `abort` in case that fails.

Second possibility would be to attempt case sensitive first, then failing all that resort to case-insensitive. It would be more complicated though.

A third possibility is to completely split the code paths between A/B and non-AB SAR.

My vote goes to solution 1, for least impact.

classic-gentleman:
Proposed patch:
[...]

PR #3015, opened 2020-07-21
Test for NVIDIA/Tegra partition naming scheme first

classic-gentleman:
Fixes https://github.com/topjohnwu/Magisk/issues/3014


osm0sis:
Do the scripts also need to be updated for this, or are we good?

classic-gentleman:
We're good. 👍

Issue #3008, opened 2020-07-20
Runtime apex overriding not working on Pixel 4 (uses a .apex zip)


Paschalis:
# Problem: I want to replace some bins/libs in the runtime apex
On a Google Pixel (_sailfish_/1st gen) I can successfully override the following runtime apex files with this module:

**mymodule:**
```
├── customize.sh
├── module.prop
└── system
    └── apex
        └── com.android.runtime.release
            ├── bin
            │   └── dex2oat
            └── lib64
                ├── libart-compiler.so
                └── libart.so
```

customize.sh:
```
# simply fixes permissions..
set_perm_recursive  $MODPATH                0       0       0755        0644
set_perm  $MODPATH/system/apex/com.android.runtime.release/bin/dex2oat 0       shell       0755
```

Installing the same module on a Google Pixel 4 (_flame_) the binaries/libs are not replacing the original ones.
The difference (and **probably the issue**) is that _Sailfish_ uses a folder for runtime apex:
`/system/apex/com.android.runtime.release/`
whereas _flame_ uses a `.apex` zip file:
`/system/apex/com.android.runtime.release.apex`

(also explained in another [issue comment](https://github.com/topjohnwu/Magisk/issues/2109#issuecomment-567303279))

Looking a bit on the code I can see [here](https://github.com/topjohnwu/Magisk/blob/master/scripts/util_functions.sh#L309) that there is support when the apex is a file, and it is supposed to unzip it, mount it, and then I guess replace with whatever I've put on my module in this path:
mymodule: `system/apex/com.android.runtime.release/`

Also on issue #2114  there was some preliminary support, but either part of the code was restructured/rewritten or wasn't included (I don't know the internals of `magisk`, so not very certain on this).
For example [here](https://github.com/topjohnwu/Magisk/blob/2c885344ddbe3c2f4e9cf5957869240e85234b44/scripts/util_functions.sh#L120) there's additional work on `recovery_actions`.

So is this a known issue?
Has anyone successfully replaced runtime apex bins/libs in cases where a zip file (.apex) file is used?
 
Thanks in advance!

# Additional info
Phone: Google Pixel 4/flame (The same module on Google Pixel/sailfish worked just fine)

Magisk Manager: 7.5.1
Magisk: 20.4.1

topjohnwu:
Github issues are for reporting issues of Magisk itself, not for support on how to use Magisk, including developing modules.
The issue here might be the system verifying the signature of APEX zips

[...]

PR #2114, opened 2019-11-29
Android Q Apex Fix

Zackptg5:
Fix for android Q devices in twrp based recoveries. [See issue for more details](https://github.com/topjohnwu/Magisk/issues/2109)

Mounts /system/apex modules to /apex location, renaming them as required. Doesn't touch system and enables dynamically linked binaries to work as expected in recovery

Issue #2109, opened 2019-11-26
Fix for Android Q Roms in TWRP - Apex


Zackptg5:
Due to the new apex stuff in Android 10, some key binaries and libs have been moved out of /system/bin, lib, etc and into /system/system/apex and then symlinked to /system/bin. Unfortunately, these include linkers:
`lrwxr-xr-x 1 root   shell         36 2008-12-31 19:00 linker -> /apex/com.android.runtime/bin/linker
lrwxr-xr-x 1 root   shell         38 2008-12-31 19:00 linker64 -> /apex/com.android.runtime/bin/linker64`
(and a ton more). 
The problem is that /system/apex is mounted to /apex at boot but TWRP hasn't been updated for Q yet (and it'll likely be a while before that happens) so this doesn't happen in TWRP. As a result, any dynamic binaries or libs such as toybox don't work in TWRP unless apex is mounted similarly. 

This can be fixed by bind mounting /system/apex as part of recovery_actions. However, even more complicated is that one of the apex modules (com.android.runtime.release) is mounted with a different name (com.android.runtime) so it's not as simple as simply bind mounting the apex folder. 

[See pull request for fix](https://github.com/topjohnwu/Magisk/pull/2114)

Zackptg5:
Opened new pull request, kinda botched the last one (not used to doing this on my mac)

Zackptg5:
Turns out the solution is more complex. Some Q roms have already extracted apex modules like the first one I tested it on but most don't. So the apex_payload.img needs to be extracted and then mounted to a loop device in these cases. I reworked the fix in the pull request #2114

Zackptg5:
Closed due to discussion in different group

PR #3017, opened 2020-07-23
Allowing modifications at /apex/

Paschalis:
Magisk allows the replacement of files at `/apex` only for devices that are using extracted APEX modules, but wasn't working for APEX zip files. The tiny changes in this PR allows overlaying files to `/apex` after a `.apex` file is extracted and mounted.
Not sure if that's the way to do it, but it serves my purpose so I'm just sharing it.
It follows a detailed explanation of the issues I've had at #3008.

#### Working example with extracted APEX folders `@v20.4`:
The `system.img` of Google Pixel (sailfish) has extracted APEX modules.
To replace e.g. the runtime and dex2oat one could use a module like:
```
├── customize.sh
└── system
    └── apex
        └── com.android.runtime.release
            ├── bin
            │   └── dex2oat
            └── lib64
                └── libart.so
```

During boot, android mounts this folder to: `/apex/com.android.runtime/`
Then magisk applies the changes at `/system/apex/com.android.runtime.release/`,
which are reflected to the `/apex` mount.  With `customize.sh` relevant permissions/ownerships can be set.

#### Issue with .apex  zip files
Google Pixel 4 (flame) and possible other devices use APEX zip files, like:
`/system/apex/com.android.runtime.release.apex`.

Those are extracted and mounted (`apex_payload.img`) to `/apex/com.android.runtime/`.
I've tried many things unsuccessfully, some of which include:
- replacing files with a module that puts relevant files at:
    + `/apex/com.android.runtime`
    + `/system/apex/com.android.runtime`               (this PR will make this to work)
    + `/system/apex/com.android.runtime.release` (this was working fine with `sailfish`)
- combinations of above with attempts to remove the `.apex` file by masking it with either:
    + `boot.img:ramdisk/overlay.d/` approach
    +  an empty file at `/system/apex/com.android.runtime.release.apex`
    + the `REPLACE` variable in customize.sh
- other attempts to extract the `.apex` at `/system/apex/`, so magisk could then apply the changes on that folder
- modifying `env` variables like `ANDROID_RUNTIME_ROOT`, and put my changes directly in `/system/lib` to trick `linker` to use my versions

### Handling `/apex` like other special read only partitions (e.g. `/vendor`):
The minor changes in `92af79c` in combination with the module below do the trick:
```
├── customize.sh
├── service.sh
└── system
    └── apex
        └── com.android.runtime
            ├── bin
            │   └── dex2oat
            └── lib64
                └── libart.so
```
The  folder: `/system/apex/com.android.runtime` must be used,
and not the renamed one (that has appended `.release` or `.debug`).
Just for the libraries (binaries are ok), one must set permissions/ownership using `service.sh` (instead of `customize.sh`) like:
```
chown system.system    /apex/com.android.runtime/lib/libart.so
chmod 0644                    /apex/com.android.runtime/lib/libart.so
```

So Android extracts the `/system/apex/com.android.runtime.apex` zip file,
then mounts it at `/apex/com.android.runtime`, and then magisk directly overlays the changes of the module.

### Minor bugfix `086fa726b`
This must have been introduced during a cleanup. It did not affect/fix my issue.
Related to  #2114 #2109.

### DIFF
--- a/native/jni/core/bootstages.cpp
+++ b/native/jni/core/bootstages.cpp
@@ -95,6 +95,7 @@ static bool magisk_env() {
 		if (0) {}
 		mount_mirror(system, MS_RDONLY)
 		mount_mirror(vendor, MS_RDONLY)
+		mount_mirror(apex, MS_RDONLY)
 		mount_mirror(product, MS_RDONLY)
 		mount_mirror(system_ext, MS_RDONLY)
 		mount_mirror(data, 0)
@@ -110,6 +111,7 @@ static bool magisk_env() {
 		LOGI("link: %s\n", buf1);
 	}
 	link_mirror(vendor);
+	link_mirror(apex);
 	link_mirror(product);
 	link_mirror(system_ext);
 
--- a/native/jni/core/module.cpp
+++ b/native/jni/core/module.cpp
@@ -595,7 +595,7 @@ void magic_mount() {
 		return;
 
 	// Handle special read-only partitions
-	for (const char *part : { "/vendor", "/product", "/system_ext" }) {
+	for (const char *part : { "/vendor", "/product", "/system_ext", "/apex" }) {
 		struct stat st;
 		if (lstat(part, &st) == 0 && S_ISDIR(st.st_mode)) {
 			if (auto old = system->extract(part + 1); old) {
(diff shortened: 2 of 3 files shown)


Paschalis:
Turns out it breaks stuff on `sailfish`, so I'll just use it as a temporary workaround for modifying art/compiler in `flame`.

Anonymous941:
Will this be reconsidered?  I don't know of any way to modify apex without this

PR #3037, opened 2020-08-01
Copy required files to system for addon.d

jesec:
/data/adb/magisk is not accessible on devices with encrypted data.

### DIFF
--- a/scripts/addon.d.sh
+++ b/scripts/addon.d.sh
@@ -7,37 +7,8 @@
 #
 ########################################################
 
-trampoline() {
-  mount /data 2>/dev/null
-  if [ -f $MAGISKBIN/addon.d.sh ]; then
-    exec sh $MAGISKBIN/addon.d.sh "$@"
-    exit $?
-  elif [ "$1" = post-restore ]; then
-    ps | grep zygote | grep -v grep >/dev/null && BOOTMODE=true || BOOTMODE=false
-    $BOOTMODE || ps -A 2>/dev/null | grep zygote | grep -v grep >/dev/null && BOOTMODE=true
-
-    if ! $BOOTMODE; then
-      # update-binary|updater <RECOVERY_API_VERSION> <OUTFD> <ZIPFILE>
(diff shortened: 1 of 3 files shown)


topjohnwu:
Sorry, but there is no plan to support running addon.d without data access as upgrading through Magisk Manager will not also update addon.d

jesec:
It can be made to update addon.d. I can implement the changes to make it happen if you want. This is rather an improtant issue for our LineageOS (and our downstream projects) users as we are not going to implement data encryption in Lineage recovery. 

An alternative solution is to move Magisk files from /data/adb to /data/unencrypted which is unencrypted in FBE systems and fulfills all the criteria documented in https://github.com/topjohnwu/Magisk/blob/master/docs/details.md#paths-in-data. It is even better. Here is a proof of concept and working implementation: https://github.com/jesec/Magisk/tree/dataunencrypted.

mu198:
oh that would be such a wonderful relief, but it's a shame that @topjohnwu closed it right away

synt4x93:
I don't understand why he is refusing to do this?

shrihanDev:
Bump. Does this still remain an issue, by the way?

mcgi5sr2:
Bump, imho this should not be closed

mu198:
it is still an issue, but what is completely ignored here, use instead of [Lygisk](https://github.com/programminghoch10/Lygisk), is a fork of @programminghoch10  and @Linux4, this fork bypasses the problem

shrihanDev:
No. Nope. Nada. Zero.

dtristany:
So just to clarify, this is only an issue if you use LOS official recovery? People on TWRP don't need Lygisk and don't have to do anything after OTA? That is, provided your install method was flashing on TWRP and any future updates to magisk also need to be flashed in TWRP recovery and not in the app, to make sure addon.d script is installed and kept updated? (I'm confused here re. dirty flashing vs sideloading in TWRP, first one doesn't require computer while obviously adb sideload does - which one is recommended for Magysk)

MaxMorfiX:
> @srccrow No. Nope. Nada. Zero.

But why?

PR #3055, opened 2020-08-07
Update to APK Signature Scheme v2

vvb2060:


topjohnwu:
The original code was specifically optimized for extreme low memory usage (on Android it can very easily trigger `OutOfMemoryError`). The code is also used to sign the release zip files and have to workaround some dumb recovery unzip implementations. This change will need further review/modification before it can be merged.

osm0sis:
Does a signed zip still work with recovery unzip?

jcgasca25:
> _No description provided._

https://github.com/topjohnwu/Magisk/pull/3055#issue-464656110

jcgasca25:
https://github.com/topjohnwu/Magisk/pull/3055#issue-464656110

topjohnwu:
@osm0sis the busybox in Magisk can unzip any valid zip file regardless of signed or not (no adjustment needed).
The v2 scheme seems to be doing similar stuff zipadjust is doing, so I would assume yes

osm0sis:
Nice. Thanks. Figured it worked since nobody has complained. 😄

PR #3113, opened 2020-08-28
Use proper AXML string patch for repackage

Fox2Code:
Actually Magisk modify `AndroidManifest.xml` by doing a hex patch of the file  
(Changing defined bytes of the file without changing the size of these bytes)  
This can cause the Stub to be detected if the file name is not 14 exactly charters long by [checking null bytes in the name](https://github.com/Fox2Code/MagiskStubFinder#null-byte-name-scanner)  
This also prevents setting a name larger than 14 charters.

I used some knowledge I found to create an Axml patcher to modify Strings in `AndroidManifest.xml` 
 that support changing String size *(Increasing and decreasing)* and automatically align the file correctly.

This allow:
- Changing the package length of the manager (This patch contain random package name generation)  
- Applying package name larger than 14 charters (This patch increase the name length limit to 32 but the theoretical limit is 255)  
- Fixing  null byte name checking 

Also in the patch call `findAndPatchAXML(xml, APP_ID, pkg, 9)` instead of  `findAndPatchAXML(xml, APP_ID, pkg)` to support patching larger strings because the new system is length sensitive.  

For example `findAndPatchAXML(xml, APP_ID, pkg, 9)` mean patch `"com.topjohnwu.magisk"` to `pkg` but also change 
every string  that start with `com.topjohnwu.magisk` with 9 aditional charters to `pkg` with the 9 aditionals charters unmodified  
at the end.  
So in this example `"com.topjohnwu.magisk.provider"` become `pkg + ".provider"` since the string `".provider"`  is 9 charter long.

I already tested the patch for:
- Checking that repackaging and name larger than 14 charters are working.
- Checking that SuReqests work correctly repackaged.
- Checking that MagiskStubFinder null byte name check doesn't trigger.

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/utils/AxmlUtil.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/utils/AxmlUtil.kt
@@ -1,0 +1,182 @@
+package com.topjohnwu.magisk.core.utils
+
+import java.io.ByteArrayOutputStream
+import java.nio.charset.Charset
+import java.util.*
+
+private val UTF_16LE = Charset.forName("UTF-16LE")
+private const val AXML_STRING_INDEXES_OFFSET = 0x20
+private const val AXML_CHUNK_SIZE_OFFSET = 0x04
+
+
+/** Hold a [ByteArray] reference that can be replaced */
+class ByteArrayPtr(var ptr: ByteArray)
+
+/** transform a byte to it's unsigned [Int] value */
+private fun Byte.getValue(): Int = when {
+    (this.toInt() < 0) -> this.toInt() + 256 else -> this.toInt()
+}
+
+/** this represent an index translation map */
+class IndexMap {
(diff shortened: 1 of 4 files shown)


Fox2Code:
Ready for review and merge. 
I spend a lot of time making and testing the patch to be sure it's working.

topjohnwu:
Thank you for your contribution! After seeing this PR, I've done some research (reading the AOSP sources for parsing binary XML) and rewritten everything from scratch (fbaf2bded6a4e94c6a0d927139b710e6fcec913e)

Fox2Code:
Ok, thanks @topjohnwu for the patch and I'm happy my code helped you.  
I also have other ideas for using this like to change every activity names for the stub to avoid detection on each hide  
I don't know exactly how to submit stub obfuscation ideas, it is ok if I create an issue for obfuscation ideas or I should put these ideas somewhere else?

topjohnwu:
The component names will always be different for each stub build

PR #3114, opened 2020-08-29
Optimize network status display

vvb2060:
Catch retrofit2 RuntimeException.
Hide SafetyNet when no network.
Update "loading" to "not available" when fetch update failed.
Fix NetworkObserver on Android 11.  `NetworkInfo` was deprecated in API level 29, `getNetworkInfo` can't get accurate information when network lost. This is just a temporary fix.

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/ktx/XAndroid.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/ktx/XAndroid.kt
@@ -5,10 +5,7 @@ import android.content.ComponentName
 import android.content.Context
 import android.content.ContextWrapper
 import android.content.Intent
-import android.content.pm.ApplicationInfo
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/topjohnwu/magisk/ktx/XAndroid.kt
```diff
@@ -63,7 +60,11 @@ val PackageInfo.processes
             receivers?.processNames.orEmpty() +
             providers?.processNames.orEmpty()
 
-val Array<out ComponentInfo>.processNames get() = mapNotNull { it.processName }
+val Array<out ComponentInfo>.processNames
```
topjohnwu, line 63: I'll keep this out for now as toggle isolated processes does not do anything yet.
Would be better to add this after some functionality is introduced.

PR #3175, opened 2020-09-13
Fix text readability of failed SafetyNet result

RikkaW:


topjohnwu:
Fixed LOL 60e1e07e873d7b72aa3b6f80a0c8f65dad588699

RikkaW:
Save Context-related values into a ViewModel is definitely not a good idea since context (resources/theme) can change during the life of the ViewModel.

By the way, if I doing this from scratch, I prefer color state list 😶, so simple and clean 🏃‍♂️.

PR #3176, opened 2020-09-13
Better dark theme colors

RikkaW:
The original colors make cards too hard to recognize.

### DIFF
--- a/app/src/main/res/values-night/themes_md2.xml
+++ b/app/src/main/res/values-night/themes_md2.xml
@@ -9,8 +9,8 @@
         <item name="colorSecondary">#3E78AF</item>
         <item name="colorSecondaryVariant">#803E78AF</item>
         <item name="colorSurface">#0D0D0D</item>
-        <item name="colorSurfaceVariant">#171717</item>
-        <item name="colorSurfaceSurfaceVariant">#292929</item>
+        <item name="colorSurfaceVariant">#272727</item>
+        <item name="colorSurfaceSurfaceVariant">#323232</item>
         <item name="colorOnPrimary">#F9F9F9</item>
         <item name="colorOnPrimaryVariant">#D9E6E6E6</item>
         <item name="colorOnSecondary">#F9F9F9</item>


diareuse:
I can't see a reason why would you need to recognize the borders of the cards. The whole point is to have a _very subtle_ container.

If you feel like the contrast is insufficient between `colorSurfaceVariant` and `colorSurfaceSurfaceVariant` then I'd opt into increasing brightness in `colorSurfaceSurfaceVariant` only. So what's the real intention behind this @RikkaW 🙂 ?

---

Speaking of which, the light mode's `colorSurface` and `colorSurfaceVariant` could be changed to `#FFF` and `#F7F7F7` respectively so the aesthetic matches the dark theme.

RikkaW:
I messed up 🙈, now GitHub does not recognize new changes and I can't reopen this pull request. Maybe I have to open a new pull request.

My original thought is:
On lower screen brightness, `colorSurfaceVariant` looks too close the the `colorSurface` (especially when disabled) and light theme's `colorSurface` and `colorSurfaceVariant` has higher contrast, so I change `colorSurface` to `#272727`. It does look better (especially disabled option is not that close to the background) in my option. Then, `#272727` is too close to `colorSurfaceSurfaceVariant ` `#292929`, so I make `colorSurfaceSurfaceVariant ` a little lighter.

diareuse:
I wouldn't concern too much with a lower brightness levels. We absolutely cannot cover all possible use cases and permutations of brightness / theme types. 

One could argue that light colors are way too light even when brightness is low and other would come out and say that now the full brightness version is way too dim. This can be an endless fight, I'd like to stay out of.

As dark mode is here to [save battery](https://www.businessinsider.com/google-confirms-dark-mode-on-android-phones-saves-battery-life-2018-11?op=1) there's absolutely no need to have the brightness very low, hence we will not consider the card legibility important in this particular use case.

I absolutely appreciate bringing this to our attention and your effort in improving MM though! The PR will be nevertheless closed for aforementioned reasons.

RikkaW:
@diareuse One more little question: what do you think about the color of AppBar, should it be lighter or darker than the cards? The current color is `#141414` on Amoled theme and `#202020` on others.

diareuse:
According to Material Design's behavior the elements with higher elevation should be brighter than the content underneath it. So if we were to hold true to the MD it would need to be brighter.

Nevertheless I don't think I'm sticking to the principles 100% and it doesn't make any difference to me. The important thing is that it has different color than the cards which should be currently in place if it wasn't changed by anybody else.

The colors are different for AMOLED and other themes so the brightness difference is perceived in a same way.

PR #3177, opened 2020-09-13
Black splash screen background for dark theme

RikkaW:


diareuse (approved):
@topjohnwu This is fine. 👍

palmbeach05:
Wouldn't it be better to alter this to be an option within Magisk Manager settings for the user to decide rather than default to what the system is set to? For example, what if you have a user who uses dark mode on their system, but prefers the green background splash screen. They are now forced to change their system to not be in dark mode.

The user is currently able to select the theme of the Magisk Manager as light, dark, or follows system theme. I would imagine it would not be too hard for the original author of this commit to alter his code and add/extend this existing concept to the splash screen.

RikkaW:
@palmbeach05 The "splash screen" is controlled by the system when the app first starts, it's impossible to have option to control it.

diareuse:
*Not impossible. Just very, very hard to do.

PR #3199, opened 2020-09-23
init/mount: support for dm-verity verified root

alaviss:
This commit adds support for kernel initialized dm-verity on legacy SAR
devices.

Tested on a Pixel 2 XL with a kernel patch to initialize mappings
specified via the `dm=` kernel parameter even when an initramfs is used.

I don't expect for this to be merged as it's a really niche use case
(dm-verity/AVB on a rooted device), but since it's relatively simple
I'm gonna give it a shot.

### DIFF
--- a/native/jni/init/mount.cpp
+++ b/native/jni/init/mount.cpp
@@ -12,11 +12,19 @@
 
 using namespace std;
 
+static string rtrim(string &&str) {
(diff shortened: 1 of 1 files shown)


topjohnwu:
@alaviss I'm more than happy to merge, but I'll merge after the next release

topjohnwu (approved):

PR #3217, opened 2020-09-30
Update Romanian

sn-o-w:


bgtha1:
I'm new so guides would be appreciated

topjohnwu (approved):

PR #3351, opened 2020-10-15
Some manager updates

vvb2060:


osm0sis:
While you're still adding commits here, can you redo fd552e68a9b3a5056dbe9abd4566dff78cdb3558 so it can be togglable in the list like system apps, not a hardcoded exclusion? There are some useful cases for being able to hide OEM apps.

osm0sis:
Thanks! 🙌

topjohnwu:
Manually merged most changes into `master`

Issue #2417, opened 2020-02-08
[Security Issue] App can fake click on SuRequest to accept them


Fox2Code:
After reading #2414 I dig into magisk manager code and seen the SuRequest don't have `FLAG_SECURE` flag
If this flag is not enabled it allows apps with accessibility service to click on the buttons of the **SuRequest** 
This is the lines of codes
```Java
this.getWindow().setFlags(WindowManager.LayoutParams.FLAG_SECURE,
     WindowManager.LayoutParams.FLAG_SECURE);
```
I took from here https://github.com/LineageOS/android_packages_apps_Jelly/blob/lineage-17.1/app/src/main/java/org/lineageos/jelly/MainActivity.java (line 289)
I am a bit busy today so I can't make the patch and a pull request today so I prefer making an issue to make this problem fixed ASAP

PR #3371, opened 2020-10-19
Tapjacking protection

vvb2060:
#2417 
> Secure surfaces are used to prevent content rendered into those surfaces by applications from appearing in screenshots or from being viewed on non-secure displays. An application can use the absence of this flag as a hint that it should not create secure surfaces or protected buffers on this display because the content may not be visible.

This flag cannot block the accessibility services click button. @Fox2Code

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
@@ -70,6 +70,10 @@ abstract class BaseUIActivity<VM : BaseViewModel, Binding : ViewDataBinding> :
         ensureInsets()
     }
 
+    fun setAccessibilityDelegate(delegate: View.AccessibilityDelegate?) {
+        viewRoot.rootView.accessibilityDelegate = delegate
+    }
+
     override fun onResume() {
         super.onResume()
         viewModel.requestRefresh()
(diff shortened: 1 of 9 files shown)


Fox2Code (approved):

Fox2Code:
The `window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)` was to prevent screen readers from watching the popup.  

You can add it back if you think it can be useful. @vvb2060

vvb2060:
Screen readers do not view the window as an image, they read the structure data of the window through the accessibility API.

vvb2060:
If it were me, I would guess the button position by the screen size.

Fox2Code:
Same, but I still think the flag could be useful because Android has other APIs to allow faking user input other than accessibility.

vvb2060:
Like what? As far as I know, only the accessibility feature can do that. Or use root

RikkaW:
@Fox2Code `FLAG_SECURE` only prevents screenshot, stop "I think".

By the way, I personally don't agree to kill accessibility features since they are designed for disabilities only. If some brainless user insists to use some unsafe accessibility apps, let him go.

vvb2060:
If you're still worried, you can add [Android Protected Confirmation](https://android-developers.googleblog.com/2018/10/android-protected-confirmation.html) to Magisk, which uses strong security from the hardware to avoid everything you're worried about.

Fox2Code:
I already saw AnyDesk use a permission called `INJECT_EVENT` (I think) to do user interaction without using an Accessibility Service.  

But actually I don't have the time to check the AOSP source code to see how it's works due to personal reasons.

RikkaW:
@Fox2Code Stop your imagine pls, if you want you can use Android Protected Confirmation, the whole process is on secure hardware.

vvb2060:
```
  permission:android.permission.INJECT_EVENTS
  package:android
  label:null
  description:null
  protectionLevel:signature
```
If you're worried that your phone's OS will attack you, I think the only solution is to use protected confirmation, secure hardware designed to work without trusting the OS.

topjohnwu (approved):

PR #3372, opened 2020-10-19
Update details.md

deril1:
magiskpolicy/rules.cpp


deril1 (commented):
magiskpolicy/rules.cpp

PR #3491, opened 2020-11-15
Deal with '\r\n' in sepolicy

### LINKED ISSUE #3490: [Bug] Cannot deal with '\r\n' in sepolicy.rule
0df9f6a5-fd75-4710-9a9a-e929aa443aca
As the above screenshot is shown, when the sepolicy file is in CRLF format, sepolicy will fail. Changing to LF solves this problem.
This happens when the module is compiled in windows (such as EdXposed). This might be relative to #3486 and #3466.

yujincheng08:
Fix #3490. Didn't compile and test. Hope it works :).


topjohnwu:
This is incorrect, but still thanks :D

Issue #3493, opened 2020-11-15
Modules fail to install in TWRP on Magisk 21.1


gossbill:
I have Magisk 21.1 (21100, beta) installed. I've found that trying to flash a module in TWRP gives an error 127 due to "xargs: /data/adb/magisk/busybox: No such file or directory". (See line 766 and following in recovery log).
However, if I first flash the Magisk zip (line 796ff) and then the module (line 935) it works.

After rebooting the phone (just rebooting recovery doesn't cause the issue) the installation of modules will fail again.

Notes:
- see also the discussion starting here https://forum.xda-developers.com/showthread.php?t=2239421&p=83926347
- I've also tried MagiskHidePropsConf-v5.3.6 and I get the same error
[21.1-module-recovery.zip](https://github.com/topjohnwu/Magisk/files/5543087/21.1-module-recovery.zip)

topjohnwu:
Fixed

PR #3494, opened 2020-11-15
Fix finding busybox binary

Displax:
Close https://github.com/topjohnwu/Magisk/issues/3493

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -106,8 +106,6 @@ ensure_bb() {
     bb=$TMPDIR/busybox
   elif [ -f $MAGISKBIN/busybox.bin ]; then
     bb=$MAGISKBIN/busybox.bin
-  elif [ -f $MAGISKBIN/busybox ]; then
-    bb=$MAGISKBIN/busybox
   else
     abort "! Cannot find BusyBox"
   fi
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -104,6 +104,8 @@ ensure_bb() {
   local bb
   if [ -f $TMPDIR/busybox ]; then
     bb=$TMPDIR/busybox
+  elif [ -f $MAGISKBIN/busybox.bin ]; then
+    bb=$MAGISKBIN/busybox.bin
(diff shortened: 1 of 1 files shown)


osm0sis:
Hmm I don't see why the -f would fail.. the new busybox script is still a file. 😕

osm0sis:
Also I think the wrapper should be used still, since this will basically undo the recent Samsung fix that the wrapper was created for.

tsynik:
> Hmm I don't see why the -f would fail.. the new busybox script is still a file. 😕

I think this related to this PR. With latest canary magisk installed, there is no /data/adb/magisk/busybox, so for example xposed-framework-magisk.zip module can't be installed from TWRP recovery, here is log:

```
I:operation_start: 'Flashing'
Installing zip file '/sdcard/xposed-framework-magisk.zip'
Checking for Digest file...
I:Update binary zip
I:Extracting updater binary 'META-INF/com/google/android/update-binary'
I:Zip does not contain SELinux file_contexts file in its root.
[amonet] Load microloader from recovery... 
[amonet] Remove boot patch...
[amonet] OK
I:Legacy property environment not used in updater.
xargs: /data/adb/magisk/busybox: No such file or directory
Updater process ended with ERROR: 127
```

tsynik (approved):
this works

osm0sis:
No, I think a different way will need to be found that keeps the fix for Samsung kerels in https://github.com/topjohnwu/Magisk/commit/f9c7a4c933f6116e41d65856555d9114bdbaa674 intact.

You might be right that it could be related to the shebang, we'll need to investigate further.

CC: @Displax

tsynik:
> No, I think a different way will need to be found that keeps the fix for Samsung kerels in [f9c7a4c](https://github.com/topjohnwu/Magisk/commit/f9c7a4c933f6116e41d65856555d9114bdbaa674) intact.
> 
> You might be right that it could be related to the shebang, we'll need to investigate further.
> 
> CC: @Displax

But this fix related only to util_functions.sh and recovery mode install, so it will keep Samsung fix intact.

edg84 (approved):

PR #3505, opened 2020-11-20
add Albania translators

cocka11:


topjohnwu:
You placed it in the wrong location

PR #3573, opened 2020-12-13
move SSLSocketWrapper to Kotlin

jumpstreet13:
I moved SSLSocketWrapper to Kotlin, also i can move all classes in module to kotlin


topjohnwu:
The class is used in the stub, and no Kotlin is used to reduce final APK size (Kotlin requires its large standard library)

PR #3622, opened 2020-12-30
Show a warning when the manager is not a user app or not in internal storage

### LINKED ISSUE #3606: Add warning when Magisk Manager is installed on external storage
### The problem

I am using my mobile phone always with the developer setting `Force allow apps on external` turned on.
This gives me the choice of what to put on the SD-Card and what not to.
It will also automatically install new apps to external storage.

When I installed Magisk Manager, it was also put onto the SD-Card.
Me knowing of this, I have moved it back, but before that I noticed something.

Magisk Manager will **not** detect a _running and working_ Magisk installation, when the Manager is installed on external storage.

Obviously that is fine as it is, as there is no point in having Magisk Manager installed on external storage, 
but for people who also use this setting and are not aware of the effect is has on Magisk Manager, 
this might be confusing.

### The solution

This is why I propose addition of a warning message/popup when Magisk Manager is installed on external storage.
It might also prevent unnecessary issues from users who are not aware that their Manager is installed incorrectly and not the framework.

RikkaW:
Fix #3606

### DIFF
--- a/app/shared/src/main/AndroidManifest.xml
+++ b/app/shared/src/main/AndroidManifest.xml
@@ -1,13 +1,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    package="com.topjohnwu.shared">
+    package="com.topjohnwu.shared"
+    android:installLocation="internalOnly">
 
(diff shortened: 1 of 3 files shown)


RikkaW:
Replaced by https://github.com/topjohnwu/Magisk/pull/3692

PR #3625, opened 2020-12-30
Support rootwait cmdline parameter on legacy SAR

bylaws:
On devices where the primary storage is slow to probe it makes sense to
wait forever for the system partition to mount, this emulates the
kernel's behaviour when waiting for rootfs on SAR if the rootwait
parameter is supplied.

This issue was encountered with some SD cards on the Nintendo Switch.


topjohnwu (approved):

akkim31:
#3570

akkim31 (commented):
#

akkim31 (commented):
#

akkim31 (commented):
#

Issue #3250, opened 2020-10-04
Issue with the design of the new Magisk Manager


xmxrtyx:
Hey topjohnwu,
Nice job on the new Magisk Manager, looks really nice.
I came across this issue some time back when I was flashing a mod.
![Screenshot_20201004-193636549](https://user-images.githubusercontent.com/56784444/95017809-30889900-0679-11eb-8434-1cdcad7f784d.jpg)
The 'Flashing' covers the text. Maybe you can reduce the opacity or move the 'Flashing' to a corner
Thank You

magicgoose:
noticed the same thing. I don't think it needs to be over the console, there's plenty of space in the toolbar at the top. please move it.

osm0sis:
@diareuse can address this perhaps. A couple more options could be some overscroll so one can pull the text up to read it, or, when you tap and hold on it it could disappear temporarily.

diareuse:
Good idea @osm0sis 👍 I'll try to whip up something like that 🙂

Mikesew1320:
Yeah, that's better. 😍

PR #3627, opened 2020-12-31
Remove "Flashing" overlay

### LINKED ISSUE #3579: Request: Do something different with the "Flashing" overlay
I think we need to do something different with that "Flashing" animation overlay thing that you get when installing a zip, patching the boot img, etc.

I ran into a scenario where I was installing a module that had a bit of an install process that was interactive (via volume keys), and it got to a part where it was warning about compatibility with some devices, however I couldn't read all of it because the flashing overlay covered it up.

Maybe a simple workaround would be if the user clicks it, then it becomes mostly transparent?

RikkaW:
Replace "Done" and "Failed" with SnackBar, maybe it can be replaced with other forms in the future.

Fix #3579


ammmze:
Also fixes #3250

ammmze (approved):

topjohnwu (approved):

Issue #3787, opened 2021-01-28
LG G5 (H830) bootloop (Magisk 21.1+)


Disappear9:
System:
```
Fulmics ROM 5.5
Android ver: 7.0
Kernel ver: 3.18.97-H830_v1.5.5-mk2000
sec patch: 2017/11/06
```

Tested magisk ver:
```
CANARY 6fb20b3e BOOTLOOP
20.4 OK
21.0 OK
21.1 BOOTLOOP
21.2 BOOTLOOP
21.3 BOOTLOOP
21.4 BOOTLOOP
```

Logs:

[console-ramoops.txt](https://github.com/topjohnwu/Magisk/files/5887477/console-ramoops.txt)
[console-ramoops-long.txt](https://github.com/topjohnwu/Magisk/files/5887481/console-ramoops-long.txt)
[dmesg.txt](https://github.com/topjohnwu/Magisk/files/5887482/dmesg.txt)
[recovery.txt](https://github.com/topjohnwu/Magisk/files/5887483/recovery.txt)

DenyDarko:
https://github.com/topjohnwu/Magisk/issues/3635

I've already provided everything you did but didn't bother reopening another issue as I don't have the knowledge to build the kernel with debug flags enabled.

Lishoo:
Hi, try the build from my last post on this topic https://github.com/topjohnwu/Magisk/issues/3770

Disappear9:
> Hi, try the build from my last post on this topic #3770

It works!

Issue #3770, opened 2021-01-25
Latest canary version 07bd36c9 (21505) damages boot.img / removes RAW data


Lishoo:
Latest version of Magisk still damages the boot.img when patching, output boot.img is 10mb instead of 22mb.
Differences in logs 
patched and working v21.4 (magiskboot) 
```
CHECKSUM        [b8aa7cfc33c9db071a53a39d5aae8fd4233e6d9a000000000000000000000000]
KERNEL_DTB      [5140711]
**KERNEL_FMT      [lz4_legacy]**
RAMDISK_FMT     [gzip]
**EXTRA_FMT       [raw]**
Repack to boot image: [new-boot.img]
HEADER_VER      [0]
**KERNEL_SZ       [18114732]**
RAMDISK_SZ      [5325552]
SECOND_SZ       [0]
EXTRA_SZ        [0]
OS_VERSION      [9.0.0]
OS_PATCH_LEVEL  [2019-05]
PAGESIZE        [4096]
NAME            []
```

Not working lates canary build
```
- Repacking boot image
CHECKSUM        [b8aa7cfc33c9db071a53a39d5aae8fd4233e6d9a000000000000000000000000]
KERNEL_DTB      [5140711]
**KERNEL_FMT      [lz4_lg]**
RAMDISK_FMT     [gzip]
Repack to boot image: [new-boot.img]
HEADER_VER      [0]
**KERNEL_SZ       [5140719]**
RAMDISK_SZ      [5325552]
SECOND_SZ       [0]
EXTRA_SZ        [0]
OS_VERSION      [9.0.0]
OS_PATCH_LEVEL  [2019-05]
PAGESIZE        [4096]
NAME            []
```
KERNEL_FMT      [lz4_legacy] | **KERNEL_FMT      [lz4_lg]**
EXTRA_FMT       [raw] |
KERNEL_SZ       [18114732]  | **KERNEL_SZ       [5140719]**

![Screenshot_20210125-222656](https://user-images.githubusercontent.com/1045996/105769354-fe6bb180-5f5d-11eb-873e-d20d9380e60c.png)

[org_boot.zip](https://github.com/topjohnwu/Magisk/files/5869823/org_boot.zip)
[magisk_patched_working.zip](https://github.com/topjohnwu/Magisk/files/5869856/magisk_patched_working.zip)

[...]

Issue #3722, opened 2021-01-17
Magisk newer than v21.0 can't properly patch the boot image


Lishoo:
Hi, I'v got a LG g6 h870 bootloader unlocked and magisk patched boot.img with version 20.4.
Some time ago I've got a notification that an update is available so I gave it a try, the version was 21.2 and I tried to install it directly (recommended), the patch ended with success but after reboot there was a bootloop, I still have the previously patched boot.img so I flashed it back and everything was ok.
Few days later I've got once again the notification that a newer version is available it was the v21.3, this time I've patched manually the boot.img and tried to flash it with the fastboot command, but the result was the same, still bootloop.

The original boot.img size is 22 632KB
The boot.img patched with the v20.4(working) is 22 854KB
The boot.img patched with the v21.2, v21.3 and the canary is always about 10 230 KB (+- 4KB depending what option are selected)

I'v tried to patch the stock boot.img and the magiskpatched_boot.img but the result is always the same.

![Screenshot_20210117-135655](https://user-images.githubusercontent.com/1045996/104844228-4a1fba80-58cf-11eb-8d63-a34152aeafd4.png)

[magisk_patched_working.zip](https://github.com/topjohnwu/Magisk/files/5826144/magisk_patched_working.zip)
[orginal_boot.zip](https://github.com/topjohnwu/Magisk/files/5826146/orginal_boot.zip)
[magisk_patched_7WZM0.zip](https://github.com/topjohnwu/Magisk/files/5826157/magisk_patched_7WZM0.zip)

[magisk_install_log_2021-01-17T133430Z.log](https://github.com/topjohnwu/Magisk/files/5826148/magisk_install_log_2021-01-17T133430Z.log)
[magisk_install_log_2021-01-17T134934Z.log](https://github.com/topjohnwu/Magisk/files/5826149/magisk_install_log_2021-01-17T134934Z.log)
[magisk_install_log_2021-01-17T135018Z.log](https://github.com/topjohnwu/Magisk/files/5826151/magisk_install_log_2021-01-17T135018Z.log)
[magisk_install_log_2021-01-17T135048Z.log](https://github.com/topjohnwu/Magisk/files/5826152/magisk_install_log_2021-01-17T135048Z.log)

I hope you can fix that issue soon :)

UPDATE
I just checked the version 21.0 and it boots up correctly, so I update the title, the 21.1 is the first that won't boot.

Frotty:
[...]

Issue #3635, opened 2021-01-02
LG G5 (H850) bootloop [H85030f]


DenyDarko:
So, this is an as clean as it gets installation (by formatting all partitions (including system, boot etc) on bootloader and using LGUP right after) and after magisk installation it simply won't boot when latest canary (as of today) is installed.

- The last successful working magisk was the release 20400.
- The firmware installed is the latest available: H85030f_00_OPEN_EU_OP_1122
- Encrypted or not encrypted storage makes no difference.

Bootloop video (nothing special):
https://user-images.githubusercontent.com/3320229/103456234-e3aa7200-4cfc-11eb-8c94-866bc4a8b776.mp4

Magisk installation logs & recovery logs:
[dmesg.log](https://github.com/topjohnwu/Magisk/files/5760166/dmesg.log)
[recovery.log](https://github.com/topjohnwu/Magisk/files/5760167/recovery.log)

Original boot IMG:
[boot_6.zip](https://github.com/topjohnwu/Magisk/files/5760151/boot_6.zip)

Patched (by magisk) boot IMG:
[boot.emmc.zip](https://github.com/topjohnwu/Magisk/files/5760153/boot.emmc.zip)

cat /proc/mounts output without root:
[mounts(NOROOT).txt](https://github.com/topjohnwu/Magisk/files/5760156/mounts.NOROOT.txt)

Now, I'm more than happy to provide any boot related logs, however, I am unable to find any of those. 
Check the video too, the loop is so fast (it doesn't even try) and it simply won't generate a thing.

etmatrix:
I confirm this issue. I don't understand because 21.1 now is a stable release. Today I tested and I got bootlop.
For me 21.2 is still beta. I rollback to 20.4.

japanesephundroid:
Same here on an LG V20 (US996) that shares many parts with the LG G5 and has similar software.
This happens with Android 8.0 Stock software.
21.0 beta works fine (I currently use it).
21.1 beta and 21.2 stable immediately bootloop.

See also:
https://www.reddit.com/r/lgv20/comments/kvcni5/magisk_212_stable/
Lineage versions of this phone seem to work fine.

topjohnwu:
I just released a new canary build, may you please try it out and report back?

japanesephundroid:
> I just released a new canary build, may you please try it out and report back?

I tried the latest Canary build, it bootlooped as well.
Give me a while and I will upload the logs.

[...]

PR #3812, opened 2021-02-03
Update format.hpp

Lishoo:
This fixes the boot.img patching bug which changes the size of the output magisk_patched.img and bootloop


Disappear9:
https://github.com/topjohnwu/Magisk/issues/3722
https://github.com/topjohnwu/Magisk/issues/3770
https://github.com/topjohnwu/Magisk/issues/3635
https://github.com/topjohnwu/Magisk/issues/3787

PR #3875, opened 2021-02-20
Update Italian translation

Fs00:
I couldn't find a good translation for "proxy app" in Italian that preserves its original meaning, so I kept it almost untranslated.
After all, to understand what "proxy" means in this context you should dig a bit into how the hiding mechanism works.


Fs00:
Will update the PR with the changed strings in 6809094 soon.

topjohnwu (approved):

PR #3969, opened 2021-03-02
Auto close issues

vvb2060:


Fox2Code:
I like this, this will help cleaning the issue section.

However issues with the `confirmed` label should not be closed, and feature requests should be possible to make.

But except that, I think this PR is good.

vvb2060:
@Fox2Code  only `open` action

yujincheng08 (approved):

topjohnwu (approved):

PR #4030, opened 2021-03-13
Clean up magisk sepolicies

aviraxp:
* Support deodexed ROM: This should not be done and dexpreopt is mandatory since P
   Xposed: Xposed handles them just fine, at least in the latest version 89.3
   suMiscL6: For whatever audio mods, a leftover of phh time

 * TODO: Are these policies for Chainfire's apps really need now?

 * Also cleanup binder sepolicies since we allow all binder transactions.

### DIFF
--- a/native/jni/magiskpolicy/rules.cpp
+++ b/native/jni/magiskpolicy/rules.cpp
@@ -154,7 +154,6 @@ void sepolicy::magisk_rules() {
     allow("servicemanager", SEPOL_PROC_DOMAIN, "file", "open");
     allow("servicemanager", SEPOL_PROC_DOMAIN, "file", "read");
     allow("servicemanager", SEPOL_PROC_DOMAIN, "process", "getattr");
-    allow("servicemanager", SEPOL_PROC_DOMAIN, "binder", "transfer");
     allow(ALL, SEPOL_PROC_DOMAIN, "process", "sigchld");
 
     // allowLog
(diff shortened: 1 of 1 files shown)


vvb2060 (approved):
Based on the principle of least privilege, Magisk should only have the necessary rules built in. Some ROM and apps need special rules that should be added by them.

osm0sis:
LiveBoot got a recent update and remains relevant, I believe.

aviraxp:
> LiveBoot got a recent update and remains relevant, I believe. I don't think those should be removed.

I tested on latest version of liveboot and these denials are not triggered.

aviraxp:
Commit updated.

Fox2Code:
Most modules install deodexed version of apk in `/system/(priv-,)app`, so these sepolicy rules may be still necessary.
An alternative way to solve this issue is to only apply these rules when required.

For example:
- if API 26/27 (Oreo) apply `Support deodexed ROM on Oreo` sepolicy rules
- if API 28 (Pie) and `/system/framework/com.samsung.device.jar` exists 
  apply `Support deodexed ROM on Pie (Samsung)` sepolicy rules
**Note:** policies are applied before modules mount so there are no risks of a module triggering the file check.

I think this is the best thing to do, except if these sepolicy rules are proven to be unnecessary.

aviraxp:
I think it is not related to apks, but jars in /system/framework.

Fox2Code:
We could drop Pie sepolicy rules and only keep Android Oreo sepolicy rules only when running on Android oreo.

I'm thinking something like:
```C++
    // This indicates API 26 or API 27 (Android Oreo)
    if (new_rules && !exists("untrusted_app_27")) {
        // Support deodexed ROM on Oreo
        allow("zygote", "dalvikcache_data_file", "file", "execute");
    }
```

But after all, it's supposed to be the job of the ROM maintainer to put the required sepolicy rules to make the ROM boot properly.  
So it also makes sense to remove these rules completely.

aviraxp:
ROM can boot without these sepolicies. It will just lag as hell.

Fox2Code:
It's also not the role of Magisk to fix ROM issues.
I think it's the job of the ROM maintainer to fix these issues.
But I also think some Oreo ROMs might be no longer maintained by their maintainer, 
this is why I suggested only applying these rules on Android Oreo ROMs.

I will let you make the decision on what you think is the best to do in this case, 
since I think you have all the information needed to make the right decision.

Or the third option could be to merge this PR as is and only add back the required
rules if Magisk Canary users report issues on their systems related to these changes.

As none of my devices would be affected by this change, I can't test it myself.

PR #4054, opened 2021-03-18
Ensure the installer knows the API version when running from addon.d

mincrmatt12:
`install_magisk` checks the API version to avoid running the bootsigner on older phones which do not support it. Although the variable it checks (`$API`) is set when it's called from the main Magisk installer, when running from the `addon.d` script the helper `api_level_arch_detect` is never called and so the variable is empty.

Although it still works in most cases, if the boot image was signed (and would have been resigned by the normal install process) the `addon.d` script fails to resign it after installation. Adding a call to `api_level_arch_detect` here fixes this problem (which was causing bootloops after OTA updates for me)


osm0sis (approved):

osm0sis:
Good catch! Thanks for the submission.

topjohnwu (approved):

Issue #4049, opened 2021-03-16
Segmentation fault during patching boot with integrated TWRP


Displax:
Device: Xiaomi Mi A1
Android version: 9
Magisk version name: f152b4c2
Magisk version code: 22005

Yes, I know Magisk shouldn't be perfectly compatible with integrated TWRP in boot.img. But anyway, this behavior has not happened before 22000, I think.
If wontfix - it's OK, but maybe someone take a look at this.

boot.img: https://drive.google.com/file/d/1Huw6LPW3bCLrngckWUCu7aGdOx5Ht_Dx/view?usp=sharing
log:
```
- Target image: /dev/block/mmcblk0p23
- Device platform: arm64-v8a
- Installing: f152b4c2 (22005)
- Unpacking boot image
Parsing boot image: [/dev/block/mmcblk0p23]
HEADER_VER      [1]
KERNEL_SZ       [9866617]
RAMDISK_SZ      [18271145]
SECOND_SZ       [0]
RECOV_DTBO_SZ   [0]
OS_VERSION      [9.0.0]
OS_PATCH_LEVEL  [2020-08]
PAGESIZE        [2048]
NAME            []
CMDLINE         [console=ttyHSL0,115200,n8 androidboot.console=ttyHSL0 androidboot.hardware=qcom msm_rtb.filter=0x237 ehci-hcd.park=3 lpm_levels.sleep_disabled=1 androidboot.bootdevice=7824900.sdhci earlycon=msm_hsl_uart,0x78af000 buildvariant=user veritykeyid=id:5560e7863b4d8118c2f1b065595cf93bb2447992]
CHECKSUM        [b5a3f652ac387570269b5683ff05ae4004b3324e000000000000000000000000]
Segmentation fault
- Checking ramdisk status
- Stock boot image detected
- Patching ramdisk
Add entry [init] (0750)
Create directory [overlay.d] (0750)
Create directory [overlay.d/sbin] (0750)
Add entry [overlay.d/sbin/magisk32.xz] (0644)
Add entry [overlay.d/sbin/magisk64.xz] (0644)
Patch with flag KEEPVERITY=[true] KEEPFORCEENCRYPT=[true]
Create directory [.backup] (0000)
Add entry [.backup/.magisk] (0000)
Dump cpio: [ramdisk.cpio]
Parsing boot image: [/dev/block/mmcblk0p23]
- Repacking boot image
HEADER_VER      [1]
KERNEL_SZ       [9866617]
[...]

PR #4075, opened 2021-03-21
Fix install regressions and a few other fixes/improvements

### LINKED ISSUE #3119: No full clean after customize.sh
It looks like module_installer.sh does not clean everything correctly when customize.sh with SKIPUNZIP=1 is used.

To reproduce it try the following:
1. Create a simple module with pure customize.sh file with only one string SKIPUNZIP=1. I have attached that one.
[test.zip](https://github.com/topjohnwu/Magisk/files/5152959/test.zip)
2. Install it.
3. Reboot.

In theory this addon simply does nothing and no new modules should appear in Magisk as well as no new folders should be created in /data/adb/modules/
But it is not true in practice - /data/adb/modules/ contains this new module.

1. Installation log file: 
[magisk_install_log_2020-09-01T01_40_54Z.log](https://github.com/topjohnwu/Magisk/files/5152968/magisk_install_log_2020-09-01T01_40_54Z.log)
Pay attention on cp error.
2. New module folder in /data/adb/modules/ (archived, in fact - contains empty update file only): 
[testtesttest.zip](https://github.com/topjohnwu/Magisk/files/5152972/testtesttest.zip)
3. New empty item in /data/adb/modules_update (archived, in fact - empty): 
[modules_update.zip](https://github.com/topjohnwu/Magisk/files/5152975/modules_update.zip)

Looks like some mechanism that should completely wipe off the module folder works incorrectly.

### LINKED ISSUE #3961: 64-bit detection not reliable
Hi,

d2bc2cfcf87e4506569b0c6904b6213297f45646 installs the 64-bit Magisk binary into the initramfs if it detects a `/system/lib64` directory. On my [TV](https://github.com/yath/tpm171e), however, the kernel is an `armv7l` one and cannot execute 64-bit binaries, yet there is a `/system/lib64` directory (with only an `arm64` `libnanovoice.so`, I dunno why that is :shrug:). This results in a `magisk64.xz` binary installed into the initrd, and init failing to execute it (“<11>[   58.786037] init: cannot execve('/sbin/magisk'): Exec format error”).

I don’t know of a good way to solve this, otherwise I’d send a PR. I guess the most reliable way (apart from guessing from `uname -m`) would be attempting to execute the 64-bit binary in `boot_patch.sh` and if it does not succeed, set `IS64BIT` to false. init could alternatively fall back to the 32-bit version if exec fails with `ENOEXEC`.

Thanks for your work!\
Sebastian

osm0sis:
scripts: clean up remaining Manager references

scripts: fix find_boot_image using wrong partition list on non-slot
- revert logic changes introduced by ec8fffe which break find_boot_image on NAND devices and any others using non-standard naming supported by the A-only device boot partition name list
- despite being accepted equivalents in modern shells -n does not work on Android in some shell/env scenarios where ! -z always does

scripts: add boot_patch unpack error catching
- failure to unpack wasn't being caught so boot_patch would continue to build a new cpio out of nothing and attempt to repack it (identified in #4049)

scripts: improve boot_patch 64bit detection
- use existing api_level_arch_detect function
Fixes #3961

App: add versionCode to magisk_patched.img filenames

scripts: fix empty module cleanup
- should be sufficient for all basic modules, see https://github.com/topjohnwu/Magisk/issues/3119#issuecomment-704000783 for ideas for fixing it further on the daemon module-processing side
Fixes #3119

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -255,7 +255,7 @@ abstract class MagiskInstallImpl protected constructor(
                 val alpha = "abcdefghijklmnopqrstuvwxyz"
                 val alphaNum = "$alpha${alpha.toUpperCase(Locale.ROOT)}0123456789"
                 val random = SecureRandom()
-                val filename = StringBuilder("magisk_patched_").run {
+                val filename = StringBuilder("magisk_patched-${BuildConfig.VERSION_CODE}_").run {
                     for (i in 1..5) {
                         append(alphaNum[random.nextInt(alphaNum.length)])
                     }
--- a/app/src/main/java/com/topjohnwu/magisk/ui/settings/SettingsItems.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/ui/settings/SettingsItems.kt
@@ -87,7 +87,7 @@ object Hide : BaseSettingsItem.Input() {
         get() = if (isError) null else result
 
     @get:Bindable
-    var result = "Manager"
+    var result = "Settings"
         set(value) = set(value, field, { field = it }, BR.result, BR.error)
 
     val maxLength
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -1,5 +1,5 @@
 ##################################
-# Magisk Manager internal scripts
+# Magisk app internal scripts
 ##################################
 
 run_delay() {
@@ -115,8 +115,8 @@ id=hosts
 name=Systemless Hosts
 version=1.0
 versionCode=1
-author=Magisk Manager
-description=Magisk Manager built-in systemless hosts module
+author=Magisk
+description=Magisk app built-in systemless hosts module
 EOF
   magisk --clone /system/etc/hosts hosts/system/etc/hosts
   touch hosts/update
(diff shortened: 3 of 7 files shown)


dwdjanika (approved):
- [x]

app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
```diff
@@ -255,7 +255,7 @@ abstract class MagiskInstallImpl protected constructor(
                 val alpha = "abcdefghijklmnopqrstuvwxyz"
                 val alphaNum = "$alpha${alpha.toUpperCase(Locale.ROOT)}0123456789"
                 val random = SecureRandom()
-                val filename = StringBuilder("magisk_patched_").run {
+                val filename = StringBuilder("magisk_patched-${BuildConfig.VERSION_CODE}_").run {
```
vvb2060, line 258: `magisk_patched-`  → `magisk_patched_`

osm0sis: I've looked at it both ways and I think - looks better there:

magisk_patched_22005_31P8J.img
vs.
magisk_patched-22005_31P8J.img

Plus it matches the APK naming in a way: Magisk-f152b4c2(22005).apk

dwdjanika (approved):

topjohnwu (approved):

vvb2060:
https://github.com/topjohnwu/ndk-busybox/blob/android-ndk-1.34.1/coreutils/test.c

busybox supports -n

osm0sis:
We've discussed this since in another PR or issue and discovered what the actual issue was with -n, yes

PR #4076, opened 2021-03-22
Update Japanese translations

Kaz205:


topjohnwu (approved):

Kaz205:
@topjohnwu I changed the "yes" string so that it'd mean "supported" and vice versa. Therefore it might break if you use the string in other occasions. Should I revert the change?

PR #4094, opened 2021-03-25
Fix spelling in French translation

tbertels:
Restorer -> Restaurer


JordanViknar:
I'd like also to mention a good one : "Attestion accordée !" in Magisk 22.0
It should be "Attes**ta**tion accordée !".

Seems like someone didn't do their French homework, did they ?

topjohnwu (approved):

Issue #3735, opened 2021-01-18
Magisk Wrong sepolicy.rule Directory Installation


reiryuki:
Magisk app since 8.0.7 (4834) until now and Magisk zip v21.4 until the latest sepolicy.rule bug

If I toggling disable/enable any module or remove/restore, or installing a module with Magisk app, sepolicy.rule files are installing to /cache/magisk but the sepolicy.rules folder is linked to /sbin/.magisk/mirror/persist/magisk. It doesn't syncronized. I have to move the /cache/magisk folder to /sbin/.magisk/mirror/persist/ manually because of it.

$RULESDIR detecting sepolicy.rule directory is /cache/magisk, but actually is /mnt/vendor/persist/magisk or /sbin/.magisk/mirror/persist/magisk for my device (R4A)

I modify util_functions.sh, moving the "Remove all existing rule folders (rm -rf ...)" to the after $RULESDIR detection. It's fixing for module installation but app toggler still bug. I think it's no longer using utilfunctions.sh for those toggler.

Looks like the $active_dir is not detected after rm -rf. Placing the rm -rf before detecting link is a faulty. The "Remove all existing rule folders (rm -rf ...)" should be placed after detecting $RULESDIR, not before it, so the $active_dir can be detected correctly..

Thank you

![Screenshot_20210118-160646612](https://user-images.githubusercontent.com/46316533/104896161-58420980-59a9-11eb-88ad-02cd2645bf46.jpg)

![Screenshot_20210118-160653636](https://user-images.githubusercontent.com/46316533/104896186-642dcb80-59a9-11eb-9094-a4a77173a18c.jpg)
[logcat_01-18-2021_16-07-15.txt](https://github.com/topjohnwu/Magisk/files/5829135/logcat_01-18-2021_16-07-15.txt)

![Screenshot_20210119-114412178](https://user-images.githubusercontent.com/46316533/104989058-2df55800-5a4c-11eb-8e56-5cd5ff798b7e.jpg)

osm0sis:
Similar to this I had noticed recently that sepolicy.rule files are somehow being copied to both /metadata/magisk and /mnt/vendor/persist/magisk on my OP8T. The way the copy_sepolicy_rules function reads there should only be one location. :thinking_face:

vvb2060:
https://github.com/topjohnwu/Magisk/commit/0783f3d5b6d46d0a0ed8897de4c8027838af1688 I think it's fixed now.

[...]

PR #4095, opened 2021-03-25
Fix copying sepolicy.rule 

### LINKED ISSUE #4098: sepolicy wrong status on latest canary
On latest version its starts showing On LSPosed that message.
Screenshot_20210325-143723_LSPosed

API version
93

Framework version
v1.3.4 (5501)

Manager version
v1.3.4 (5501)

System version
11 (API 30)

Device
Samsung SM-G996B

System ABI
arm64-v8a

Device: SM-G996B
Android version: 11
Magisk version name: 66e30a77
Magisk version code: 22006

aviraxp:
* Somehow commit 8b5cb4c causes sepolicy.rule being linked
   to /metadata/magisk which does not even exists.(?!) Reverting the
   change fixes the behavior, though a reflash of affected modules
   is needed.

Close #4098

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -594,7 +594,7 @@ copy_sepolicy_rules() {
   # Find current active RULESDIR
   local RULESDIR
   local active_dir=$(magisk --path)/.magisk/mirror/sepolicy.rules
-  if [ -L $active_dir ]; then
+  if [ -e $active_dir ]; then
     RULESDIR=$(readlink -f $active_dir)
   elif [ -d /data/unencrypted ] && ! grep ' /data ' /proc/mounts | grep -qE 'dm-|f2fs'; then
     RULESDIR=/data/unencrypted/magisk


aviraxp:
This is not an ideal solution but a band aid. I don't know why this happens, maybe @osm0sis can provide ideas?

aviraxp:
Seems correct behavior is:(/dev/xxx... is omitted)

/mirror/matadata/magisk exists
symlink under /mirror does not exist

Wrong behavior is totally the opposite.

aviraxp:
Seems we should go into metadata path, but  -L makes us fall into this path, but sepolicy is not copied at all.

osm0sis:
Reverting this brings back the broken behavior of #3735, so yes, something different will need to be figured out.

aviraxp:
Turns out readlink is broken for me. readlink -f $(magisk --path)/.magisk/mirror/sepolicy.rules just returns itself rather than metadata one. ls -l points correct output however.

osm0sis:
I think it would still need to be -fm so the full path is displayed.

aviraxp:
Nice, busybox has a broken impl on readlink.

https://git.busybox.net/busybox/commit/coreutils/readlink.c?h=1_32_1&id=1b49c25e0a719ec3051eafa2329e68012c815abb

So -f is actually enough, but who would have thought busybox is broken?

Another issue isnbusybox readlink does not have -m, so what can we do? Relying on toybox (but break ancient old Android) or sth?

PR #4104, opened 2021-03-26
Don't use readlink when getting sepolicy rules dir

### LINKED ISSUE #4098: sepolicy wrong status on latest canary

aviraxp:
* `readlink -f` implementation of busybox is broken because
   its `-f` is actually `-e` which requires all components to exist,
   which is not the case here and results in wrong output. We
   can pass `-m` here instead, but busybox does not implement 
   this argument. To work around this issue, use `ls -l` which
   can generate correct path. This closes #4098.

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -595,7 +595,7 @@ copy_sepolicy_rules() {
   local RULESDIR
   local active_dir=$(magisk --path)/.magisk/mirror/sepolicy.rules
   if [ -L $active_dir ]; then
-    RULESDIR=$(readlink -f $active_dir)
+    RULESDIR=$(ls -l $active_dir | awk -F'->' '{print $2}' | tr -d ' ')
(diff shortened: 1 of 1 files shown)


aviraxp:
I think it is a bit overhead to implement `readlink -m` for busybox (seems no one did it before). Toybox has no issues (https://github.com/landley/toybox/commit/cd3f81ebe5c82f1fd5f610b0f2e0fa07ac27903e) but Android 9+ is required.

PR #4113, opened 2021-03-26
resetprop: load_prop_file: delete props before setting them

Linux4:
* In some cases not deleting a prop before setting it can lead
  to the prop being duplicated rendering the device unbootable
* An example for such a case is the 'Magisk Hide Props Config'
  module, if ro.build.fingerprint is not set in the device's
  /system/build.prop but instead being autogenerated at runtime
  setting a build fingerprint with this module will break booting.


mu198:
thanks for this fix, could finally use my build magisk HideProps without getting into a permanent bootanimation.
@topjohnwu  I hope they will integrate this fix.

PR #4130, opened 2021-03-30
Update instructions for some type III devices

Didgeridoohan:
Added info that some devices (mainly Xiaomi) need to patch the boot image even though there's no ramdisk in the boot partition.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -19,7 +19,9 @@ Download and install the latest Magisk app. We use the app to gather some inform
 
 Pay special attention to the **Ramdisk** info. If the result is **Yes**, congratulations, your device is perfect for installing Magisk! However, if the result is **No** this means your device's boot partition does **NOT** include ramdisk. This means you will have to go through some extra steps to make Magisk work properly.
 
-> **If your device does not have boot ramdisk, read the [Magisk in Recovery](#magisk-in-recovery) section after installing. The information in that section is VERY important!**
+Unfortunately there are exceptions to this, where you will have to follow the instructions as if your device's boot partition does include ramdisk even if it doesn't. The only way to know for sure is to try, so if you cannot seem to get Magisk to install by patching the recovery image when the **Ramdisk** info reports **No**, try patching the boot image instead. Some Xiaomi devices are known to have this setup.
+
+> **If your device does not have boot ramdisk and has to be installed to recovery, read the [Magisk in Recovery](#magisk-in-recovery) section after installing. The information in that section is VERY important!**
 
 If you are using a Samsung device and the **SAR** result is **Yes**, please check [its own section](#samsung-system-as-root).
 
@@ -66,7 +68,7 @@ If you insist on using custom recoveries, rename the Magisk APK to `uninstall.zi
 
 ## Magisk in Recovery
 
-If your device does not have ramdisk in boot images, Magisk has no choice but to be installed in the recovery partition. For these devices, you will have to **reboot to recovery** every time you want Magisk.
+If your device does not have ramdisk in boot images, it is very likely that Magisk has no choice but to be installed in the recovery partition. For these devices, you will have to **reboot to recovery** every time you want Magisk.
 
 When Magisk is installed in your recovery, **you CANNOT use custom recoveries to install/upgrade Magisk!** The only way to install/upgrade Magisk is through the Magisk app. It will be aware of your device state and install to the correct partition and reboot into the correct mode.
 


pndwal:
Nice pull request!

Re. first appended section, may I suggest, for accuracy and clarity:

"However, there are exceptions to this where patching and flashing boot image (as if your device's boot partition does include ramdisk) works because the assumption that bootloader is incompatible where OEM doesn't supply ramdisk in boot partition may not be true.

Magisk will add needed ramdisk functionality to boot image where lacking, but Magisk will continue to indicate factory ramdisk state, ie. 'no' to indicate that bootloader support cannot be relied on. However, the only way to know if bootloader does or does not support the Magisk-added ramdisk in boot is to try. It will work on most recent Xiaomi devices for example.

While Magisk should install in recovery partition by patching / flashing also in these cases, installing in boot will be a better alternative.

> **If your device does not support boot ramdisk, Magisk must be installed to recovery. Read the [Magisk in Recovery](#magisk-in-recovery) section for VERY important information!**"

Regards PW

topjohnwu:
@Didgeridoohan for those using Xiaomi, using recovery images should still be a working solution. It is just that the have the *option* to use the boot image instead. Am I right?

Didgeridoohan:
@topjohnwu No. It's actually one of the more common query on XDA. Xiaomi users coming in and asking why they can't get root working when they follow the instructions. But when they use the boot image instead, things work.

pndwal:
"@Didgeridoohan for those using Xiaomi, using recovery images should still be a working solution. It is just that the have the option to use the boot image instead. Am I right?"
Just my 2c worth:

Yes, but they won't know this since instructions are misleading at present, stating "If your device does not have ramdisk in boot images, Magisk has no choice but to be installed in the recovery partition. For these devices, you will have to reboot to recovery every time you want Magisk."

In fact, as long as there is bootloader support, Magisk in boot is not only a choice, but a better one (thanks to your commit adding basic ramdisk).

Also, while Magisk in boot seems to work in almost every case for Xiaomi, some users (eg. Redmi 7A) report needing 3rd party patching (eg https://github.com/aAlex3/patcher-oss ) to fix ramdisk. (eg https://forum.xda-developers.com/t/magisk-general-support-discussion.3432382/post-84806507 )

Hope it helps, PW

pndwal:
FWIW, I found this explanation for Magisk manual initramfs patching not working for some Xiaomi devices like R7A, while 3rd party patching (mentioned above) works:

"Xiaomi installed a 32-bit operating system because the smartphone only has 2 GB of RAM. Not enough to work properly with a 64 bit operating system. Topjohnwu published the information on its Github page and criticized Xiaomi very much for this behavior."
https://forum.xda-developers.com/t/redmi-7a-upgrade-to-magisk-v22-issues.4257161/post-84793607

"The problem is, the boot.img is a 32 bit img with a fake Ramdisk for a 64 bit processor."
https://forum.xda-developers.com/t/redmi-7a-upgrade-to-magisk-v22-issues.4257161/post-84789613

Nb. The 3rd party method only adds proper (32 bit?) ramdisk support to boot image. Magisk root can be successfully applied after this via zip Installation from TWRP. (I assume patching image from Magisk app would also work.) PW

dueringa:
What may also be added to the instructions: If the Magisk App reports "Ramdisk: No", "A/B device: No", although you *do* have a A/B device, and the extracted payload.bin from the OTA image only contains a boot.img, I think your instruction apply as well (OnePlus 6T?)

pndwal:
> What may also be added to the instructions: If the Magisk App reports "Ramdisk: No", "A/B device: No", although you _do_ have a A/B device, and the extracted payload.bin from the OTA image only contains a boot.img, I think your instruction apply as well (OnePlus 6T?)

Thats an interesting case.

Devices discussed above are Type III devices, covered here:
https://topjohnwu.github.io/Magisk/boot.html

Any device where Magisk reports "Ramdisk: No", "A/B device: No" would appear to be type III.

If OnePlus 6T is in fact an A/B device, it may be unaccounted for here, since these devices should apparently at least have Recovery ramdisk in boot partition. PW

dueringa:
I forgot to add, though, that I'm using a custom ROM. That changes the situation again according to the article.

Apr 11, 2021 6:04:40 AM pndwal ***@***.***>:

pndwal:
> @Didgeridoohan for those using Xiaomi, using recovery images should still be a working solution. It is just that the have the _option_ to use the boot image instead. Am I right?

Just tested Magisk in recovery on my RN8T (Ramdisk = no) - hadn't previously as Magisk in boot always worked.

Surprisingly, any attempt to flash Magisk to recovery finds recovery's 'stock boot image' and patching succeeds, but BOOTING FAILS with 'System Has Been Destroyed'; Fastboot only available.(flashing stock recovery fixes this).

I've tried builds from latest back to 20.3 with same results... Will try patching custom recovery.img and report soon.

I'm not sure what the problem is with recovery, but I understand other Xiaomi users have also failed to gain root by official Ramdisk = no method also. Fortunately, it generally works with boot-ramdisk. Please let me know if you'd like logs etc. to diagnose further.

Hope this helps. PW

Didgeridoohan:
@pndwal Yes, that is the kind of results that I've seen reported over and over and is why I felt that the instructions needed a note about this.

I'm just going by what I've seen posted here an there though, I have no real-world experience with Xiaomi, so can't do that kind of testing.

pndwal:
> @pndwal Yes, that is the kind of results that I've seen reported over and over and is why I felt that the instructions needed a note about this.
> 
> I'm just going by what I've seen posted here an there though, I have no real-world experience with Xiaomi, so can't do that kind of testing.

Agreed; I've thought more accurate info. in / linked from Installation Instructions on this would be very useful for some time now also.

I went straight to working Magisk in boot method (zero frustration) as I keep up with methods posted on John's XDA, but now that I understand Official Magisk in recovery method is fatally flawed and doomed from the outset for possibly the majority of Xiaomi modders (and have tested this as above), it seems an imperative to spare users avoidable angst.

zgfg (approved):
Yes guys, many Xiaomi users are unfortunately misslead by the existing instructions, hence I also agree that special section for Xiaomi (like for Samsung or Huawei) will be welcome

I never actually tried to patch the recovery img on my Mi 9T, but I was successfully patching the boot.img through all stock upgrades since MIUI 10/Android 9 till MIUI 12/Android 10, as well as the beta firmwares MIUI 12.1/12.5/12.6/Android 11

And by watching the XDA forums for Mi 9, Mi 9Se, Mi 9T Pro, I've seen that patching the boot.img works on those devices as well, while users who strictly followed the official Magisk Installation Instructions and therefore tried to patch the recovery.img report failures, ask for help, etc

pndwal (changes requested):
Nice pull request!

Re. first appended section, may I suggest, for accuracy and clarity:

"However, there are exceptions to this where patching and flashing boot image (as for device with ramdisk in boot partition) works because the assumption that bootloader is incompatible where OEM doesn't supply ramdisk in boot may not be true.

Magisk will add needed ramdisk functionality to boot image where lacking, but Magisk app will continue to indicate factory ramdisk state, ie. 'no', to indicate that bootloader support cannot be relied upon. The only way to know for sure whether bootloader does or does not support the Magisk-added ramdisk is to try patching / flashing boot image. Notably, this will work on most recent Xiaomi devices.

While Magisk may (possibly) install in recovery partition by patching / flashing recovery also in these cases, installing in boot will be a better alternative.

If your bootloader does not support boot ramdisk, Magisk must be installed to recovery. Read the Magisk in Recovery section for VERY important information!"

Regards, PW

topjohnwu:
@Didgeridoohan I have covered the same info in my updated instructions (4f73534837bda59fd58ca010fe25d3d66556883e). Thank you :)

Issue #4103, opened 2021-03-25
No root installed magisk N/A on all versions since 22001 on Intel x86 Android 6.0.1 patched boot.img


martyfender:
Device: nextbook ares 8a intel x86
Android version:6.0.1
Magisk version name: canary 22006
Magisk version code: 22006

I save the patched boot.img from 22001 and it works fine. All newer versions show magisk as installed as na. 
Attached is a twrp recovery log, but the only magisk seen in twrp file manager is from 22001, so it appears that no new magisk log is being created after flashing and rebooting the system. 
[recovery.log](https://github.com/topjohnwu/Magisk/files/6208239/recovery.log)
[boot.zip](https://github.com/topjohnwu/Magisk/files/6208250/boot.zip)

I also tried direct install and flashing suceeds but upon roboot no root is installed, in magisk manager it also shows installed N/A.

Attached is the magisk.log saved in mm from the working canary 22001, if that may be useful. 
[magisk_log_2021-03-25T19:15:34Z.log](https://github.com/topjohnwu/Magisk/files/6208380/magisk_log_2021-03-25T19.15.34Z.log)

UPDATE: attached is a boot logcat after flashing 22006.
[bootlogcat.txt](https://github.com/topjohnwu/Magisk/files/6208620/bootlogcat.txt)

All of the previous versions since 22001 work fine on my 2, Onn 8, Android 9 devices, using the patched boot.img method and flashing in TWRP.
Thanks

shakalaca:
any log for launching Magisk app ?

Haus3master:
In not sure about x86, but, the issue you mentioned previously happened on arm64-v8a devices, but, this was fixed about 2-3 updates ago...

martyfender:
> 
> 
> any log for launching Magisk app ?
[logcat.txt](https://github.com/topjohnwu/Magisk/files/6212018/logcat.txt)

shakalaca:
> > any log for launching Magisk app ?
> > [logcat.txt](https://github.com/topjohnwu/Magisk/files/6212018/logcat.txt)
> 
> Is this what you mean? This is an adb logcat where I launched the magisk app then closed it.

yeap, but I didn't see the log of su handler. Can you show me the result of `pm dump com.topjohnwu.magisk|grep CpuAbi` ?

martyfender:
> > > any log for launching Magisk app ?
[...]

PR #4142, opened 2021-04-03
Fix random return value of faccessat() in x86

shakalaca:
faccessat() should return 0 when success, but it returns random number with errno == 0 in x86 platform.

It’s a side effect of commit bf80b08b5febd3f311108ff1c6e6a6bf2d1113b7 when Magisk binaries ‘correctly’ linked with library of API16 .. lol


shakalaca:
may fix #4103

topjohnwu (approved):

PR #4148, opened 2021-04-04
Update gradle config & dependencies

Goooler:


Goooler:
```
android.databinding.incremental=true
kapt.incremental.apt=true
```
These two properties have been enabled  by default, could be removed.

image
image

build.gradle.kts
```diff
@@ -10,16 +10,15 @@ plugins {
 buildscript {
     repositories {
         google()
-        jcenter()
```
topjohnwu, line 13: Removing jcenter will not work properly because `koin` v2 is only available on jcenter.

Goooler: Koin have not been `classpath`ed in `buildscript`'s dependencies, subproejcts `implementation` it with jcenter.

Goooler: And I just moved `subprojects` repositories into `allprojects` closure

image

Goooler:
@topjohnwu Gradle 7.0 is a big release, I think we can merge this PR now.

topjohnwu:
I believe this is already addressed in some changes

PR #4153, opened 2021-04-05
Prevent overriding API variable if already set

aviraxp:
* We have already exported $API when initializing shell, doing
   this again will clear it if device's build.prop does not contain
   ro.build.version.sdk, as seen on oppo devices.

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -508,7 +508,7 @@ remove_system_su() {
 }
 
 api_level_arch_detect() {
-  API=`grep_prop ro.build.version.sdk`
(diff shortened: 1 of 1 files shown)


aviraxp:
Should fix https://github.com/LSPosed/LSPosed/issues/457.

osm0sis:
Wouldn't it be better to just use [ -z $API ] instead of $BOOTMODE, so that it'll only ever attempt it if it isn't already populated?

osm0sis:
Drop the `if ` and you're good. 🙂

osm0sis (approved):

joker2326 (approved):

Issue #3963, opened 2021-03-01
Broken display logic for "Install to Inactive Slot (After OTA)" option


shugo-chara:
![Screenshot_2021-03-01-10-38-23-579_l oybq yjfpfvi c](https://user-images.githubusercontent.com/20610965/109442075-7d329d80-7a7a-11eb-9a9a-75d291889465.png)
mi11
miui12.5
After installing magisk22.0/22001, the "install to another slot" option disappears.
Reinstall magisk and system still can't enable this feature.

DEVICE DESCRIPTION
Model: Mi11
Android version: 11
System build version: 21.2.25

MAGISK SETUP DESCRIPTION
Manager version: 22.0/22001
Installation method: patching boot image/install
![Screenshot_2021-03-01-10-36-05-143_l oybq yjfpfvi c](https://user-images.githubusercontent.com/20610965/109441937-1a410680-7a7a-11eb-87e7-3ed1b951366c.png)

Displax:
Because it's not supported on Android 11. Please close this.

vvb2060:
Did it work properly before?  Mi 11 is a virtual A/B device, we believe that virtual A/B does not support installation to another slot.

osm0sis:
I have told you it's NOT correctly hiding it on Virtual A/B devices either @vvb2060, so sounds like it may also be hiding it on devices it shouldn't either?

shugo-chara:
> Did it work properly before? Mi 11 is a virtual A/B device, we believe that virtual A/B does not support installation to another slot.
I was using this method to install magisk until I upgraded to 22. The Mi11 is a virtual a/b, but it also works fine with the "install to another slot" feature.
Unfortunately after upgrading to 22, he disappeared. I have tried many ways to find out why. However, some users of this model seem to be able to use it normally. I don't know what the problem is.
@osm0sis

vvb2060:
@shugo-chara  Maybe the SELinux of your device is permissive, so Magisk app detected the virtual A/B.
@osm0sis Maybe only OnePlus virtual A/B devices cannot use "install to other slots".?

osm0sis:
@vvb2060 Pixel devices on Virtual A/B also don't work with "Install to Inactive Slot", which is why we originally hid it on them specifically.

@shugo-chara Please provide the output of `getprop | grep virtual` at both a user and root prompt.

Here's mine:
```
:/ $ getprop | grep virtual
[persist.sys.wfd.virtual]: [0]
[...]

PR #4170, opened 2021-04-09
Revert "Hide OTA option on virtual A/B devices"

aviraxp:
* Does not work because of SELinux, and installation on virtual
   A/B devices work just fine.

This reverts commit 00659e4795dd985b0bbb1b630d089116b9da9e9c.

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/Info.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/Info.kt
@@ -32,7 +32,6 @@ object Info {
     @JvmStatic val env by lazy { loadState() }
     @JvmField var isSAR = false
     @JvmField var isAB = false
-    @JvmField val isVirtualAB = getProperty("ro.virtual_ab.enabled", "false") == "true"
     @JvmStatic val isFDE get() = crypto == "block"
     @JvmField var ramdisk = false
     @JvmField var hasGMS = true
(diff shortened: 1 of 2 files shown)


osm0sis:
It does not work on any Virtual A/B device that uses Google Play services mediated Incremental OTAs, which is at least Pixels, OnePlus 8T and newer in North America, and probably others.

You're correct that the option hiding is broken however, since it still shows on my 8T erroneously.

See: https://github.com/topjohnwu/Magisk/issues/3963

aviraxp:
Can you provide more information about Google Play services mediated Incremental OTAs? I am using OP8T and it works with the default updater.

osm0sis:
It works for you because you're presumably not using the North American variant? Here's the lengths I've worked out that needed to be gone through to retain SafetyNet/Certified passing and not bricking it post-OTA: https://forum.xda-developers.com/t/tools-zips-scripts-osm0sis-odds-and-ends-multiple-devices-platforms.2239421/page-149#post-84764713

osm0sis:
And a bit more supplemental info here: https://twitter.com/osm0sis_xda/status/1378019206466314244

ameaninglessname:
Would this commit be merged soon?

osm0sis:
It's not really sufficient. The logic is overall still broken even with this revert.

osm0sis:
I think just whatever effect SELinux has on it needs to be fixed. Someone reported it only showed when SELinux was enforcing for some reason. 🤔

osm0sis (approved):

osm0sis:
Don't think this has been resolved yet, no?

aviraxp:
Nope, but I am thinking of other mechanisms, such as a warning dialog.

aviraxp:
This commit does not have any issues at all.

@osm0sis the issue lies in install instruction. You need to either turn on 'Auto update after reboot' in dev options or reboot with Google play update dialog. Without this, update engine won't do dm-verity related work and cause bootloop.

PR #4212, opened 2021-04-13
Update french translations

### LINKED ISSUE #4210: In order to fix FR translations : What is the difference between "OS apps" and "System apps" ?
Device: Any
Android version: Any
Magisk version name: 22.1
Magisk version code: 22102

Both [this translation](https://github.com/topjohnwu/Magisk/blob/c4e8dda37cd932049dbf637f47a150ecd08705f0/app/src/main/res/values-fr/strings.xml#L101) and [this one](https://github.com/topjohnwu/Magisk/blob/c4e8dda37cd932049dbf637f47a150ecd08705f0/app/src/main/res/values-fr/strings.xml#L102) mean *Show System apps*.

The only difference between them is that the first one would translate to "System applications" whereas the second one would be "System apps". 

So apart from the way the word application is written, there is no clear difference here between OS apps and System apps.

So how do you differentiate OS apps from System apps

cl-ement05:
Especially to fix #4210 and more

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -148,7 +148,7 @@
     <string name="settings_update_custom_msg">Saisissez une URL personnalisée</string>
     <string name="settings_magiskhide_summary">Rendre Magisk invisible face à diverses formes de détection.</string>
     <string name="settings_hosts_title">Fichier hôtes hors partition système</string>
-    <string name="settings_hosts_summary">Utilisation d\’un fichier hôtes hors de la partition système pour les applications de blocage de publicité.</string>
+    <string name="settings_hosts_summary">Utilisation d’un fichier hôtes hors de la partition système pour les applications de blocage de publicité.</string>
     <string name="settings_hosts_toast">Ajout d’un module pour fichier hosts hors système</string>
     <string name="settings_app_name_hint">Nouveau nom</string>
     <string name="settings_app_name_helper">L’application sera réempaquetée sous ce nom</string>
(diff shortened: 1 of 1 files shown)


cl-ement05:
I just don't understand why check are not passing (build app fails). I only changed an xml file with strings

androidacy:
```
/home/runner/work/Magisk/Magisk/app/src/main/res/values-fr/strings.xml:242:5-101: AAPT: error: unescaped apostrophe in string
```

androidacy:
You need to escape `\` things like '  and "

osm0sis:
Why are you changing English values when you say you're updating the French translation?

cl-ement05:
> Why are you changing English values when you say you're updating the French translation?
> 
> Edit: Haha never mind, Google Translate was pretty seamless and even changed fr to en in the path 😅

Lol I don't understand. Where did I change english values ??

cl-ement05:
> You need to escape `\` things like '  and "

Oh my bad yes I totally forgot to escape commas

cl-ement05:
> 
> 
> You need to escape `\` things like ' and "

The strange thing here is that some comas are not escaped in this file (see [this example](https://github.com/cl-ement05/Magisk/blob/291a2f84efc9ad539022b7be683f61b8df89819f/app/src/main/res/values-fr/strings.xml#L54)). However, build succeeds but how ?

topjohnwu:
@cl-ement05 those are Unicode quotes, not the same a the ASCII quotes 😉

topjohnwu (approved):

PR #4283, opened 2021-04-26
Don't kill USAP processes for now

aviraxp:
* It is observed that every application fails to launch after
   enabling MagiskHide with the toggle in the app. After
   debugging, it seems that `usap` processes are killed in
   `init_list()` when starting MagiskHide, and zygote fails
   to re-create these processes. This results in zygote
   consuming 100% CPU (possibly runs in a dead loop).
   As a band-aid, for now, don't kill USAP processes.

 * Tested on Oneplus8T Open Beta 3.

### DIFF
--- a/native/jni/magiskhide/hide_utils.cpp
+++ b/native/jni/magiskhide/hide_utils.cpp
@@ -241,8 +241,6 @@ static bool init_list() {
 
     // If Android Q+, also kill blastula pool and all app zygotes
     if (SDK_INT >= 29) {
-        kill_process("usap32", true);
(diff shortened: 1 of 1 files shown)


aviraxp:
I would like to let the maintainer see what the real issue is as I am no expert in this. I tried runn
[++.txt](https://github.com/topjohnwu/Magisk/files/6374500/%2B%2B.txt)
ing `pkill -9 usap32(64)` and new USAP processes started without issues. I don't know why the case is different with Magisk.

Issue #4225, opened 2021-04-15
[v22.1+] The busybox binary is messed


seffparker:
After upgrading from 22.0 to 22.1, noticed many applets in the built-in busybox binary is messed.

For example, the _date_ applet is not following device timezone. The _start-stop-daemon_ applet is writing invalid PID into pid-file. 

Device: Xiaomi POCO F1
Android version: 9
Magisk version name: 22.1
Magisk version code: 22103

osm0sis:
Can confirm on 22103, something is weird with Magisk busybox, even vs. my SELinux compile which is also NDK Clang through @topjohnwu's box-kitchen:
```
:/ $ busybox date
Sat Apr 17 10:31:56 ADT 2021
127|:/ $ su
:/ #
:/ # date
Sat Apr 17 10:34:06 ADT 2021
:/ # busybox date
Sat Apr 17 10:35:49 ADT 2021
:/ # /data/adb/magisk/busybox date
Sat Apr 17 13:33:44 GMT 2021
```

My guess would be something with the old libc in use for Magisk compiles?

seffparker:
> What's the last version with working busybox?

busybox in Magisk 22.0 worked fine for me till the update into 22.1. I only noticed the issues with these two applets since I use `date` for logging and `start-stop-daemon` to overcome the issue #3224 

There may be other applets broken too.

osm0sis:
I've been in touch with @topjohnwu about this and he's confirmed it's almost certainly the libc in use for the busybox build, since he went back to using the r10e NDK libc starting with v22.1. The fix would be to use the current NDK libc for the busybox builds.

osm0sis:
Well yeah, mine works fine, but you'd have to manually copy it over Magisk's to get it to use it for internal functions, and I think that still wouldn't help with the Magisk app stuff.

Zarkoob:
> still wouldn't help with any Magisk app stuff which might be affected since it's busybox is built in.

So just stay on v22.0 then? Or is it ok to use v22.1+ @osm0sis ?

osm0sis:
It's not really a critical error; we haven't seen it affect anything integral to Magisk at this point.

PR #4312, opened 2021-05-05
Pure 64bit support

vvb2060:


vvb2060:
NDK r10e has many bugs, like #4225 and arm64 zlib:
image

So, I switched back to the official NDK 22.

vvb2060:
NDK r10e has many bugs, like #4225 and arm64 zlib: 
image 
 
So, I switched back to the official NDK 21 at ffe98d6b401cbbd831fe38d0750e3a461d1a2ae5, this commit is not included in this PR. Because I think topjohnwu may have other options.

Kissadas:
095bfc7e6908938305228389919c90673e6635cd Mon Sep 17 00:00:00 2001
From: vvb2060 <vvb2060@gmail.com>
Date: Fri, 7 May 2021 22:59:29 +0800
Subject: [PATCH 1/5] Fix release build

Disable resource optimization for AGP, it only has path shortening and shortening twice will crash the app.
Remove configs that are no longer needed.
---
 gradle.properties | 4 +---
 1 file changed, 1 insertion(+), 3 deletions(-)

diff --git a/gradle.properties b/gradle.properties
index f36b05a1a4b..d6a368a5ea4 100644
--- a/gradle.properties
+++ b/gradle.properties
@@ -19,12 +19,10 @@ org.gradle.parallel=true
 
 # Android
 android.

topjohnwu (approved):

PR #4374, opened 2021-05-19
Update install.md

todlove:
Magisk-v22.0.apk → Magisk-v22.0.zip


giacomoferretti:
Magisk now ships in `.apk`, not in `.zip`.

Read more here: https://github.com/topjohnwu/Magisk/releases/tag/v22.0

PR #4417, opened 2021-06-04
Rename docs/changes.md to docs/kale1992/Magisk

kale1992:
- [General] Magisk and Magisk Manager is now merged into the same package!
- [App] The term "Magisk Manager" is no longer used elsewhere. We refer it as the Magisk app.
- [App] Support hiding the Magisk app with advanced technique (stub APK loading) on Android 5.0+ (it used to be 9.0+)
- [App] Disallow re-packaging the Magisk app on devices lower than Android 5.0
- [App] Detect and warn about multiple invalid states and provide instructions on how to resolve it
- [MagiskHide] Fix a bug when stopping MagiskHide does not take effect
- [MagiskBoot] Fix bug when unpacking `lz4_lg` compressed boot images
- [MagiskInit] Support Galaxy S21 series
- [MagiskSU] Fix incorrect APEX paths that caused `libsqlite.so` fail to load


kale1992 (commented):
docs/kale1992

PR #4552, opened 2021-08-14
Add `ccache` to cache build from github action

yujincheng08:


yujincheng08:
`version.cpp` is necessary. Otherwise, ccache will miss because the compile option (`-DMAGISK_VERSION`) changed.

topjohnwu:
Included in `master`

PR #4567, opened 2021-08-23
English language string fixes

comradekingu:
More improvements to be had, but it would be better to handle translations through Weblate to get more eyeballs on the job.


app/src/main/res/values/strings.xml
```diff
@@ -24,12 +24,12 @@
     <string name="home_notice_content">Download Magisk ONLY from the official GitHub page. Files from unknown sources can be malicious!</string>
     <string name="home_support_title">Support Us</string>
     <string name="home_item_source">Source</string>
-    <string name="home_support_content">Magisk is, and always will be, free and open-source. You can however show us that you care by sending a small donation.</string>
+    <string name="home_support_content">Magisk will always be copylefted libre software.\nUse,study, change and share; with all.\nShow you care by sending a donation.</string>
```
topjohnwu, line 27: The rest of the PR is fine. May you preserve the original message here?

comradekingu: I find "Free and open-source software"/"Free and Open Source" to be confusing.
Not in the least because of the spelling, but because "free" is not qualified.
In addition (without the Oxford comma) it would mean gratis.
It doesn't thereby explain why it will remain anything.
That is also further, not explained.
"You can however" makes it seem like there is an additional point to stating something beyond "show".

What is it the user should care about, and why is it assumed the user knows without it being explained?

app/src/main/res/values/strings.xml
```diff
@@ -24,12 +24,12 @@
...
     <string name="home_installed_version">Installed</string>
     <string name="home_latest_version">Latest</string>
     <string name="invalid_update_channel">Invalid Update Channel</string>
     <string name="uninstall_magisk_title">Uninstall Magisk</string>
-    <string name="uninstall_magisk_msg">All modules will be disabled/removed!\nRoot will be removed!\nYour data potentially encrypted if not already!</string>
+    <string name="uninstall_magisk_msg">All modules will be disabled/removed!\nRoot will be removed!\nEncrypt your data if it isn\'t already!</string>
```
Fs00, line 32: I would have read the last sentence as "Your data will be potentially encrypted if it isn't already". Why the user would want to encrypt their data before removing Magisk? Am I missing something?

Fs00, line 32: I would have read the last sentence as "Your data will be potentially encrypted if it isn't already". Why the user would want to encrypt their data before removing Magisk? Am I missing something?

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
     <string name="unsupport_magisk_msg">This version of the app does not support Magisk version lower than %1$s.\n\nThe app will behave as if no Magisk is installed, please upgrade Magisk as soon as possible.</string>
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
+    <string name="unsupport_other_su_msg">Please remove the program that sent a \"su\" command and handle all requests through Magisk.</string>
```
Fs00, line 235: To be honest, the new string looks more confusing to me. Isn't this message shown when another su binary (e.g. SuperSU) is installed?
More specifically, it's the part _"the program that sent a \"su\" command"_ that confuses me. All apps that use root permissions "send" `su` commands.

Fs00 (commented):
@topjohnwu @comradekingu I've found two changes in this PR that made me wonder if I had interpreted those strings correctly when I updated the Italian translation.🤔

app/src/main/res/values/strings.xml
comradekingu: It isn't "Encrypts".

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="unsupport_other_su_msg">Please remove that just tried to handle a \"su\" command and handle all requests through Magisk.</string>
```

Fs00: You probably meant to say "...remove **the app** that just..." :)
In any case, the message worded as it is now makes sense only if the other root app is detected when an app tries to access su.
If this check instead happens as soon as Magisk is installed, it would be misleading to say that the other su app _just tried_ to handle a su command.

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
     <string name="unsupport_magisk_msg">This version of the app does not support Magisk version lower than %1$s.\n\nThe app will behave as if no Magisk is installed, please upgrade Magisk as soon as possible.</string>
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
+    <string name="unsupport_other_su_msg">Please remove that just tried to handle a \"su\" command and handle all requests through Magisk.</string>
```
comradekingu, line 235:
```suggestion
    <string name="unsupport_other_su_msg">Please remove the app that just tried to handle a \"su\" command and handle all requests through Magisk instead.</string>
```

osm0sis:
These all seem pretty unnecessary, often making as much or more sense with the originals, and hasn't addressed any of the criticism or conflicts that have arisen, so closing as abandoned.

comradekingu:
@osm0sis Do you want me to list all the errors made?

app/src/main/res/values/strings.xml
```diff
@@ -24,12 +24,12 @@
     <string name="home_notice_content">Download Magisk ONLY from the official GitHub page. Files from unknown sources can be malicious!</string>
     <string name="home_support_title">Support Us</string>
     <string name="home_item_source">Source</string>
-    <string name="home_support_content">Magisk is, and always will be, free and open-source. You can however show us that you care by sending a small donation.</string>
```
comradekingu, line 27: Misspelling of "open source", a term that produces errors in translation.

app/src/main/res/values/strings.xml
```diff
@@ -24,12 +24,12 @@
...
+    <string name="home_support_content">Magisk will always be copylefted libre software.\nUse,study, change and share; with all.\nShow you care by sending a donation.</string>
     <string name="home_installed_version">Installed</string>
     <string name="home_latest_version">Latest</string>
     <string name="invalid_update_channel">Invalid Update Channel</string>
     <string name="uninstall_magisk_title">Uninstall Magisk</string>
-    <string name="uninstall_magisk_msg">All modules will be disabled/removed!\nRoot will be removed!\nYour data potentially encrypted if not already!</string>
```
comradekingu, line 32: This is broken English.
"Your data potentially encrypted" means the data belonging to the user possibly encrypted something.

osm0sis: I disagree, as a Canadian English speaker this makes perfect sense to me.

app/src/main/res/values/strings.xml
```diff
@@ -78,7 +78,7 @@
 
     <string name="superuser_toggle_notification">Notifications</string>
     <string name="superuser_toggle_revoke">Revoke</string>
-    <string name="superuser_policy_none">No app has asked for superuser permission yet.</string>
```
comradekingu, line 81: The convention is to use plural here.

app/src/main/res/values/strings.xml
osm0sis: Fair enough, it's a dash though.

app/src/main/res/values/strings.xml
osm0sis: But it's still correct.

app/src/main/res/values/strings.xml
```diff
@@ -136,7 +136,7 @@
     <string name="settings_hide_app_title">Hide the Magisk app</string>
     <string name="settings_hide_app_summary">Install a proxy app with random package ID and custom app label</string>
     <string name="settings_restore_app_title">Restore the Magisk app</string>
-    <string name="settings_restore_app_summary">Un-hide the app and restore it back to the original APK</string>
```
comradekingu, line 139: This unreferences one from the other, so that it has to be inferred.

app/src/main/res/values/strings.xml
```diff
@@ -151,7 +151,7 @@
     <string name="settings_hosts_summary">Systemless hosts support for Adblock apps</string>
     <string name="settings_hosts_toast">Added systemless hosts module</string>
     <string name="settings_app_name_hint">New name</string>
-    <string name="settings_app_name_helper">App will be repackaged to this name</string>
```
comradekingu, line 154: Unmarked noun error.

app/src/main/res/values/strings.xml
```diff
@@ -168,10 +168,10 @@
     <string name="request_timeout">Request Timeout</string>
     <string name="superuser_notification">Superuser Notification</string>
     <string name="settings_su_reauth_title">Reauthenticate after upgrade</string>
-    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after an application upgrades</string>
```
comradekingu, line 171: Not consistent with "app/s" elsewhere.
No apps upgrade on their own, but this is suggested.

app/src/main/res/values/strings.xml
osm0sis: I disagree; it's technically correct.

app/src/main/res/values/strings.xml
```diff
@@ -168,10 +168,10 @@
...
-    <string name="settings_su_tapjack_title">Enable Tapjacking Protection</string>
```
comradekingu, line 172: This is a setting, so "enable" is implied.

app/src/main/res/values/strings.xml
```diff
@@ -168,10 +168,10 @@
...
+    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after upgrading apps</string>
+    <string name="settings_su_tapjack_title">Tapjacking Protection</string>
     <string name="settings_su_tapjack_summary">The superuser prompt dialog will not respond to input while obscured by any other window or overlay</string>
-    <string name="settings_su_biometric_title">Enable Biometric Authentication</string>
```
comradekingu, line 174: This is a setting, so "enable" is implied.

app/src/main/res/values/strings.xml
osm0sis: Also still sensible and understandable.

app/src/main/res/values/strings.xml
```diff
@@ -185,7 +185,7 @@
     <string name="settings_user_independent">User-Independent</string>
     <string name="owner_only_summary">Only owner has root access</string>
     <string name="owner_manage_summary">Only owner can manage root access and receive request prompts</string>
-    <string name="user_indepenent_summary">Each user has his/her own separate root rules</string>
```
comradekingu, line 188: This is very unelegant.

app/src/main/res/values/strings.xml
```diff
@@ -215,7 +215,7 @@
     <string name="done">Done!</string>
     <string name="failure">Failed!</string>
     <string name="hide_app_title">Hiding the Magisk app…</string>
-    <string name="open_link_failed_toast">No application found to open the link</string>
```
comradekingu, line 218: Not consistent with "app" elsewhere. That a search was made is not relevant.

app/src/main/res/values/strings.xml
osm0sis: I agree the use of "an" is Incorrect here with the plural.

app/src/main/res/values/strings.xml
osm0sis: Right, but it's fine.

app/src/main/res/values/strings.xml
osm0sis: I agree this should be "their", since otherwise it's inelegant. 😉

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
     <string name="unsupport_magisk_msg">This version of the app does not support Magisk version lower than %1$s.\n\nThe app will behave as if no Magisk is installed, please upgrade Magisk as soon as possible.</string>
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
```
comradekingu, line 235: This should have been "a \"su\" command".
How something belongs to Magisk is not explained.
"unsupported su" does not explain what it refers to.

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
...
+    <string name="unsupport_other_su_msg">Please remove the app that just tried to handle a \"su\" command and handle all requests through Magisk instead.</string>
     <string name="unsupport_external_storage_msg">Magisk is installed to external storage. Please move the app to internal storage.</string>
-    <string name="unsupport_nonroot_stub_msg">The app cannot continue to work in the hidden state as root was lost. Please restore it back to the original APK.</string>
```
comradekingu, line 237: "state" is redundant here.

app/src/main/res/values/strings.xml
osm0sis: It's also just fine the way it is.

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
...
+    <string name="unsupport_nonroot_stub_msg">The app cannot stay hidden as root was lost. Please restore it back to the original APK.</string>
     <string name="unsupport_nonroot_stub_title">@string/settings_restore_app_title</string>
     <string name="external_rw_permission_denied">Grant storage permission to enable this functionality</string>
     <string name="add_shortcut_title">Add shortcut to home screen</string>
     <string name="add_shortcut_msg">After hiding this app, its name and icon might become difficult to recognize. Do you want to add a pretty shortcut to the home screen?</string>
-    <string name="app_not_found">No application found to handle this action</string>
```
comradekingu, line 242: Not consistent with "app" used elsewhere.

app/src/main/res/values/strings.xml
osm0sis: I agree with "an" being incorrect, otherwise this makes more sense as currently written. Magisk is root so "su" can belong to it.

app/src/main/res/values/strings.xml
comradekingu: Undisputed. My problem is that it is less clear and demands more of the user to understand it.

app/src/main/res/values/strings.xml
osm0sis: It's consistent with the other use of "No application" and it's perfect English.

app/src/main/res/values/strings.xml
comradekingu: Insofar as it is possible to understand what it says, yes. It has meaning.
My problem is that it isn't localized to English.

app/src/main/res/values/strings.xml
comradekingu: @osm0sis 
Then what does that sentence mean?
Could you ask someone to explain it to you and then either give up or enlighten me, because I really can't work it out.

app/src/main/res/values/strings.xml
osm0sis: As a native English speaker it seems perfectly fine.

app/src/main/res/values/strings.xml
osm0sis: Because any force decryption flags Magisk may have added to the fstab will be reverted any device using these to run decrypted will potentially be encrypted again.

app/src/main/res/values/strings.xml
comradekingu: Nono, it is much worse than that.
If it isn't clear what the user should care about, the likelihood of receiving donations dwindle.
If the requirement is to already know, only people already in the know will have a chance to grasp what they already likely know. In my experience this is the main point to get right if you want donations.

app/src/main/res/values/strings.xml
osm0sis: Actually it's more accurate for the user by explicitly saying the original APK was changed.

app/src/main/res/values/strings.xml
comradekingu: Do you also agree that upgrading apps is not an emergent phenomenon?

app/src/main/res/values/strings.xml
osm0sis: I'll concede this would read better in the plural: "No apps have asked for superuser permissions yet."

app/src/main/res/values/strings.xml
comradekingu: @osm0sis And it should be "was detected" too. Doesn't fix much though. Instead of trying to decipher all undeniable errors, try to envision a user making use of the information conveyed.

Right now it is in the form of a riddle following a garden-path sentence.
Only at the end of it all is there anything actionable.
What is "a su"? A command, the app requesting it, the app handling it?
Why is it unsupported?
What constitutes belonging to Magisk.

app/src/main/res/values/strings.xml
comradekingu: OK.

app/src/main/res/values/strings.xml
osm0sis: I'll agree this is a bit clumsy, and your proposed change for this one is good.

app/src/main/res/values/strings.xml
osm0sis: I mean sometimes they do have an auto-updater built in, so technically not incorrect. 😉

app/src/main/res/values/strings.xml
osm0sis: I'll reopen it for you since clearly it's not abandoned after all. 😛

app/src/main/res/values/strings.xml
osm0sis: You knowing what root is should be somewhat of a prerequisite.

app/src/main/res/values/strings.xml
comradekingu: Top stuff. I am even planning to improve the language in a later PR.

app/src/main/res/values/strings.xml
comradekingu: Good point.
Also trying to imply the vastly more common scenario.

app/src/main/res/values/strings.xml
```diff
@@ -168,10 +168,10 @@
     <string name="request_timeout">Request Timeout</string>
     <string name="superuser_notification">Superuser Notification</string>
     <string name="settings_su_reauth_title">Reauthenticate after upgrade</string>
-    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after an application upgrades</string>
-    <string name="settings_su_tapjack_title">Enable Tapjacking Protection</string>
+    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after upgrading apps</string>
```
comradekingu, line 171:
```suggestion
    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after apps are upgraded</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -78,7 +78,7 @@
 
     <string name="superuser_toggle_notification">Notifications</string>
     <string name="superuser_toggle_revoke">Revoke</string>
-    <string name="superuser_policy_none">No app has asked for superuser permission yet.</string>
+    <string name="superuser_policy_none">No apps have asked for superuser permission.</string>
```
comradekingu, line 81:
```suggestion
    <string name="superuser_policy_none">No apps have asked for superuser permission yet.</string>
```

app/src/main/res/values/strings.xml
osm0sis: Feel free to workshop the donations part more, but the first sentence needs to be restored per John's wishes.

app/src/main/res/values/strings.xml
comradekingu: Ideally, yes. However, let's try to not throw concepts around and try to explain better where possible what is going on.
The outcome here is potentially disastrous.

This should be clear after reading the sentence, and we can increase chances of it happening:
* What to do
* what happened
* what it entailed
* what the resulting action/outcome should be

Regular people donate money too.

app/src/main/res/values/strings.xml
comradekingu: Unless you translate it, then all balls are in the air.
Keep it simple, and keep it short. More text and concepts equates to higher cognitive load, more stress, and less confidence as a result.
This is a win.

app/src/main/res/values/strings.xml
osm0sis: "Reauthenticate superuser permissions after application upgrades" is the easiest fix to the original string. It's often best to fix these things as simply as possible.

app/src/main/res/values/strings.xml
comradekingu: I find your application not always applicable.

app/src/main/res/values/strings.xml
osm0sis: Would need to be "permissions", no?

app/src/main/res/values/strings.xml
comradekingu: Nothing is ever fine. It is either sub-optimal or OK.

app/src/main/res/values/strings.xml
comradekingu: I agree. OK.

app/src/main/res/values/strings.xml
comradekingu: I think you can only get away with that in English.
In other languages it necessitates there being more than one permission for each app.

osm0sis: Yeah fair enough, English is quite loose in a lot of ways. 😛

app/src/main/res/values/strings.xml
comradekingu: It is one more word to get an idea, and "No application found" could also be the application looking for something.

app/src/main/res/values/strings.xml
```diff
@@ -136,7 +136,7 @@
     <string name="settings_hide_app_title">Hide the Magisk app</string>
     <string name="settings_hide_app_summary">Install a proxy app with random package ID and custom app label</string>
     <string name="settings_restore_app_title">Restore the Magisk app</string>
-    <string name="settings_restore_app_summary">Un-hide the app and restore it back to the original APK</string>
+    <string name="settings_restore_app_summary">Un-hide the app and restore it back to its original APK</string>
```
comradekingu, line 139:
```suggestion
    <string name="settings_restore_app_summary">Un-hide the app and change it back to its original APK</string>
```

app/src/main/res/values/strings.xml
comradekingu: Great point. Saying "restore" twice is not productive.

app/src/main/res/values/strings.xml
osm0sis: This is well within standard error wordings though, e.g. "No file found" "Directory not found" "Missing file"

app/src/main/res/values/strings.xml
comradekingu: I allow even the authorities that be to reconsider, even if I will accept a veto.

app/src/main/res/values/strings.xml
osm0sis: I think it's OK then. Some apps do crazy things where enabling toggles disable features, Magisk is more sensible than that but it's still good to know the feature will explicitly be enabled.

app/src/main/res/values/strings.xml
comradekingu: Aha. Semantically I understand: "Your data will potentially be encrypted, if it is not already!" to mean that.

app/src/main/res/values/strings.xml
osm0sis: Your proposed solution is still sub-optimal. And I don't think this one error message will have any effect on donations so that's neither here nor there.

app/src/main/res/values/strings.xml
osm0sis: The original "continue to work" was better. This falls under you changing too much again

app/src/main/res/values/strings.xml
osm0sis: Your proposed change is more awkward.

app/src/main/res/values/strings.xml
comradekingu:
```suggestion
    <string name="uninstall_magisk_msg">All root privileges will be removed\nAll Magisk modules will be turned of and removed!\nStorages decrypted in Magisk will be encrypted again!</string>
```

app/src/main/res/values/strings.xml
osm0sis: I agree removing the enable is probably fine. That's just me nitpicking. 😉

app/src/main/res/values/strings.xml
comradekingu: Rather than introducing inconsistent conventions in the app, I think letting crazy be crazy is better.
Having positive toggles without the fluff is what Android AOSP does.
This avoids the idea that it could be "on", but disabled.
Also there will be no translations in the imperative mood, which could result in ~do turn on Tapjacking Protection now.

app/src/main/res/values/strings.xml
```diff
@@ -215,7 +215,7 @@
     <string name="done">Done!</string>
     <string name="failure">Failed!</string>
     <string name="hide_app_title">Hiding the Magisk app…</string>
-    <string name="open_link_failed_toast">No application found to open the link</string>
+    <string name="open_link_failed_toast">No app could open the link</string>
```
comradekingu, line 218:
```suggestion
    <string name="open_link_failed_toast">No app found to open the link</string>
```

app/src/main/res/values/strings.xml
comradekingu: OK

app/src/main/res/values/strings.xml
```diff
@@ -232,13 +232,13 @@
...
     <string name="unsupport_external_storage_msg">Magisk is installed to external storage. Please move the app to internal storage.</string>
-    <string name="unsupport_nonroot_stub_msg">The app cannot continue to work in the hidden state as root was lost. Please restore it back to the original APK.</string>
+    <string name="unsupport_nonroot_stub_msg">The app cannot stay hidden as root was lost. Please restore it back to the original APK.</string>
     <string name="unsupport_nonroot_stub_title">@string/settings_restore_app_title</string>
     <string name="external_rw_permission_denied">Grant storage permission to enable this functionality</string>
     <string name="add_shortcut_title">Add shortcut to home screen</string>
     <string name="add_shortcut_msg">After hiding this app, its name and icon might become difficult to recognize. Do you want to add a pretty shortcut to the home screen?</string>
-    <string name="app_not_found">No application found to handle this action</string>
+    <string name="app_not_found">No app could carry out this action</string>
```
comradekingu, line 242:
```suggestion
    <string name="app_not_found">No app found to handle this action</string>
```

app/src/main/res/values/strings.xml
Fs00: IMHO, the only changes needed to make this sentence more understandable are:
- "disabled/removed" -> "disabled and removed"
- "Your data potentially encrypted if not already!" -> "Your data will potentially be encrypted, if it is not already!"

Anything else is unnecessary and has the potential to reduce clarity.
As it is now, it looks much more confusing than the original ("Storages decrypted in Magisk" doesn't mean anything really).

comradekingu: You are saying "potential to reduce clarity" and "more confusing", but I don't know why and there is no explanation.
It could very well be that Magisk isn't for me, but I make good use of it.

"Storages decrypted in Magisk" means any storage that was decrypted in Magisk (presumably by the user), understood as  ~since it was installed."
It could be "Storages you used Magisk to decrypt will be re-encrypted". Is that better? It assumes Magisk isn't decrypting volumes on its own.

I can't understand the original to mean changed mount options, because
"Your data will potentially be encrypted, if it is not already!" could also mean:

* All my data could be encrypted, potentially, since it isn't already. This seems like a risk of uninstalling Magisk because Magisk cannot guarantee whether encryption will occur or not.
* I have nothing to worry about, because if all my data is encrypted, there is no potential for it to be, since it _is_ already.
* I could understand data to mean the stuff I handle ~user data, like notes encrypted by note apps, as opposed to system files and the like.

app/src/main/res/values/strings.xml
comradekingu: Why is it sub-optimal, and what meaning is lost?

This is the type of string that translators will either skip, or the point at which they will stop translating alltogether.
That leads to different parts translated by different people, incomplete translations.
The lacking quality and vagueness leads to confusion over what is what, long before it enters translation.

Nobody wants to donate to something that conveys a message of not being for them.
The ~"you should have just known this" is exclusive to a limited set of users, when there is no need for it.
I among those have a harder time grasping it.

The basic error of a/an and confusion over time doesn't seem professional.
Consider that this happens when something potentially disastrous is going on.

Failing graciously and instilling confidence drives donations, the original is neither of those.

osm0sis: It's a technical error message involving 2 competing root solutions being installed simultaneously. The original is clunky but yours is not an improvement.

So, to convey the correct meaning, instead of:
```
An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.
```

Something like:
```
A \"su\" command not from Magisk has been detected. Please remove any competing root solution.
```

app/src/main/res/values/strings.xml
osm0sis: It's a better way of explaining it.

app/src/main/res/values/strings.xml
osm0sis: Consider that part vetoed then. 😛

app/src/main/res/values/strings.xml
osm0sis: "turned of" != "turned off", make it say "disabled" again there, less ambiguity.

osm0sis (changes requested):
Renamed for the eventual Squash+Merge.

Just a few more things I've suggested in the unresolved threads. 👍

app/src/main/res/values/strings.xml
```diff
@@ -25,12 +25,12 @@
...
     <string name="home_support_title">Support Us</string>
     <string name="home_item_source">Source</string>
-    <string name="home_support_content">Magisk is, and always will be, free and open-source. You can however show us that you care by sending a small donation.</string>
+    <string name="home_support_content">Magisk will always be copylefted libre software.\nUse,study, change and share; with all.\nShow you care by sending a donation.</string>
     <string name="home_installed_version">Installed</string>
     <string name="home_latest_version">Latest</string>
     <string name="invalid_update_channel">Invalid Update Channel</string>
     <string name="uninstall_magisk_title">Uninstall Magisk</string>
-    <string name="uninstall_magisk_msg">All modules will be disabled/removed!\nRoot will be removed!\nYour data potentially encrypted if not already!</string>
+    <string name="uninstall_magisk_msg">All root privileges will be removed\nAll Magisk modules will be turned of and removed!\nStorages decrypted in Magisk will be re-encrypted!</string>
```
comradekingu, line 33:
```suggestion
    <string name="uninstall_magisk_msg">All root privileges will be removed\nAll Magisk modules will be disabled and then removed!\nStorage devices decrypted through the use of Magisk will be re-encrypted!</string>
```

app/src/main/res/values/strings.xml
comradekingu: OK

app/src/main/res/values/strings.xml
```diff
@@ -137,7 +137,7 @@
     <string name="settings_hide_app_title">Hide the Magisk app</string>
     <string name="settings_hide_app_summary">Install a proxy app with random package ID and custom app label</string>
     <string name="settings_restore_app_title">Restore the Magisk app</string>
-    <string name="settings_restore_app_summary">Un-hide the app and restore it back to the original APK</string>
+    <string name="settings_restore_app_summary">Un-hide the app and change it back to its original APK</string>
```
comradekingu, line 140:
```suggestion
    <string name="settings_restore_app_summary">Un-hide the app by restoring it back to its original APK</string>
```

app/src/main/res/values/strings.xml
comradekingu: It does the opposite of explaining.
Either it stops working, _and_ it is in the hidden state
or
It stops being able to be in the hidden state.

Moreover "The app cannot continue to work in the hidden state as root" is not meaningful until "was lost" is added.

app/src/main/res/values/strings.xml
comradekingu: The rest of the PR wasn't fine, and I had given no explanation at that point.
Between changing one error, rewriting some lines, and keeping the entire message, I think consideration for what is in the message is warranted. It can wait till more pressing issues are fixed.

app/src/main/res/values/strings.xml
comradekingu: Good, but the request being made is lost. If that isn't important, how about:

>"Please uninstall the competing rooting solution installed on your device."?

Then the user doesn't have to know what \"su\" is, or a binary is.

app/src/main/res/values/strings.xml
Fs00: > Either it stops working, and it is in the hidden state
> or
> It stops being able to be in the hidden state.

Those two don't exclude each other. The app is in the hidden state, then the user loses root for some reason, and when the user enters the app receives that error message and is given the ability to "un-hide" the app (aka restoring it to the original APK).

> Moreover "The app cannot continue to work in the hidden state as root" is not meaningful until "was lost" is added.

"Since root permissions were lost, the app cannot continue to work in the hidden state anymore" sounds better?

app/src/main/res/values/strings.xml
osm0sis: Looks good!

app/src/main/res/values/strings.xml
osm0sis: Maybe just "will be removed" would be more clear. I was trying to remember if that actually happens when using that option or if they technically just stay in in /data/adb/modules

app/src/main/res/values/strings.xml
osm0sis: Then the why is lost, the su binary part is important.

app/src/main/res/values/strings.xml
osm0sis: This is a case of "you can write it any of these ways" because English. 😛

As means because here, if that is tricky in translation then let's just be clear:
```
The hidden app cannot continue to work because root was lost. Please restore it to the original APK.
```

Removed some redundant language in there too to make you happier. 😉

app/src/main/res/values/strings.xml
osm0sis: Wait, this one should probably end up matching the "Please restore it to the original APK." one we're workshopping below.

app/src/main/res/values/strings.xml
comradekingu: Sentence one I can tolerate, but it seems to suggest there is one hidden Magisk, and one non-hidden Magisk.
That is if the user understands "the hidden app" as Magisk.
If two different Magisk apps is technically true, it is confusing to bring up like that.

The second sentence is off somehow, and I think it is because both an app and root can be restored. "… to the original APK" keeps entertaining this duality.
How about:
"Magisk has lost the root access it used to stay hidden. Please reestablish it."?
or something like
""Magisk must be uninstalled because it has lost the root access it used to stay hidden. Please reestablish root in a fresh version of the app."? (if that is how it works)

app/src/main/res/values/strings.xml
comradekingu: It is an improvement for sure, but you lost me at "because the why is important"
I am thinking
"Please remove the competing rooting solution on your device by uninstalling it."
says the same thing.

app/src/main/res/values/strings.xml
comradekingu: That it is ultimately removed isn't incorrect, and shorter is better. Better to add it back in once known.

app/src/main/res/values/strings.xml
osm0sis: Those are all crazy wordy and don't seem to capture the necessary information. Have you ever run hidden Magisk after root is lost (e.g. you flash stock boot image back)? Maybe you should try so you understand the situation better.

app/src/main/res/values/strings.xml
osm0sis: It being because of "su" is not covered. I believe this is meant to cover pre-installed su in ROMs and if someone tried to run Magisk app over SuperSU su. Knowing su is what triggered the error is extremely helpful for the first case, since it might not be as obvious. Then the second sentence covers both.

app/src/main/res/values/strings.xml
comradekingu: The second option is the shortest yet.
I'd only be using my own nb_NO translation in so doing, which is as vague as the original because I wasn't 100% sure at the time.
In German and Italian it says specifically to recover the app to its original state. It is not translated in Swedish or Spanish. There is no Danish translation.

When you say "capture necessary information" that does sound bad, but I don't know why.
Here is my problem with the proposals:
>The hidden app cannot continue to work because root was lost. Please restore it to the original APK.
`The hidden app`
What is the hidden app (unanswered)
`cannot continue to work`
how?
`because root was lost.`
So by regaining root one would think "the hidden app" would continue to work.
It doesn't say whether this is general root access, or root access granted to "the hidden app"
`Please restore it to the original APK.`
"It" is unexplained here. It could either be
~root access to the original APK (presumably of "the hidden app")", or
~"the hidden app" restored into its original APK form."

vs

"Grant Magisk the root access it needed to stay hidden to hide it again, or unhide it now."

`Grant Magisk`
Direct action, we know what app it is.
`the root access`
We know the quality of the action
`it needed to stay hidden`
It was hidden, now it isn't. Root was lost. Thinking it can then be rehidden.
`to hide it again`,
Causality. This is at least a precursor to getting there. Further action may be required.
`or unhide it now.`
option b, which is the lesser of the two.
The user knows what is going on, and what the differing option entails.
All risk of the system discovering Magisk is undertaken at the risk of the user doing it, but it is not disclosed.
There is clear preference for re-rooting and keeping Magisk hidden, which has to cause fewer errors.

app/src/main/res/values/strings.xml
```diff
@@ -236,14 +236,14 @@
     <string name="unsupport_magisk_msg">This version of the app does not support Magisk version lower than %1$s.\n\nThe app will behave as if no Magisk is installed, please upgrade Magisk as soon as possible.</string>
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
+    <string name="unsupport_other_su_msg">Please remove the app that just tried to handle a \"su\" command and handle all requests through Magisk instead.</string>
```
comradekingu, line 239:
```suggestion
    <string name="unsupport_other_su_msg">Please remove the competing \"su\" rooting solution on your device by uninstalling it.</string>
```

app/src/main/res/values/strings.xml
comradekingu: Now I understand.
I think "uninstalling" alludes to it being an app moreso than "binary", seeing as that could be just a file.
If the user removes just the system file, a lot of stuff stops working and there is more support.
"On your device" makes it clear that it isn't external access of some sort.

osm0sis: Right, so my suggestion is still more complete:

```
A \"su\" binary not from Magisk has been detected. Please remove any competing root solution.
```

app/src/main/res/values/strings.xml
osm0sis: You can't grant root to it though, root was lost. Again you don't seem to understand the problem the string applies to.

app/src/main/res/values/strings.xml
osm0sis: Still needlessly wordy. "remove" it "by uninstalling it" is redundant. The one I proposed in the other thread is still more direct and succinct for a technical error message like this:

```
A \"su\" binary not from Magisk has been detected. Please remove any competing root solution and/or reinstall Magisk.
```

app/src/main/res/values/strings.xml
osm0sis: Complete uninstall does remove all modules, so let's keep it slim and go with:
```
Root will be removed!\nAll modules will be removed!\nAny storage device decrypted through the use of Magisk will be re-encrypted!
```
🙌

app/src/main/res/values/strings.xml
comradekingu: What makes it more complete (other than the new latter part)?

app/src/main/res/values/strings.xml
osm0sis: So to match my other proposal above this should be:
```
Un-hide the app and restore the original APK
```

app/src/main/res/values/strings.xml
osm0sis: It mentions the su binary was detected without being overly wordy, which I've continually told you is important. Then it tells them to remove root and/or reinstall Magisk. It's all succinct and accurate.

app/src/main/res/values/strings.xml
comradekingu: Not if the user didn't install Magisk, or forgot, or doesn't for whatever reason not make the connection.
Or in other words, the understanding a central concept is only available given the recollected experience of installing the app, and then solving for a lot of unknown factors, in the form of a riddle.
It relies heavily on technical proficiency.
Other than there not being an option to re-root, why are those two sentences the best possible solution for mitigating my concerns?

The relationship between "Magisk", "hidden" and "root" can be perfectly clear without relying on context.

If there is no other option than to unhide the app, why isn't it:
"Magisk will be un-hidden (and restored to its original state) because root was lost."
"APK" and "restore" isn't there, but is that something the user has to know about? If so, why?
Do you have concerns with it?

app/src/main/res/values/strings.xml
comradekingu: They could be removed, but still run from memory. Going with the shorter option.

app/src/main/res/values/strings.xml
osm0sis: That's ridiculous and frankly it feels like you're being intentionally obtuse. Nobody is opening the hidden Magisk app and not knowing what it is. We can't rewrite everything in the app for some imaginary idiot you envision who has no idea how anything on their device has occurred. 😕🤨

app/src/main/res/values/strings.xml
osm0sis: What the fuck are you talking about? It removes them and then reboots the device. I'm increasingly seeing you have no idea what the strings you are trying to "improve" actually mean nor what the features they're attached to do.

app/src/main/res/values/strings.xml
comradekingu: It can't possibly be succinct when there is a disconnect between the two sentences _and_ it relies on inferring the relationship.

Notwithstanding "and/or reinstall Magisk"

>_A \"su\" binary not from Magisk has been detected. Please remove any competing root solution._

has more words and is longer than 

>"Please uninstall the competing \"su\" rooting solution installed on your device."

app/src/main/res/values/strings.xml
comradekingu: I understand it as "Restores the original APK state, which un-hides the app." (in that order) which I think would mean:
"Resets the app, which will un-hide it."
How "original APK" has to come in confuses me.

app/src/main/res/values/strings.xml
comradekingu: If you uninstall (remove) a program from a Linux system, it will continue to run ephemerally till the session ends if open. Unloading modules on the fly is not something Linux kernel devs support or test for, so one could carry this over. Least the understanding could be that, however ill devised it is. By removing "disabled" the user isn't specifically in the know.
I went with your shorter option.

app/src/main/res/values/strings.xml
comradekingu: But "stay hidden" somehow is worse than "the hidden state"?

Magisk has fewer users than it deserves, because it is a great tool.
No matter how much better it gets, these types of considerations is where it either looks scary or not.

I am not the most technical user, which makes me the bulk of the current audience.
I think readability greatly enhances my enjoyment.
Think of the time it takes to decipher the meaning of something as comparable latency on the back-end.
Just because it happens at a different interfaces doesn't make it OK.

All the translations I looked at directly translate difficult passages, and that is not good.
Sure, only the users one doesn't wish for have a less than stellar command of English, but at the best of times that is you and me both.

app/src/main/res/values/strings.xml
comradekingu: Revised it with that in mind. It can't be more direct than a requested action.

Didgeridoohan: And to be clear, the original "an" before "su" wasn't wrong (entirely depending on how you pronounce "su", of course):
https://www.merriam-webster.com/words-at-play/is-it-a-or-an

app/src/main/res/values/strings.xml
osm0sis: Because the hidden apk is modified, so restoring it uses the original again. I believe I already explained this to you.

app/src/main/res/values/strings.xml
osm0sis: Yeah the "an" wasn't his, so we're good there.

app/src/main/res/values/strings.xml
osm0sis: This should still say "App will" actually, since there's very little space in that dialog.

app/src/main/res/values/strings.xml
osm0sis: Actually I changed my mind, removing the "an" here makes it unclear whether it refers to the Magisk app or apps in general. To correctly understand the latter the "an" must be put back.

app/src/main/res/values/strings.xml
osm0sis: I don't think I have the energy to relitigate this with you for the Nth time. I've explained how and why it needs to be, fix it as I said so we can move on and eventually merge this.

app/src/main/res/values/strings.xml
osm0sis: I don't think we're going to get any further here. My suggestions are about as good as they're going to get at the moment, from both a developer and user standpoint. I think you need to go with mine here so we can move forward with the rest of it.

comradekingu: That explanation I understand. I went with your version and clarified my concerns with it.

app/src/main/res/values/strings.xml
comradekingu: @Didgeridoohan Not being wrong doesn't mean OK. For as many people that happen to pronounce it differently, the chosen solution looks wrong.

app/src/main/res/values/strings.xml
comradekingu: I couldn't find it, but now I understand. It is much less of an issue here.

app/src/main/res/values/strings.xml
```diff
@@ -236,14 +236,14 @@
     <string name="unsupport_magisk_msg">This version of the app does not support Magisk version lower than %1$s.\n\nThe app will behave as if no Magisk is installed, please upgrade Magisk as soon as possible.</string>
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
+    <string name="unsupport_other_su_msg">Please uninstall the "su" rooting-solution competing with Magisk from your device.</string>
```
comradekingu, line 239:
```suggestion
    <string name="unsupport_other_su_msg">Please uninstall the "su" rooting-solution competing with Magisk from your device. Reinstalling Magisk will also do this.</string>
```

app/src/main/res/values/strings.xml
comradekingu: Sorry, I changed the suggestion a while back thinking you would see it.
Now I changed the translation. No disconnect, nothing inferred, and what reinstalling Magisk does is explained.

app/src/main/res/values/strings.xml
```diff
@@ -236,14 +236,14 @@
...
     <string name="unsupport_general_title">Abnormal State</string>
     <string name="unsupport_system_app_msg">Running this app as a system app is not supported. Please revert the app to a user app.</string>
-    <string name="unsupport_other_su_msg">An \"su\" command that does not belong to Magisk is detected. Please remove the other unsupported su.</string>
+    <string name="unsupport_other_su_msg">Please uninstall the \"su\" rooting-solution competing with Magisk from your device. Reinstalling Magisk will also do this.</string>
     <string name="unsupport_external_storage_msg">Magisk is installed to external storage. Please move the app to internal storage.</string>
-    <string name="unsupport_nonroot_stub_msg">The app cannot continue to work in the hidden state as root was lost. Please restore it back to the original APK.</string>
+    <string name="unsupport_nonroot_stub_msg">The hidden Magisk app stopped working because root was lost. Please restore the original Magisk APK.</string>
```
comradekingu, line 241: It is clear what the hidden app is
What is being restored is clear
It is clear what the original APK is of (could have been the rooting app)
Exactly as long as the suggested alternative.
`The hidden app cannot continue to work because root was lost. Please restore it to the original APK.

comradekingu:
```suggestion
    <string name="unsupport_nonroot_stub_msg">The hidden Magisk app stopped working because root was lost. Please revert it to the original Magisk APK.</string>
```

app/src/main/res/values/strings.xml
Fs00: A technical question on this string: when you say "storage devices", you mean device partitions?
Because I think it's pretty difficult to translate "storage devices" into Italian without risking to mislead the user. An external microSD is a storage device too, and after reading what @osm0sis said above _(any force encryption flags Magisk may have removed from the fstab)_, I understand that this may impact only internal device partitions.

Also, two nits now that the situation has settled a bit:
- _Root will be removed_ -> _Root permissions will be lost_ (root is not something you "remove")
- _through the use of Magisk_ -> _by Magisk_ (as it is now it seems to imply that the action of "using root" decrypts the data, which is not accurate to say)

app/src/main/res/values/strings.xml
```diff
@@ -174,10 +173,10 @@
     <string name="request_timeout">Request Timeout</string>
     <string name="superuser_notification">Superuser Notification</string>
     <string name="settings_su_reauth_title">Reauthenticate after upgrade</string>
-    <string name="settings_su_reauth_summary">Reauthenticate superuser permissions after an application upgrades</string>
-    <string name="settings_su_tapjack_title">Enable Tapjacking Protection</string>
+    <string name="settings_su_reauth_summary">Ask for superuser permissions again after upgrading apps</string>
```
comradekingu, line 176: Good?

app/src/main/res/values/strings.xml
```diff
@@ -135,9 +135,9 @@
     <string name="settings_download_path_title">Download path</string>
     <string name="settings_download_path_message">Files will be saved to %1$s</string>
     <string name="settings_hide_app_title">Hide the Magisk app</string>
-    <string name="settings_hide_app_summary">Install a proxy app with random package ID and custom app label</string>
+    <string name="settings_hide_app_summary">Install a proxy app with a random app ID and custom app label</string>
```
comradekingu, line 138: Technically "application ID"(?)

app/src/main/res/values/strings.xml
comradekingu: I moved a shorted explanation up to the "New name" and removed the helper.

app/src/main/res/values/strings.xml
osm0sis: Needlessly wordy. Mine is better.

app/src/main/res/values/strings.xml
osm0sis: This is more wordy again, restore works better. Stop changing things just to change them.

app/src/main/res/values/strings.xml
osm0sis: You're changing more than you have to again, you should be trying to fix things with as few changes to the original as possible.

app/src/main/res/values/strings.xml
osm0sis: No. package is more accurate. Stop changing things you know nothing about.

osm0sis:
148 messages in and this is completely exhausting to try and explain things to you repeatedly. Your options are that you need to go with my last suggestions so I can approve of it and this can be merged, or, I'll simply close this pull requests and take the good changes from it and make my own.

app/src/main/res/values/strings.xml
osm0sis: Removing a string is not your decision to make. You've broken the build.

app/src/main/res/values/strings.xml
osm0sis: Yeah, let's make it say internal storage to be clear.

app/src/main/res/values/strings.xml
comradekingu: I thought it was "ApplicationId" instead of what I now think is "PackageName".
Helps to be able to look things up.

app/src/main/res/values/strings.xml
comradekingu: Yes, but saying "re-authenticate" twice isn't doing more work the second time.
Shorter, consistent, and understandable. Good qualities in my book.

app/src/main/res/values/strings.xml
comradekingu: Sorry for that. Made it an empty string now.
Between broken English, and strings out of bounds in English or in translations, it seems to me to be the better option.

osm0sis (approved):

app/src/main/res/values/strings.xml
```diff
@@ -25,12 +25,12 @@
...
     <string name="home_support_title">Support Us</string>
     <string name="home_item_source">Source</string>
-    <string name="home_support_content">Magisk is, and always will be, free and open-source. You can however show us that you care by sending a small donation.</string>
+    <string name="home_support_content">Magisk is, and always will be, free, and open source. You can however show us that you care by making a donation.</string>
     <string name="home_installed_version">Installed</string>
     <string name="home_latest_version">Latest</string>
     <string name="invalid_update_channel">Invalid Update Channel</string>
     <string name="uninstall_magisk_title">Uninstall Magisk</string>
-    <string name="uninstall_magisk_msg">All modules will be disabled/removed!\nRoot will be removed!\nYour data potentially encrypted if not already!</string>
+    <string name="uninstall_magisk_msg">Root will be removed!\nAll modules will be removed!\nAny internal storage decrypted through the use of Magisk will be re-encrypted!</string>
```
vvb2060, line 33: Restore

vvb2060 (changes requested):

app/src/main/res/values/strings.xml
osm0sis: They wouldn't be disabled in that case, just dormant.

comradekingu: @vvb2060 If you can put in a full suggestion it makes it easier to decide for or against.
"All modules will be turned off" seems to cover both disabled modules and removed ones, though it sounds strange.
How far towards removal does disabling modules go, and is it the case for all of them?

vvb2060: Are you a magisk user, or just a translator? Have you checked how the strings look like when displayed on the app UI? If user has chosen to restore the boot image, then because there is no magisk, of course no module will be loaded, but the module itself is not removed.

comradekingu: @vvb2060 I checked how they looked beforehand. This looks to be a bit longer.
I understand that, but I can't and am not trying to assume your knowledge beyond that.
If it is "disabled and removed", the "disabled" part is redundant to the end result.
If it is "removed or disabled" I don't know specifically what modules that applies to, or what "disabled" really entails.
What gets left behind short of "removed"?

"All modules will be unloaded" covers all, but isn't precise.

osm0sis: Right, so it depends on the option chosen. I guess the original "disabled/removed" is more accurate since it applies to two very different scenarios.

osm0sis: Corrected. 👍

vvb2060:
Magisk does not have the ability to decrypt data, just disable forced encryption. Also, the encryption is for the DATA partition, not just the internal storage.

osm0sis:
Feel free to change it if you've got an idea for how to make it more accurate without overcomplicating it.

topjohnwu (approved):

PR #4599, opened 2021-08-29
Use isolated devpts if kernel support

yujincheng08:
Some apps detect su process by checking tty. This pr could avoid.

### DIFF
--- a/native/jni/magiskhide/hide_policy.cpp
+++ b/native/jni/magiskhide/hide_policy.cpp
@@ -35,6 +35,10 @@ void hide_unmount(int pid) {
 
     // Unmount dummy skeletons and MAGISKTMP
     targets.push_back(MAGISKTMP);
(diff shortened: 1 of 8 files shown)


native/jni/magiskhide/hide_policy.cpp
```diff
@@ -35,6 +35,10 @@ void hide_unmount(int pid) {
 
     // Unmount dummy skeletons and MAGISKTMP
     targets.push_back(MAGISKTMP);
+    auto magiskpts = MAGISKTMP + "/" SHELLPTS;
```
topjohnwu, line 38: This should already be handled by unmounting MAGISKTMP, shouldn't it?

yujincheng08: No. If `magiskpts` is not unmounted, `MAGISKTMP` cannot be mounted with `target is busy.`

topjohnwu (approved):

PR #4602, opened 2021-08-29
Update busybox

yujincheng08:
should be fixed by submodule. don't merge.


osm0sis:
I have updated the patch on my side: https://github.com/osm0sis/android-busybox-ndk/commit/091a1302a92ebfc7c973bc9e444676f43e47ebdd

Would just need ndk-box-kitchen to be refreshed and update the submodule. 👍

PR #4637, opened 2021-09-04
Use gradle to setup emulator

yujincheng08:
Windows users will be happier

### DIFF
--- a/native/build.gradle.kts
+++ b/native/build.gradle.kts
@@ -20,3 +20,16 @@ android {
         }
     }
 }
+
+afterEvaluate {
+    val adb = androidComponents.sdkComponents.adb.get().asFile.absolutePath
+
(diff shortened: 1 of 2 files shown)


osm0sis:
Hmm builds failed

yujincheng08:
hmm, gradle is hard. it runs getprop everytime...

yujincheng08:
Ok now. But there's a bug of stub codegen because mac's fs is case insensitive.

yujincheng08:
`emulator.sh` will support loading modules' sepolicies if #4642 is merged.

Otherwise we can:
```sh
for file in /data/adb/modules/*/sepolicy.rule; do
  ./magiskpolicy --live --apply $file
done
```
but it will fail if some modules requires sepolicies in `post-fs-data.sh`

PR #4642, opened 2021-09-05
Try to live patch sepolicy if necessary

yujincheng08:
Some devices has `/cache` or `/metadata` in its fstab but
does not actually mount. In such case, `magiskinit` will
fail to load sepolicy. This commit tries to live patch
modules' sepolicy as well as install the sepolicy again.

Moreover, it's useful for `emulator.sh` to load modules'
sepolicy


xxroot:
兄弟，有办法实现不需要app程序点允许授权 就可以直接运行su来获取ROOT的方法

patpolster (approved):

PR #4661, opened 2021-09-08
Update GitHub actions

topjohnwu:


vvb2060:
```
 echo 'org.gradle.caching=true' >> gradle.properties
```
enable gradle cache
```
    - name: Cache Gradle Build
      uses: actions/cache@v2
      with:
        path: |
          ~/.gradle/caches/build-cache-*
        key: gradle-builds-${{ github.sha }}
        restore-keys: |
          gradle-builds-
```

PR #4663, opened 2021-09-08
Fix ccache

yujincheng08:
1. `ccache dir` seems to need manually `mkdir`
2. `compiler_check` needs to be changed from `mtime` since ndk is generated every time.


yujincheng08:
Cache result can be seen: https://github.com/LSPosed/Magisk/actions/runs/1213651985

Aliomarn (commented):
```

```

PR #4683, opened 2021-09-16
Update strings-sq

cocka11:


AhAow353 (commented):
app/src/main/res/values-sq/strings.xml

PR #4717, opened 2021-09-23
Update Korean translation

asdf8965:


topjohnwu:
Why are there so many untranslated strings?

asdf8965:
sry

Issue #4447, opened 2021-06-22
Android 11 MIUI devices need enabled verification in boot.img AVBv2 signature


qianmo-xw:
The Xiaomi device using MT6893 cannot start. After patching the boot, it shows infinite restart. I have disabled the vbmeta test. The models are RedmiNote10Pro and RedmiK40Gaming.  I have used magiskboot extra cpio to prompt me that it failed. When I use Android_image_kitchen, the window flashes and nothing is unpacked. I think his boot has been specially encrypted. If you need it, I will upload the boot file for  the study...
Device: RedmiNote10Pro
Android version:11
Magisk version name: V23
Magisk version code: 23001

qianmo-xw:
this say my file is too big....

osm0sis:
So find another way to upload it or you're not providing anything useful and we'll have to close this issue.

qianmo-xw:
> So find another way to upload it or you're not providing anything useful and we'll have to close this issue...

This is link
https://drive.google.com/file/d/1AefUeYQcBviQ7XSFFCnWfkkDZqx8ceSy/view?usp=drivesdk

osm0sis:
> > So find another way to upload it or you're not providing anything useful and we'll have to close this issue...
> 
> This is link
> https://drive.google.com/file/d/1AefUeYQcBviQ7XSFFCnWfkkDZqx8ceSy/view?usp=drivesdk

It's asking me to request access. Can you set better permissions on it so I can just download directly?

qianmo-xw:
> > > So find another way to upload it or you're not providing anything useful and we'll have to close this issue...
> > 
> > 
> > This is link
> > https://drive.google.com/file/d/1AefUeYQcBviQ7XSFFCnWfkkDZqx8ceSy/view?usp=drivesdk
> 
> It's asking me to request access. Can you set better permissions on it so I can just download directly?

Sorry,That is I first use google drive,I don't know how to change set permissions,I think this link will be OK 
,try it,Thank you
https://mega.nz/file/U94XnYbB#iAhwO8En4LitpbePmNNCTNvylBRtJh3lZB5aD4n66yM

qianmo-xw:
> > > So find another way to upload it or you're not providing anything useful and we'll have to close this issue...
> > 
> > 
> > This is link
> > https://drive.google.com/file/d/1AefUeYQcBviQ7XSFFCnWfkkDZqx8ceSy/view?usp=drivesdk
> 
[...]

PR #4719, opened 2021-09-23
Workaround for AVBv2 bootloop on Xiaomi

carey036:
Issues #4447

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -203,7 +203,7 @@ fi
 
 ui_print "- Repacking boot image"
 ./magiskboot repack "$BOOTIMAGE" || abort "! Unable to repack boot image"
-
+./magiskboot hexpatch new-boot.img "0300000000617662746F6F6C20" "0000000000617662746F6F6C20"
 # Sign chromeos boot
 $CHROMEOS && sign_chromeos
 


canyie:
No comment?

osm0sis:
What are you even hex patching here?

carey036:
> What are you even hex patching here? If it's part of the hacked AVBv2 signature then this is not the right way to fix it and won't be accepted.

that's what i can do, hope someone could find the right way

osm0sis:
Yep, knew this rang a bell! @DarthJabba9 independently confirmed this issue to be about a week or so back, and also found whatever magiskboot is hacking in AVBv2 needs that 03 -> 00 fix to boot on his MIUI devices. @topjohnwu will need to look into what that fix actually means for his vbmeta generation code somewhere here:
https://github.com/topjohnwu/Magisk/search?q=vbmeta
https://github.com/topjohnwu/Magisk/search?q=avb

bkerler:
Confirmed to be a generic issue on Android 11 MTK. With this patch, magisk will work on Realme 8 5G EU as well. Without, it will enter bootloop.

osm0sis:
Just to document exactly where in avb header the 00 -> 03 is, for later further investigation into the avb code for @topjohnwu or others
hxd-avb2
:

Displax:
> Just to document exactly where in avb header the 00 -> 03 is, for later further investigation into the avb code for @topjohnwu or others: 

Same byte that enable verify, as original vbmeta.img`s in devices has.
Most likely, you can simply disable the verify in the vbmeta imgs instead of boot.img.

Sample vbmeta.img from google with verify flag off ( 00 ).
https://developer.android.com/topic/generic-system-image?hl=bn#flash-gsi

Reynolds251 (approved):

osm0sis:
To reiterate, this PR is not the correct fix for the problem, so will not be merged, but will stay open for now to track the issue since it's more concise than the Issue thread.

vvb2060:
https://github.com/topjohnwu/Magisk/commit/027ec7026252bf1615c19205855ae58a79f1225e#diff-bede474345c35b2e82683e17b88c351ae176ad71013d208e69473c47f212ceadR678 
https://cs.android.com/android/_/android/platform/external/avb/+/ddab37d0073e2108aa57709e9c852e3bd8e56f68:libavb/avb_vbmeta_image.h;l=60;drc=de53827b226bccef7407e4c253b0152e8d9f8e04

`00` will enable verification, do not do this for all devices.

vvb2060:
I think we need a list, do we need to enable verification for all MTK devices?

Fox2Code:
@vvb2060 I just walked though the avb code you sent, apparently there is a way to tell AVB to ignore errors without disabling it.
Maybe doing this trick could allow to fix boot loops while still having AVB enabled and allowing system partition modification.
Code I'm talking about:
https://cs.android.com/android/_/android/platform/external/avb/+/ddab37d0073e2108aa57709e9c852e3bd8e56f68:libavb/avb_cmdline.c;l=324
https://cs.android.com/android/_/android/platform/external/avb/+/ddab37d0073e2108aa57709e9c852e3bd8e56f68:libavb/avb_slot_verify.h;l=90
I did not made more research tho, but it might be a good start.

osm0sis:
This should fix it once merged, try the apk-debug.apk in the Artifacts zip: https://github.com/topjohnwu/Magisk/actions/runs/1432894612

osm0sis:
#4906

PR #4723, opened 2021-09-24
Fix terminal on android 8

vvb2060:
No need to handle untrusted_app_all_devpts on modern devices, but devpts policy is different for older devices.


topjohnwu (approved):

pojavluncher2021:
Hey vvb z I would like you to add a branch named earlylsposedswitch for testing z And create a folder named GLPoseSwitch with a json , I will make it using json genie

PR #4732, opened 2021-09-25
Add execmem to allow hook

vvb2060:
Fix zygisk on android8

### DIFF
--- a/native/jni/magiskpolicy/rules.cpp
+++ b/native/jni/magiskpolicy/rules.cpp
@@ -176,6 +176,10 @@ void sepolicy::magisk_rules() {
     // Allow Zygisk to prctl PR_SET_MM
     allow("zygote", "zygote", "capability", "sys_resource");
 
+    // Allow hook
+    allow("zygote", "zygote", "process", "execmem");
+    allow("system_server", "system_server", "process", "execmem");
+
     // Allow update_engine/addon.d-v2 to run permissive on all ROMs
     permissive("update_engine");
 


native/jni/magiskpolicy/rules.cpp
```diff
@@ -176,6 +176,10 @@ void sepolicy::magisk_rules() {
     // Allow Zygisk to prctl PR_SET_MM
     allow("zygote", "zygote", "capability", "sys_resource");
 
+    // Allow hook
+    allow("zygote", "zygote", "process", "execmem");
+    allow("system_server", "system_server", "process", "execmem");
```
aviraxp, line 181: This is not needed.

vvb2060: appdomain allowed, zygote allowed, why not also let system_server allow execmem? system_server is a common hook target.

Fox2Code: `process execmem` SELinux is used for JIT

I know this because GrapheneOS [`disables theses rules`](https://github.com/GrapheneOS/platform_system_sepolicy/commit/680e6bcc1b0d1e5438e3467b54da132261c4f0b3) because they only allow AOT and not JIT

But modules or Zygisk itself might be dependent on that if they want to load custom code.

Fox2Code (approved):
I think it allows a bit more than just hooking.
Also, I think it would be nice to be noted that some devices already include theses rules, making execmem restrictions more consistent across systems for modules.
Except for the comment, I think this commit is good to be merged.

PR #4743, opened 2021-09-28
Erase previous prop info when setting/deleting prop

canyie:
If a prop named "test" and has value "abcd", we setting its value to "a", then its value is actually "a\0cd\0", this is a special case so apps can infer the original value.
Also, deleting a prop (like resetting readonly prop) just removes the pointer to the prop info from the trie tree, apps can find the previous prop by searching the area.
In example, [Riru](https://github.com/RikkaApps/Riru) changes "ro.dalvik.vm.native.bridge" to inject into zygote processes and recover it after system server starts. Because the previous value still exists in the prop area, apps can detect Riru by simply search whether `libriruloader.so` is in the area.

**Only simple tested and passed, further testing is recommended**

### DIFF
--- a/native/jni/external/systemproperties/include/system_properties/prop_area.h
+++ b/native/jni/external/systemproperties/include/system_properties/prop_area.h
@@ -110,7 +110,7 @@ class prop_area {
 
   const prop_info* find(const char* name);
   bool add(const char* name, unsigned int namelen, const char* value, unsigned int valuelen);
-  bool rm(const char *name);
+  prop_info* rm(const char *name);
 
   bool foreach (void (*propfn)(const prop_info* pi, void* cookie), void* cookie);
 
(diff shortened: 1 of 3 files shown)


yujincheng08 (approved):

tujiacai (approved):

Fox2Code:
If true your commit do more than that, it also fix a memory leak in resetprop.
Because with what you just explained, if an app spammed reset prop the old values wouldn't be cleared from memory, wasting memory by filling it of referenced values.

yujincheng08:
clear memory != free memory

canyie:
@Fox2Code No, I just fill the old values with '\0', not "allow the memory to be reused".
Allocator in properties system is just "bump pointer allocator", so implementing "free" needs to decrease the pointer and let the memory can be reused --- It's dangerous!  If we are implementing a garbage collector in VM, all code accessing the memory is controlled, we have many methods to prevent UAF by others, such as STW, read barrier, etc. But we aren't --- who knows what will happen when others access a freed and reused memory?

topjohnwu:
Addressed in 2fef98a5afd1b4a344bca92d094e98ad6ee54006 and 7f18616cc05b872da34b63d1c1ffe44d4b4731a9, thank you!

PR #4761, opened 2021-10-06
Refine zygisk

yujincheng08:
- Skip modules when it's name starts with `riru`
- Except for modules supporting `zygisk` (with zygisk dir) to allow modules support riru and zygisk simultaneously
- Call `onLoad` as `ModuleBase *` to avoid `private` or `protected`
- Proper declare friend, otherwise `zygisk.h: error: 'impl' is a private member of 'zygisk::Api'`

### DIFF
--- a/native/jni/core/module.cpp
+++ b/native/jni/core/module.cpp
@@ -691,9 +691,9 @@ static void collect_modules() {
         if (faccessat(modfd, "disable", F_OK, 0) == 0)
             return;
         // Riru and its modules are not compatible with zygisk
-        if (zygisk_enabled && (
-                entry->d_name == "riru-core"sv ||
-                faccessat(modfd, "riru", F_OK, 0) == 0))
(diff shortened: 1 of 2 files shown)


native/jni/core/module.cpp
```diff
@@ -691,9 +691,9 @@ static void collect_modules() {
         if (faccessat(modfd, "disable", F_OK, 0) == 0)
             return;
         // Riru and its modules are not compatible with zygisk
-        if (zygisk_enabled && (
-                entry->d_name == "riru-core"sv ||
-                faccessat(modfd, "riru", F_OK, 0) == 0))
+        if (zygisk_enabled && faccessat(modfd, "zygisk", F_OK, 0) == ENOENT &&
```
canyie, line 694: `faccessat(modfd, "zygisk", F_OK, 0) == ENOENT` ?
> On success, faccessat() returns 0. On error, -1 is returned and errno is set to indicate the error.

Israelromero197 (approved):

PR #4307, opened 2021-05-03
a21s: whitelist "exynos850" to work around incorrect detection as skip_initramfs

jameskdev:
```
<12>[    2.044416]  [0           init:    1] magiskinit: Kernel cmdline iofo:
<!2>[    2.044430]  [0:           init:    1] magiskInit: skip_initramfs=[1]
<12>[    2.044443]  [0:           init:    1] }agiskanit: dorce_normalWboot=[0]
<12>[    2.044451]  [0:           init:    1] magiskinit: rootwait=[1]
<12>[    2.044460]  [0:           init:    1] magiskinit: slot=[]
<12>[    2.044471]  [0:           init:    1] magiskinit: dt_dir=[]
<12>[    2.044479]  [0:           init:    1] m!giskiNit: fstab]suffix=[]
<12>[    2.044487]  [0:           init:    1] magiskinit: hardware=[exynos850]
<12>[    2.044517]  [0:           init:    1] magiskinit: hardware.platform=[]
<12>[    2.044691]  [0:           init:    1] magiskinit: Device tree info:
<32[    2.044700Y  [0:           init:    1] magiskinit: dt_dir=[/proc/device-tree/fizmware/android]
<1">[    2.044707]  [0:           init:    1] magisiinit: fstab_suffix=[]
<12*[    2.044718]  [0:           init*    1] magisk)n)t: hardware=[exyno3850]
<12>[    2.044'26]  [0:           init:    1] magiskinit: hardware.platfOrm=[]
<12>[    2.044735]  [0:           init:    1] magiskinit: SARInit
<12>[    2.045341]  [0:           init:    1] magisciniT: Earmy mount system_root
```

As far as I know, Galaxy A21S doesn't use SARInit, and it doesn't require skip_initramfs (this device uses dynamic partition, and it has a ramdisk. Also it was launched with Android 10)
However, for some reason, the argument skip_initramfs is passed onto the kernel cmdline when it boots, causing Magisk to identify the device as skip_initramfs device.
This caused the device to go into boot loop when flashing Magisk.
Therefore, disabling skip_initramfs for this device allowed it to boot successfully and gain root access.
There may be some Samsung devices that pass on the skip_initramfs argument, yet don't really use SARInit. If bootloop occurs on some newer Samsung devices, disabling skip_initramfs may help solve the problem.

PR #4788, opened 2021-10-20
scripts: support Samsung 2SI with skip_initramfs in dtb cmdline

osm0sis:
Samsung Galaxy A21S and Galaxy M12, probably others, are hdr_v2 boot.img with 2SI judging by the ramdisk contents (init, apex, etc.), but the dtb contains an extra cmdline with skip_initramfs present, even though this shouldn't exist on 2SI and the kernel apparently doesn't even contain a skip_initramfs function

I can't find examples of other devices where skip_initramfs is present in the dtb plainly other than these so patch it out like we do for the kernel

Closes #4307

### DIFF
--- a/native/jni/magiskboot/dtb.cpp
+++ b/native/jni/magiskboot/dtb.cpp
@@ -141,8 +141,22 @@ static bool dtb_patch(const char *file) {
         fdt = static_cast<uint8_t*>(memmem(fdt, end - fdt, DTB_MAGIC, sizeof(fdt32_t)));
         if (fdt == nullptr)
             break;
+        int node;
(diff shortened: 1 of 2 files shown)


topjohnwu:
@osm0sis do you have a boot image for me to check it out?

osm0sis:
Yep!
[boot-U5.img.zip](https://github.com/topjohnwu/Magisk/files/7447962/boot-U5.img.zip)

osm0sis:
You didn't have to keep me as primary author since you rewrote the whole thing in C++, probably should have gone the other way around and had me as co-author 😛

PR #4862, opened 2021-11-01
Disable riru modules when zygisk is on

yujincheng08:
Also disable zygisk modules when zygisk is off

Reveal in the manager and let the user know why modules are not loaded

### DIFF
--- a/native/jni/core/module.cpp
+++ b/native/jni/core/module.cpp
@@ -705,8 +705,10 @@ static void collect_modules(bool open_zygisk) {
         module_info info;
         if (zygisk_enabled) {
             // Riru and its modules are not compatible with zygisk
-            if (entry->d_name == "riru-core"sv || faccessat(modfd, "riru", F_OK, 0) == 0)
+            if (entry->d_name == "riru-core"sv || faccessat(modfd, "riru", F_OK, 0) == 0) {
+                close(openat(modfd, "disable", O_CREAT | O_WRONLY, S_IRWXU));
                 return;
+            }
             if (open_zygisk) {
 #if defined(__arm__)
                 info.z32 = openat(modfd, "zygisk/armeabi-v7a.so", O_RDONLY | O_CLOEXEC);
@@ -721,6 +723,12 @@ static void collect_modules(bool open_zygisk) {
 #else
 #error Unsupported ABI
 #endif
+            }
+        } else {
+            // Disable zygisk modules when zygisk is not enabled
+            if (faccessat(modfd, "zygisk", F_OK, 0) == 0) {
+                close(openat(modfd, "disable", O_CREAT | O_WRONLY, S_IRWXU));
+                return;
             }
         }
         info.name = entry->d_name;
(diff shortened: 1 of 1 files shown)


3033071101 (approved):

Fox2Code:
If a module want to support both riru and zygisk, what happen?
zygisk on = diabled because it's a riru module
zygisk off = diabled because it's a zygisk module

yujincheng08:
Impossible

Fox2Code:
The point of my argument is not about hybrid modules but about edges cases. 
What if zygisk support is optional for a module?
I'm not against the idea of this commits, I just would like developers to have a way to not be affected by this.
Something like adding a check for `zygisk_optional` for example would be sufficient for me to be happy with this PR

yujincheng08:
This commit is just for UI shows. https://github.com/topjohnwu/Magisk/commit/73d36fdff09f69d5bc74538be0114bebede8c6f6 has already disabled riru modules.

yujincheng08:
And for edge cases, it also applies to riru modules.

native/jni/core/module.cpp
```diff
@@ -722,6 +725,13 @@ static void collect_modules(bool open_zygisk) {
 #error Unsupported ABI
 #endif
             }
+        } else {
+            // Disable zygisk modules when zygisk is not enabled
+            if (faccessat(modfd, "zygisk", F_OK, 0) == 0) {
```
Fox2Code, line 730: Something like this would make me happy,  as developers that only use zygisk to enhance their modules and not as a core functionality would not be forced to ask the user to install or not to install the zygisk files on install
```cpp
            if (faccessat(modfd, "zygisk_optional", F_OK, 0) != 0 
                && accessat(modfd, "zygisk", F_OK, 0) == 0) {
```

yujincheng08: this pr is just for UI shows so I won't make any API changes.

Fox2Code: You disable every module that contain `zygisk` files whatever they need `zygisk` or not.
You are just breaking zygisk api itself at this point. 
- I'm ok with disabling modules for UI
- I'm also ok with disabling zygisk modules that require zygisk to work
- But I'm not ok with forcing modules maker to make guesses on whatever they should add or not zygisk files to their module if it's optional for them.

You are actually forcefully disabling any module containing a `zygisk` file whatever they are needed or not without providing any proper alternative for the ones needing them.

yujincheng08: I am ok with the last point since the same logic applies to riru modules now.

3033071101 (approved):

native/jni/core/module.cpp
Fox2Code: You know what, just leave it like this, anyone can do a PR
For now it's good enough because "zygisk" is just at it's beginning stage.
I will probably wait for your PR to be accepted before doing mine on top of the changes you made.
(Or you can just apply the code I suggested, if you want)
Anyway I will probably do a PR anyway because this new behavior need also to be documented.

Fox2Code (approved):
I'm fine with the concept of disabling incompatible modules, API may need to be more modular for module developers, but this is not the goal of this pull request.

PR #4869, opened 2021-11-02
Fix slot on devices where not in /proc/cmdline

capntrips:
It appears that `androidboot.slot_suffix` is missing from `/proc/cmdline` on Pixel 6, which prevents [`mount_partitions`](https://github.com/topjohnwu/Magisk/blob/472656517f5e353198b64cf2749dce6a4bd91ad0/scripts/util_functions.sh#L264-L269) from setting `$SLOT`, which causes [`find_boot_image`](https://github.com/topjohnwu/Magisk/blob/472656517f5e353198b64cf2749dce6a4bd91ad0/scripts/util_functions.sh#L404-L411) to default to `boot_a`.

Google uses [`ro.boot.slot_suffix`](https://android.googlesource.com/device/google/gs101/+/refs/heads/master/interfaces/boot/1.2/BootControl.cpp#194) to detect the current slot in `update_engine` for Pixel 6, so I used that for the first test. I'm not sure what implications that will have on other devices. It also worked on a Pixel 5a, for whatever that's worth.

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -261,10 +261,13 @@ mount_ro_ensure() {
 
 mount_partitions() {
   # Check A/B slot
-  SLOT=`grep_cmdline androidboot.slot_suffix`
+  SLOT=`getprop ro.boot.slot_suffix`
   if [ -z $SLOT ]; then
-    SLOT=`grep_cmdline androidboot.slot`
-    [ -z $SLOT ] || SLOT=_${SLOT}
+    SLOT=`grep_cmdline androidboot.slot_suffix`
+    if [ -z $SLOT ]; then
+      SLOT=`grep_cmdline androidboot.slot`
+      [ -z $SLOT ] || SLOT=_${SLOT}
+    fi
   fi
   [ -z $SLOT ] || ui_print "- Current boot slot: $SLOT"
 


vvb2060:
/proc/bootconfig

vvb2060:
@osm0sis Perhaps `/proc/bootconfig` should be added to `grep_cmdline()` ?  79e8962

yujincheng08:
Definitely, it's because pixel 6 uses bootconfig instead of cmdline.

capntrips:
It's in there on my device:
```
oriole:/ # grep slot_suffix /proc/bootconfig
androidboot.slot_suffix = "_b"
```

capntrips:
It detects the correct slot in `mount_partitions`, but I get `! Unable to detect target image` when I try to do a Direct Install.

yujincheng08:
could you provide those two files?

capntrips:
Which two? The only file I changed from the current `HEAD` of the `master` branch is `scripts/util_functions.sh`:

```diff
diff --git a/scripts/util_functions.sh b/scripts/util_functions.sh
index ba03eaf13..dd2d6d22e 100644
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -22,7 +22,7 @@ toupper() {
 
 grep_cmdline() {
   local REGEX="s/^$1=//p"
-  cat /proc/cmdline | tr '[:space:]' '\n' | sed -n "$REGEX" 2>/dev/null
+  { tr '[:space:]' '\n' < /proc/cmdline ; sed 's/[[:space:]]*=[[:space:]]*/=/g' < /proc/bootconfig; } | sed -n "$REGEX" 2>/dev/null
 }
 
```

yujincheng08:
I mean /proc/cmdline and /proc/bootconfig...

capntrips:
Ah, that makes more sense. I need some sleep. See attached:
[proc.zip](https://github.com/topjohnwu/Magisk/files/7461272/proc.zip)

capntrips:
Maybe `||` instead of `;`?
```bash
grep_cmdline() {
  local REGEX="s/^$1=//p"
  { tr '[:space:]' '\n' < /proc/cmdline || sed 's/[[:space:]]*=[[:space:]]*/=/g' < /proc/bootconfig; } | sed -n "$REGEX" 2>/dev/null
}
```

capntrips:
I just saw your edit, which properly returns `_b`, instead of `"_b"`. I'll close this in favor of #4871.

osm0sis:
It's not merged yet 😉

osm0sis:
My bad.

PR #4880, opened 2021-11-04
Fix UB on dtb

### LINKED ISSUE #4858: OnePlus 5T - kernel_dtb segfault
--> brick my phone,after update magisk canary version to 23011,show black screen,can access twrp recovery mode.
I had to reinstall the software again, I lost all the applications ... I had to start all over again.
And now I installed the application to update 23011, but I left Magisk to the old version 23001

Device: OnePlus 5T
Android version: 9
Magisk version name: canary
Magisk version code: 23011

yujincheng08:
`operator==` of string_view will create a tmp `string_view`.
It's an UB if the `const char *` is a nullptr.
`fdt_get_name` however will return a nullptr.

Maybe fix #4858

### DIFF
--- a/native/jni/magiskboot/dtb.cpp
+++ b/native/jni/magiskboot/dtb.cpp
@@ -82,7 +82,7 @@ static void print_node(const void *fdt, int node = 0, int depth = 0) {
 }
 
 static int find_fstab(const void *fdt, int node = 0) {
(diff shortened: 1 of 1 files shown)


topjohnwu (approved):

osm0sis:
Confirmed it fixes the dtb patch segfault 🤘

Issue #4421, opened 2021-06-08
Bootloop with Xiaomi Poco M3 Pro 5G (device needs `fastboot --disable-verity --disable-verification flash vbmeta vbmeta.img`)


eberhardweber:
Device: Xiaomi Poco M3 Pro 5G (camellian)
Android version: camellian_eea_global_images_V12.0.8.0.RKSEUXM_20210519.0000.00_11.0_eea_fc5972c36a.tgz
Magisk version name: f822ca5b
Magisk version code: 23001

Patching the Magisk-patched boot.img causes a bootloop on my new Poco M3 Pro 5G / Redmi Note 10 5G (same device, camellian).

> fastboot flash boot boot.img
> Sending 'boot_b' (65536 KB)                        OKAY [  1.441s]
> Writing 'boot_b'                                   OKAY [  0.218s]
> Finished. Total time: 2.276s

Just booting the patched boot.img gives the following error:

> fastboot boot magisk_patched-23001_1VHSg.img
> Sending 'boot.img' (65536 KB)                      OKAY [  1.525s]
> Booting                                            FAILED (Status read failed (Too many links))
> fastboot: error: Command failed

Here is the Magisk install log and the boot.img used for patching:

[magisk_install_log_2021-06-08T14_55_47Z.log](https://github.com/topjohnwu/Magisk/files/6616248/magisk_install_log_2021-06-08T14_55_47Z.log)

[boot.img](https://send.vis.ee/download/461d9facd7afda57/#aeQJtUXrc-1JL-R6lo1VKQ)

Unfortunately I don't believe there exists a custom recovery for the device yet, so my hands are a bit tied, but I figured it's better to get a report in quickly.

bucefalo2:
oh.. nice to read your report.. i did post a report but was removed not sure if i did something wrong in reporting it though!
i could flash patched boot  but than it goes in bootloop and i need to boot boot.img regular again !

bucefalo2:
I've got ROOT !! solution
i've uploaded a vbmeta.img on xda ... you need that !! in order to get root i did this
i used magisk 22.1 ( probably it works on v23 as well but .. well mine was tested on that )
ptach your boot.img on magisk as you normally do 
you need to flash vbmeta.img first : 
fastboot --disable-verity --disable-verification flash vbmeta vbmeta.img  ( the one i uploaded is an empty one )
than you just fastboot flash boot boot.img
this time will work !!!!!! it worked for me 
check my post on xda : 
[...]

PR #4901, opened 2021-11-08
Revert avbflags breaking boot on many MIUI devices

### LINKED ISSUE #4447: Android 11 MIUI devices need enabled verification in boot.img AVBv2 signature

osm0sis:
- the attempt to disable vbmeta verification (set to 3) didn't help on some devices (see #4421) and actively broke many others (see #4447)
- avbtool sets it to 0 and TWRP uses this for recovery.img and it boots fine on all devices, so revert to using 0 for all

Fixes #4447
Closes #4719

### DIFF
--- a/native/jni/magiskboot/bootimg.cpp
+++ b/native/jni/magiskboot/bootimg.cpp
@@ -735,7 +735,7 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         memcpy(footer, boot.avb_footer, sizeof(AvbFooter));
         footer->original_image_size = __builtin_bswap64(off.total);
         footer->vbmeta_offset = __builtin_bswap64(off.vbmeta);
-        vbmeta->flags = __builtin_bswap32(3);
+        vbmeta->flags = __builtin_bswap32(0); // avbtool uses 0 here; 3 breaks booting on a number of devices
     }
 
     if (boot.flags[DHTB_FLAG]) {


aviraxp:
I think you can just delete this line?

osm0sis:
> I think you can just delete this line?

Yeah, but I chose to leave it so I could comment on it so the issues with 03 were documented in the code.

vvb2060:
？

aviraxp:
Removing it will prevent the side effect that patched vbmeta for disabling verification would be replaced with original one.

osm0sis:
They still do anyway, see #4421

topjohnwu:
@osm0sis the AVB flag is required for many Samsung devices

osm0sis:
> @osm0sis the AVB flag is required for many Samsung devices

Hmm, so how to reconcile that with it breaking a LOT of MIUI devices?

Fox2Code:
All Samsung kernels set `SEANDROID_FLAG`, maybe using that to set the value to 3 when needed and not touching anything otherwise. (I personally never seen a stock kernel without this flag)
It is safe to assume custom samsung kernels already have AVB disabled so we don't have to worry about them, even if they do not set the `SEANDROID_FLAG` like the stock ones.

osm0sis:
If @Fox2Code's idea is acceptable @topjohnwu I can rewrite the patch to only set 3 when SEANDROID.

Fox2Code:
@osm0sis I just mad a PR with these changes, you may want to take a look if you want. #4906

osm0sis:
Looks good to me, we'll pick this up over there then.

PR #4915, opened 2021-11-09
fix ZIMAGE ftruncate bad sys call on arm, off_t is 64 bit

### LINKED ISSUE #3703: Magisk cannot handle g7 play, "Bad system call" on repack
``` 
g7 play
  android 10 (lunched 9)
  A/B partitions
  no recovery partition
  bootloader uses skip_initramfs
Boot: kernel mounts partition system_x as / (ramdisk is *not* used)
Recovery boot: kernel mounts ramdisk as /
-----------------------------------------------
Install Magisk (last canary 04c53c35):
  Ramdisk:Yes  A/B:Yes  SAR:Yes
Boot: kernel mounts partition system_x as / (no magisk)
Recovery boot: 'magisk init' calls RecoveryInit (enter in recovery)
-----------------------------------------------
One way for magisk work *only* in this case is:
patch getinfo.cpp:
   if (1 /*key == "RECOVERYMODE" && value == "true"*/) {
and init.cpp will call SARInit or RecoveryInit
```
 
[proc-mounts.txt](https://github.com/topjohnwu/Magisk/files/5822001/proc-mounts.txt)

[boot.img.gz](https://github.com/topjohnwu/Magisk/files/5822020/boot.img.gz)

chaosmaster:
fixes #3703

### DIFF
--- a/native/jni/magiskboot/bootimg.cpp
+++ b/native/jni/magiskboot/bootimg.cpp
@@ -554,7 +554,7 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         if (boot.flags[ZIMAGE_KERNEL] &&
             boot.k_fmt == GZIP && hdr->kernel_size() > boot.hdr->kernel_size()) {
             // Revert and try zipfoli
-            ftruncate(fd, lseek(fd, -hdr->kernel_size(), SEEK_CUR));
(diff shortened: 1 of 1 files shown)


topjohnwu:
Thank you for the PR, I have created one that covers the issue

PR #4927, opened 2021-11-12
Disable modules section if Magisk is not active

### LINKED ISSUE #4925: Feature Request: Hide/Disable the modules section if device isn't rooted.
In 23013, if the user isn't rooted, the Modules section is still visible, and opening it yields a blank screen. It will be more suitable to hide the section altogether, kinda like how the Superuser section is handled.

Also, if the user IS rooted, but has no modules installed, it'd be feasible to include a message such as "No modules installed" to improve the UX a tad bit.

Device: any
Android version: any
Magisk version name: 23013
Magisk version code: 23013

RikkaW:
Fix #4925


rahulary:
Gpay not working in Alpha magisk

topjohnwu (approved):

rahulary:
It finally worked safety net passed too🤩.thanks😍

PR #4929, opened 2021-11-12
Use std::list instead of std::vector

XeonDead:
Since we're not sure about what would be the amount of zygisk modules installed, let's not allocate a long memory array and just run a std::list-type container to conserve memory.

Signed-off-by: XeonDead <xeondead@gmail.com>


dev-harsh1998 (approved):

yujincheng08:
nonsense.

Using a list takes extra `n` pointers and its cache rate is low. A potential improvement is `unique_ptr<T[]>`, but why not vector which can do the same thing but with grammar sugar `operator []`?

PR #4951, opened 2021-11-17
Disable Zygisk option on pre 23010

canyie:


osm0sis:
Not sure this is a good idea, since it allows for it to be enabled before the reboot during the upgrade.

PR #4906, opened 2021-11-08
Only fully disable vbmeta on seandroid/mtk kernels

Fox2Code:
See #4901 

Many MIUI devices require 0 to boot properly
But Samsung/MTK may need 3 to boot properly

This is still POC as we have not many information about the issue, 
I theorized that MIUI on Android11 crashed due to the lack of avb flags due to the `0x2` flag being set.
The `0x1` disable most checks already, and should be enough for most devices.

Note: please restore stock image before trying this build, the fix target MIUI Android 11 users, but any other device owner are welcome to test the changes included in this PR to check for any regressions.

Fixes #4447

PR #4964, opened 2021-11-22
fix bootloop on redmi and xiaomi when vbmeta verity is disabled

carey036:
#4447 #4906 #4901 

add a parameter to decide to set avb flag to 0 or 3 in magiskboot repack.
and we can get the device brand or device model by boot_patch.sh.

### DIFF
--- a/native/jni/magiskboot/bootimg.cpp
+++ b/native/jni/magiskboot/bootimg.cpp
@@ -503,7 +503,7 @@ int unpack(const char *image, bool skip_decomp, bool hdr) {
 #define file_align() \
 write_zero(fd, align_off(lseek(fd, 0, SEEK_CUR) - off.header, boot.hdr->page_size()))
 
-void repack(const char *src_img, const char *out_img, bool skip_comp) {
+void repack(const char *src_img, const char *out_img, bool skip_comp, bool keep_vbmeta_verity) {
     const boot_img boot(src_img);
     fprintf(stderr, "Repack to boot image: [%s]\n", out_img);
 
@@ -761,7 +761,9 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         memcpy(footer, boot.avb_footer, sizeof(AvbFooter));
         footer->original_image_size = __builtin_bswap64(off.total);
         footer->vbmeta_offset = __builtin_bswap64(off.vbmeta);
-        vbmeta->flags = __builtin_bswap32(3);
+        if(!keep_vbmeta_verity) {
+            vbmeta->flags = __builtin_bswap32(3);
+        }
     }
 
     if (boot.flags[DHTB_FLAG]) {
--- a/native/jni/magiskboot/main.cpp
+++ b/native/jni/magiskboot/main.cpp
@@ -34,6 +34,8 @@ Supported actions:
     to [outbootimg], or new-boot.img if not specified.
     If '-n' is provided, it will not attempt to recompress ramdisk.cpio,
     otherwise it will compress ramdisk.cpio and kernel with the same method
+    If '-k' is provided, it will not attempt to disable dm_verity,
+    otherwise it will disable dm_verity
     in <origbootimg> if the file provided is not already compressed.
 
   hexpatch <file> <hexpattern1> <hexpattern2>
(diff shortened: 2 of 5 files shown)


Mkjng:
Please give me nokia6 software

native/jni/magiskboot/bootimg.cpp
```diff
@@ -761,7 +761,11 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         memcpy(footer, boot.avb_footer, sizeof(AvbFooter));
         footer->original_image_size = __builtin_bswap64(off.total);
         footer->vbmeta_offset = __builtin_bswap64(off.vbmeta);
-        vbmeta->flags = __builtin_bswap32(3);
+        if(keep_dm_verity) {
+            vbmeta->flags = __builtin_bswap32(0);
+        } else {
+            vbmeta->flags = __builtin_bswap32(3);
+        }
```
Fox2Code, line 768: I would replace:
```c++
        if(keep_dm_verity) {
            vbmeta->flags = __builtin_bswap32(0);
        } else {
            vbmeta->flags = __builtin_bswap32(3);
        }
```
By
```c++
        if(!keep_dm_verity) {
            vbmeta->flags = __builtin_bswap32(3);
        }
```
to avoid conflicting with custom ROMs that may require 3 as vbmeta flag.
(All stock ROMs are supposed to have a 0 `vbmeta->flags` by default)

scripts/boot_patch.sh
```diff
@@ -208,8 +208,11 @@ fi
 #################
 
 ui_print "- Repacking boot image"
-./magiskboot repack "$BOOTIMAGE" || abort "! Unable to repack boot image"
-
+if [ `getprop ro.product.brand` == "Xiaomi" -o `getprop ro.product.brand` == "Redmi" ]; then
```
Fox2Code, line 211: Replace this:
```bash
if [ `getprop ro.product.brand` == "Xiaomi" -o `getprop ro.product.brand` == "Redmi" ]; then
```
by this:
```bash
if [ "$(grep_get_prop ro.product.brand)" == "Xiaomi" ] || [ "$(grep_get_prop ro.product.brand)" == "Redmi" ]; then
```
As some Magisk modules may change `ro.product.brand`, this will be more reliable.

native/jni/magiskboot/main.cpp
```diff
@@ -166,13 +168,25 @@ int main(int argc, char *argv[]) {
         }
         return unpack(argv[idx], nodecomp, hdr);
     } else if (argc > 2 && action == "repack") {
-        if (argv[2] == "-n"sv) {
-            if (argc == 3)
+        int idx = 2;
```
yujincheng08, line 171: Please use `getopt_long`

scripts/boot_patch.sh
yujincheng08: Instead of hard coding it, we should make it a choice in the app like dm verify:
https://github.com/topjohnwu/Magisk/blob/f42a87b51ae64cf5ed4937e3cd17c7f865f33bc1/app/src/main/res/layout/fragment_install_md2.xml#L93-L100

native/jni/magiskboot/bootimg.cpp
```diff
@@ -761,7 +761,11 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         memcpy(footer, boot.avb_footer, sizeof(AvbFooter));
         footer->original_image_size = __builtin_bswap64(off.total);
         footer->vbmeta_offset = __builtin_bswap64(off.vbmeta);
-        vbmeta->flags = __builtin_bswap32(3);
+        if(keep_dm_verity) {
```
yujincheng08, line 764: format your code

yujincheng08 (changes requested):

osm0sis:
Should also only be Xiaomi/Redmi on Android 11+, IIRC.

native/jni/magiskboot/bootimg.cpp
carey036: make it a choice in the app like dm verify is difficult for me.

the choice interface only shows when users don't have root. I am not sure if when those who have root chose 'select and patch a file', how magisk get the parameter to decide keep vbmeta verity open or not.

Mkjng:
Nokia 6 root

osm0sis:
We still need the toggle added to the app for the env variable @topjohnwu introduced. 🤔

PR #4981, opened 2021-11-25
Fix NPE when apk could not be parsed

vvb2060:
#4983 When info is empty, the apk file we downloaded is actually the web page.


topjohnwu (approved):

luckyman1352 (approved):

3033071101:
08615805052204

Issue #4955, opened 2021-11-18
Enabling installation to inactive slot on Pixel devices


capntrips:
Device: Pixel 5a and Pixel 6
Android version: 12
Magisk version name: Canary
Magisk version code: 23014

I built a new version of `bootctl` against the Pixel 5a source:

```bash
repo init -u https://android.googlesource.com/platform/manifest -b android-12.0.0_r10
repo sync -c -j8
. build/envsetup.sh
lunch aosp_barbet-userdebug
mmma system/extras/bootctl
```
With a small change to statically link the missing libraries:

```diff
diff --git a/bootctl/Android.bp b/bootctl/Android.bp
index f63871cf..7c60f65b 100644
--- a/bootctl/Android.bp
+++ b/bootctl/Android.bp
@@ -26,10 +26,14 @@ cc_binary {
         "-Werror",
     ],
 
-    shared_libs: [
+    static_libs: [
         "android.hardware.boot@1.0",
         "android.hardware.boot@1.1",
         "android.hardware.boot@1.2",
+    ],
+    shared_libs: [
+        "liblog",
+        "libcutils",
         "libhidlbase",
         "libutils",
     ],
```

I included it in a build of Magisk with installation to the inactive slot enabled for Pixel 5a and was able to take an OTA using the method for A/B devices from the [guide](https://topjohnwu.github.io/Magisk/ota.html#devices-with-ab-partitions).

An XDA user reported success with a similar build on a Pixel 6 (`android-12.0.0_r12`, `aosp_oriole-userdebug`).

Pixel 6 uses the `devinfo` partition for boot control, rather than `gpt`. See [here](https://android.googlesource.com/device/google/gs101/+/refs/tags/android-12.0.0_r12/interfaces/boot/1.2/BootControl.cpp#208) and [here](https://android.googlesource.com/device/google/gs101/+/refs/tags/android-12.0.0_r12/interfaces/boot/1.2/BootControl.cpp#228).

I suspect the newer build of `bootctl` will work for most devices before Pixel 6, though it is easy enough to [check](https://android.googlesource.com/device/google/gs101/+/refs/tags/android-12.0.0_r12/interfaces/boot/1.2/BootControl.cpp#147) for a valid `devinfo` partition.

I'm happy to submit PRs adding support, but I figured I had better get some feedback before I do the work.

vvb2060:
[...]

PR #5006, opened 2021-12-02
Patch and restore `vbmeta` partition on Magisk install and uninstall

capntrips:
Note: All mentions of `vbmeta` below are in reference to the dedicated partition, not the segment of the `boot` partition.

- This is currently just a POC. I did just enough to make it work in ideal conditions on at least a Pixel 6. If nothing else, some checks are needed to detect if patching `vbmeta` is even necessary.
- `find_vbmeta_image` isn't needed in `uninstaller.sh` or `restore_imgs`, as `vbmeta` would be restored from backup, if it exists.
- This does nothing in `processTar`.
- I added `find_vbmeta_image` to `addon.d.sh`, as it is needed by `install_magisk`, for now, but I have no idea if it makes sense to leave it there once some checks are added.
- This uses the default block size (4096) to determine the padding size when backing up the stock `vbmeta` partition. This is not strictly necessary, since Magisk will write zeroes when flashing it on restore, but it is nice if the SHA1 matches the one from the factory image.
- Should this use the `KEEPVBMETAFLAG` env variable?
- Would it be better to add AVB 2.0 signing support? Would that even eliminate the need to patch `vbmeta`?

Related to #4955, but this does nothing to address the need for an updated `bootctl` binary

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -47,7 +47,7 @@ abstract class MagiskInstallImpl protected constructor(
 
     protected var installDir = File("xxx")
     private lateinit var srcBoot: File
-    private lateinit var srcVbmeta: File
+    private var srcVbmeta: File? = null
 
     private val shell = Shell.getShell()
     private val service get() = ServiceLocator.networkService
@@ -65,10 +65,10 @@ abstract class MagiskInstallImpl protected constructor(
         val vbmetaPath = "find_vbmeta_image; echo \"\$VBMETAIMAGE\"".fsh()
         if (vbmetaPath.isEmpty()) {
             console.add("! Unable to detect target vbmeta image")
-            return false
+        } else {
+            srcVbmeta = SuFile(vbmetaPath)
+            console.add("- Target vbmeta image: $vbmetaPath")
         }
-        srcVbmeta = SuFile(vbmetaPath)
-        console.add("- Target vbmeta image: $vbmetaPath")
         return true
     }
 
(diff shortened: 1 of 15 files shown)


capntrips:
I assume the `vbmeta` partition still needs `flags` set to 3, in that case?

vvb2060:
@capntrips Yes, `vbmeta` partition must be top-level

capntrips:
This still does nothing in `processTar`, if that is relevant.

capntrips:
With `vbmeta` `flags` to 0 in the `boot` image and 3 in the `vbmeta` image, my Pixel 5a bootloops once and goes to the factory reset screen in recovery.

vvb2060:
The flag of vbmeta partition from 0 to 3 or from 3 to 0 need to wipe data partition, because data partition encryption key has changed.

vvb2060:
I think only some of the devices need to modify the flag , pixel is not one of them.

capntrips:
So is there any reason to set the `KEEPVBMETAFLAG` with this PR?

P.S. I did not have to wipe when changing from 3 to 0.

capntrips:
I was able to flash the patched `boot` image with the `vbmeta` `flags` set to 0 by flashing a self signed `vbmeta` image without wiping data.

1. Extract img files from factory image
2. Replace the stock `boot` image with the patched one
3. Run `avbtool info_image --image vbmeta.img` to get the stock values
4. Generate or download a private key (I used the test key from [here](https://android.googlesource.com/platform/external/avb/+/refs/tags/android-12.0.0_r12/test/data/testkey_rsa4096.pem))
5. Run the below `avbtool make_vbmeta_image` command to generate the new `vbmeta` image
6. `fastboot flash` the patched `boot` image and the newly generated `vbmeta` image

 ```
avbtool make_vbmeta_image --output vbmeta.img --padding_size 4096 --algorithm SHA256_RSA4096 --key testkey_rsa4096.pem --rollback_index 1636070400 --rollback_index_location 0 --flags 0 --include_descriptors_from_image dtbo.img --include_descriptors_from_image vendor_boot.img --include_descriptors_from_image vendor.img --chain_partition vbmeta_system:1:testkey_rsa4096.pem --chain_partition boot:2:testkey_rsa4096.pem
```

**Note:** Using a patched `boot` image with the `vbmeta` `flags` set to 3 cause it to boot directly into `fastboot`. Adding `--flags 3` to  `avbtool make_vbmeta_image` bootlooped once then to the factory reset screen in recovery.

capntrips:
I was able to use the stock `vbmeta` by patching `boot` image with a new key, digest, and signature. If we can get RSA signing to work in `magiskboot`, it looks like there is no need to touch the `vbmeta` partition.

capntrips:
Closing in favor of #5018. If for some reason that doesn't work on other devices, feel free to reopen.

PR #5015, opened 2021-12-05
Only care about actual mount namespace isolating

canyie:


Fox2Code (approved):
Yeah I think it's likely some random OEM might call unshare multiple times before calling the mount namespace isolation unshare, this wouldn't be the first time custom phone vendors do weird changes to the android operating system.

topjohnwu (approved):

PR #5030, opened 2021-12-07
magisk: add command to disable all modules and reboot

meiskam:
adds `--disable-modules` to magisk utility, which calls `disable_modules()` in the daemon and reboots


vvb2060 (changes requested):
https://github.com/topjohnwu/Magisk/blob/master/docs/faq.md#q-i-installed-a-module-and-it-bootlooped-my-device-help

PR #5034, opened 2021-12-07
Further fix `oplus.fstab` support

### LINKED ISSUE #5016: With my colorOS 12,such as OPPO Find X2 Pro and ONEPLUS 8T,magisk 23 work well,but the beta,the alpha,and the canary can not  boot to the desktop but to the fastboot,how can i do
Device: OPPO Find X2 Pro
Android version:12
Magisk version name: alpha
Magisk version code: 23015

yujincheng08:
In some oneplus devices, `oplus.fstab` does exists but `init` never
loaded it and those entries in `oplus.fstab` are written directly to
`fstab.qcom`. Previous implementation will introduce duplicate entries
to `fstab.qcom` and brick the device. This commit filters those entries
from `oplus.fstab` that are already in `fstab.qcom` and further filters
duplicated entries in `oplus.fstab` (keep only the last entry).

Fix #5016

### DIFF
--- a/native/jni/init/init.hpp
+++ b/native/jni/init/init.hpp
@@ -26,6 +26,8 @@ struct fstab_entry {
     fstab_entry() = default;
     fstab_entry(const fstab_entry &) = delete;
     fstab_entry(fstab_entry &&) = default;
+    fstab_entry &operator=(const fstab_entry&) = delete;
+    fstab_entry &operator=(fstab_entry&&) = default;
     void to_file(FILE *fp);
 };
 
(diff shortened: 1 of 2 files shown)


158212892:
> In some oneplus devices, `oplus.fstab` does exists but `init` never loaded it and those entries in `oplus.fstab` are written directly to `fstab.qcom`. Previous implementation will introduce duplicate entries to `fstab.qcom` and brick the device. This commit filters those entries from `oplus.fstab` that are already in `fstab.qcom` and further filters duplicated entries in `oplus.fstab` (keep only the last entry).
> 
> Fix #5016

That's OK，thx😘😘

Marukon (approved):

topjohnwu (approved):

PR #5070, opened 2021-12-15
Remove check of `recovery_dtbo`

### LINKED ISSUE #4976: Can't boot to recovery, ramdisk says "no" after flashing magisk
After flashing the magisk, in magisk manager, ramdisk shows "no". It is "yes" before flashing magisk. Oneplus 7 pro has recovery inside of its boot partition and after flashing magisk, recovery becomes inaccessible. When trying to boot to recovery, it would just boot back to system. 

Furthermore since ramdisk shows no, flashing to opposite slot also fails. 

Before flashing magisk
photo_2021-11-24_16-17-03

After flashing magisk
photo_2021-11-24_16-08-03

Since i am unsure on what all logs you would need I am attaching a boot log and a dmesg. Do let me know if you need anything else.
[boot.log](https://github.com/topjohnwu/Magisk/files/7594975/boot.log)
[dmesg.log](https://github.com/topjohnwu/Magisk/files/7595011/dmesg.log)

Note: This issue didn't exists on android 11 or below. 

Device: Oneplus 7 ( pro ) - guacamole(b)
Android version: 12.0
Magisk version name: 5d6d2417
Magisk version code: 23015

yujincheng08:
Fix #4976

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -101,8 +101,6 @@ case $? in
     ;;
 esac
 
-[ -f recovery_dtbo ] && RECOVERYMODE=true
-
 ###################
 # Ramdisk Restores
 ###################
@@ -149,11 +147,11 @@ fi
 
 ui_print "- Patching ramdisk"
 
-echo "KEEPVERITY=$KEEPVERITY" > config
-echo "KEEPFORCEENCRYPT=$KEEPFORCEENCRYPT" >> config
-echo "PATCHVBMETAFLAG=$PATCHVBMETAFLAG" >> config
-echo "RECOVERYMODE=$RECOVERYMODE" >> config
-[ ! -z $SHA1 ] && echo "SHA1=$SHA1" >> config
+ui_print "- config: $(echo "KEEPVERITY=$KEEPVERITY" | tee config)"
+ui_print "- config: $(echo "KEEPFORCEENCRYPT=$KEEPFORCEENCRYPT" | tee -a config)"
+ui_print "- config: $(echo "PATCHVBMETAFLAG=$PATCHVBMETAFLAG" | tee -a config)"
+ui_print "- config: $(echo "RECOVERYMODE=$RECOVERYMODE" | tee -a config)"
+[ ! -z $SHA1 ] && ui_print "- config: $(echo "SHA1=$SHA1" | tee -a config)"
 
 # Compress to save precious ramdisk space
 SKIP32="#"


osm0sis:
RECOVERYMODE should ideally be a user choice in the app again for all, no?

yujincheng08:
> RECOVERYMODE should ideally be a user choice in the app again for all, no?

nah

ipdev99:
`[ -n "$SHA1" ] && ui_print "- config: $(echo "SHA1=$SHA1" | tee -a config)"`

This was pointed out to me recently. :wink:

>-n is unreliable on Android and ! -z should always be used instead.
Ran into similar issues in the past which I fixed here: topjohnwu/Magisk@93e1266

Not sure if it will help but, it might make it more compatible.

Cheers. :cowboy_hat_face:

yujincheng08:
-n needs double quotes:
https://github.com/topjohnwu/Magisk/commit/93e1266#diff-f8cb6ae933f02a7a542c3f84299aaa23d8241aca56580fd7dcb64a92bce47692L40-L42

osm0sis:
Magisk needs ! -z, I had to switch them all from -n back to ! -z at one point because -n broke the logic.

yujincheng08:
> Magisk needs ! -z because the env can differ, I had to switch them all from -n back to ! -z at one point because -n broke the logic.

I suppose that is because you didn't use `-n` correctly: `-n` needs double quotes to work correctly.
So: `[ -n "$ENV" ]` is correct and `[ -n $ENV ]` is incorrect.

osm0sis:
Yeah, wasn't me using it. It was John, I just found the regression and fixed it. 😛

yujincheng08:
See this: https://unix.stackexchange.com/questions/109625/shell-scripting-z-and-n-options-with-if/109631#comment1152278_109631

osm0sis:
Fair enough! Good to know why it didn't work.

My next question would be why use `[ -n "$var" ]` over `[ ! -z $var ]` then, since they both use the same number of chars and shouldn't we try to remain consistent?

yujincheng08:
Shell lint complains about `[ ! -z ]`

osm0sis:
I mean, shell lint can't love Android/mksh's use of == either... 🤷‍♂️

So should we switch them all to -n with quotes? My feel is more that we know it works with ! -z so who cares how lint feels.

ipdev99:
Sorry to spark this recent conversation.

Would it not be better to just use brackets then?
`[[ -n $SHA1 ]] && ui_print "- config: $(echo "SHA1=$SHA1" | tee -a config)"`

osm0sis:
Nah, don't be sorry, this is good info right here; I'd been pretty confused about why -n didn't work before, just knew that it clearly broke some devices, and didn't come across that stackexchange at the time. 🙂

topjohnwu:
@ipdev99 double brackets are `bash` only, so not available on ash/mksh

PR #5094, opened 2021-12-21
Update Install documentation

conorgil:
Wording and formatting changes focused on clarity and readability.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -20,13 +20,13 @@ Download the [latest Magisk APK](https://github.com/topjohnwu/Magisk/releases) f
 
 Follow the steps below:
 
-1. **Does the Magisk app report that your device has ramdisk in the boot partition (`Ramdisk = Yes`)?**
-   - **Yes.** Continue to next question.
-   - **No.** Read the [Magisk in Recovery](#magisk-in-recovery) section before installing. The information in that section is VERY important!
-2. **Are you using a Huawei device?**
+1. **Are you using a Huawei device?**
    - **Yes.** Huawei devices are not supported. Sorry.
    - **No.** Continue to next question.
-4. **Are you using a Samsung device _and_ the Magisk app reports `SAR = Yes`?**
+2. **Does the Magisk app report that your device has ramdisk in the boot partition (`Ramdisk = Yes`)?**
+   - **Yes.** Continue to next question.
+   - **No.** Read the [Magisk in Recovery](#magisk-in-recovery) section before installing. The information in that section is VERY important!
+3. **Are you using a Samsung device _and_ the Magisk app reports `SAR = Yes`?**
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -12,7 +12,7 @@ Before you start:
 
 ---
 
-Download and install the latest Magisk app. We use the app to gather some information about your device. In the home screen, you should see this:
+Download the [latest Magisk APK](https://github.com/topjohnwu/Magisk/releases) and install it on your device using `adb`. We use the app to gather some information about your device. In the home screen, you should see this:
```
usrDottik: `adb` isn't really needed to install the apk, you can use the device's installer

conorgil: Fair point. Updated.

usrDottik: > Fair point. Updated.

Nice :D

bbronzebaton:
Complicated

docs/install.md
```diff
@@ -1,6 +1,8 @@
 # Installation
 
-If you already have Magisk installed, it is **strongly recommended** to upgrade directly via the Magisk app using its "Direct Install" method. The following tutorial is only for the initial installation.
+This tutorial is only for the initial installation of Magisk.
+
+If you already have Magisk installed, it is **strongly recommended** that you upgrade directly via the Magisk app using its "Direct Install" method.
 
 ## Getting Started
```
Fizie92, line 7: Start

Fizie92 (changes requested):

ipdev99:
To add a suggestion.

I have never been a big fan of just blindly flashing the patched boot image.
If it does not work, you will have issues.
_Forever boot, bootloop (soft or full), reboot into Rescue Party or bootloader._

You have to restore the original boot image or even more steps to recover.
_If you can recover. Some cheap devices might be hard-bricked._


It is much easier to just temp boot using `fastboot boot` to test the magisk patched image.
If it fails to boot, the most you have to do is manually reboot the device.
_I know there is device limitations, since some do not support the `fastboot boot` command._

For me steps 6 and 7 are..

6. **Test the Magisk patched image** By temporarily booting it.
   - `$> fastboot boot /path/to/magisk_patched.img`

7. **Open the Magisk app to check if Magisk is active.** If active then install using the Direct Install option.
   - In the Magisk app, press the blue `Install` button in the top right corner of the Magisk card.
   - Choose the `Direct Install (Recommended)` option.
   - Setup Magisk.
   - Install modules.
   - Reboot.

I am not sure the best way to add the test boot option into the instructions.
I do not know if these steps work for devices that require patching the recovery image.
_I do not have a device that requires using the recovery image so I can not test it._

docs/install.md
conorgil: Header changed to `Start`

androidacybot (commented):
This really doesn't make sense. Getting started is more appropriate and standard.

PR #5099, opened 2021-12-23
Fix: window on Android 11. setSystemUIVisibility was deprecated in API  level 29

nocsg:
Fix window on Android 11. `setSystemUIVisibility` was `deprecated` in API  level 29, `setSystemUIVisibility` can't adapt accurate full screen on Android 11 when The device is Hollowed screen.

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
@@ -6,6 +6,9 @@ import android.os.Build
 import android.os.Bundle
 import android.view.KeyEvent
 import android.view.View
+import android.view.WindowInsets
+import android.view.WindowInsetsController
+import androidx.annotation.RequiresApi
 import androidx.appcompat.app.AppCompatDelegate
 import androidx.core.content.res.use
 import androidx.databinding.DataBindingUtil
@@ -41,6 +44,7 @@ abstract class BaseUIActivity<VM : BaseViewModel, Binding : ViewDataBinding> :
         AppCompatDelegate.setDefaultNightMode(theme)
     }
 
+    @RequiresApi(Build.VERSION_CODES.R)
     override fun onCreate(savedInstanceState: Bundle?) {
         layoutInflater.factory2 = LayoutInflaterFactory(delegate)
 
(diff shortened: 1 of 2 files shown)


nocsg:
I have already build on windows-latest, but I don't how to use the github pr workflow

vvb2060:
```kotlin
    @RequiresApi(Build.VERSION_CODES.R)
    override fun onCreate(savedInstanceState: Bundle?)
```

app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
```diff
@@ -54,17 +57,15 @@ abstract class BaseUIActivity<VM : BaseViewModel, Binding : ViewDataBinding> :
             .use { it.getDrawable(0) }
             .also { window.setBackgroundDrawable(it) }
 
-        window?.decorView?.let {
-            it.systemUiVisibility = (it.systemUiVisibility
-                    or View.SYSTEM_UI_FLAG_LAYOUT_STABLE
-                    or View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN
-                    or View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION)
-        }
+        window.decorView.windowInsetsController?.hide(WindowInsets.Type.statusBars())
```
RikkaW, line 60: Using R APIs for all API levels ??? This will crash on any devices lower than R.

If you do want to change this, you should use compact version API from AndroidX, but WindowInsetsController actually behaves different from systemUiVisibility (I haven't deeply investigated it).

app/src/main/java/com/topjohnwu/magisk/arch/BaseUIActivity.kt
```diff
@@ -54,17 +57,15 @@ abstract class BaseUIActivity<VM : BaseViewModel, Binding : ViewDataBinding> :
...
+        window.decorView.windowInsetsController?.hide(WindowInsets.Type.navigationBars())
+        window.decorView.windowInsetsController?.systemBarsBehavior = WindowInsetsController.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
+        window.setDecorFitsSystemWindows(false)
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             window?.decorView?.post {
                 // If navigation bar is short enough (gesture navigation enabled), make it transparent
-                if (window.decorView.rootWindowInsets?.systemWindowInsetBottom ?: 0 < Resources.getSystem().displayMetrics.density * 40) {
+                if (window.decorView.rootWindowInsets?.isVisible(WindowInsets.Type.navigationBars()) == false) {
```
RikkaW, line 68: The original purpose is to guess if gesture navigation is enabled, NOT if the navigation bar is visible. Your change is totally wrong.

RikkaW (changes requested):

RikkaW:
> setSystemUIVisibility can't adapt accurate full screen on Android 11 when The device is Hollowed screen.

Changing setSystemUIVisibility: To allow contents to show below system bars
The next step is to change views' padding/margin according to windows insets. Switching to new APIs does not help you to do this :)

PR #5132, opened 2021-12-27
Explain how to extract boot.img from the ROM zip

timabell:
I had to do some googling to make the jump from "knowing I needed a `boot.img`" to the understanding that they were inside the zip I already had used to flash my LineageOS ROM. Hopefully this patch will make it easier for those who come after me.

The golang one is easier to use because it has prebuilt binaries that just work, whereas the second one requires a python + pip install and some additional setup which is a bit of a faff.

My entire journey is recorded at https://timwise.co.uk/2021/12/17/phone-setup-notes/#magisk if you're interested in more context.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -34,7 +34,12 @@ Otherwise, continue to [Patching Images](#patching-images).
 If your device has boot ramdisk, you need a copy of the `boot.img`<br>
 If your device does **NOT** have boot ramdisk, you need a copy of the `recovery.img`.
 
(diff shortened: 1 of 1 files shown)


timabell:
Is there a reason this PR was closed?

yujincheng08:
Conflicted and no more updates.

timabell:
That's odd, github isn't showing the conflict to me.

What update(s) would you need for this to be merge-able?

timabell:
I've force-pushed an update to the branch for this PR that fixes the merge conflict and re-adds the useful context I was trying to contribute. Let me know if you want a new PR or if there's a problem with the change.

fariborz4:
Yes

PR #5133, opened 2021-12-27
Add an old cgroup path

canyie:
Fix topjohnwu#5125
cgroup root path might be mem cgroup instead of acct, especially on low-ram devices.
https://android.googlesource.com/platform/system/core/+/bc131c3244a6aa4961092ba08285f9d0435a8882%5E%21/#F0


canyie:
Close and reopen to fix checks

topjohnwu (approved):

PR #5141, opened 2021-12-29
Fix stub build

yujincheng08:
Fix aapt optimize an apk twice.
See https://github.com/topjohnwu/Magisk/actions/runs/1632417805
and https://github.com/topjohnwu/Magisk/actions/runs/1632404765


canyie:
this closes #5147

PR #5142, opened 2021-12-29
fix appCache maxSize param type error

nocsg:
fix appCache maxSize param type error. `maxSize` is a long type


vvb2060:
https://docs.oracle.com/javase/specs/jls/se11/html/jls-5.html

PR #5155, opened 2022-01-01
Only store sepolicy rules into partitions in ext4 format

### LINKED ISSUE #5013: sepolicy.rule Wrong Directory Installation via Recovery/TWRP
Device: Xiaomi Redmi 4A (rolex)
Android version: Android 11
Magisk version name: 5d6d2417
Magisk version code: 23015

Something like this is happened again when I install a module with sepolicy.rule via Recovery/TWRP
https://github.com/topjohnwu/Magisk/issues/3735#issue-788092045

sepolicy.rule files are installed to /cache/magisk but the sepolicy.rules directory is linked to /persist/magisk
[magisk_log_2021-12-05T04_43_13Z.log](https://github.com/topjohnwu/Magisk/files/7654775/magisk_log_2021-12-05T04_43_13Z.log)

Screenshot_20211205-044414
Screenshot_20211205-044348
Screenshot_20211205-044402
Screenshot_20211205-044427

canyie:
Partially fix #5013
When installing from recovery, previous implementation may select f2fs partitions to store sepolicy rules, but magiskinit won't mount them and unable to load sepolicy rules.
This is not a complete fix because his device does not mount /persist partition in recovery mode, but I didn't mount it in this PR (see https://github.com/topjohnwu/Magisk/issues/5013#issuecomment-1003553933). I just submit this to make sure sepolicy.rule won't be installed to non-ext4 partitions.

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -628,13 +628,13 @@ copy_sepolicy_rules() {
     RULESDIR=$NVBASE/modules
   elif [ -d /data/unencrypted ] && ! grep ' /data ' /proc/mounts | grep -qE 'dm-|f2fs'; then
     RULESDIR=/data/unencrypted/magisk
-  elif grep -q ' /cache ' /proc/mounts; then
+  elif grep ' /cache ' /proc/mounts | grep -q 'ext4' ; then
     RULESDIR=/cache/magisk
-  elif grep -q ' /metadata ' /proc/mounts; then
+  elif grep ' /metadata ' /proc/mounts | grep -q 'ext4' ; then
     RULESDIR=/metadata/magisk
-  elif grep -q ' /persist ' /proc/mounts; then
+  elif grep ' /persist ' /proc/mounts | grep -q 'ext4' ; then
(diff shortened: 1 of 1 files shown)


osm0sis:
I think Magisk used to mount persist, so I guess this was a regression introduced at some point:
https://github.com/topjohnwu/Magisk/blob/master/scripts/util_functions.sh#L178

topjohnwu (approved):

canyie:
f2d2f28e23af00db4a776a6ded8b0c34f8129f62

git blame shows that umount was added by yourself...

osm0sis:
Right, because we were mounting it I think, but not unmounting after. Booted Magisk mounts persist into mirror, I think zip flash should again as well.

osm0sis:
Visible in that linked diff, even 👍: https://github.com/topjohnwu/Magisk/commit/f2d2f28e23af00db4a776a6ded8b0c34f8129f62#diff-f8cb6ae933f02a7a542c3f84299aaa23d8241aca56580fd7dcb64a92bce47692L199-L203

PR #5211, opened 2022-01-17
Prefer using existing Activity when downloads are finished

RikkaW:
This will avoid creating a new activity everytime.

The reason why it always returns to the home page after flashing is still unknown, but this is part of its solution.


yujincheng08:
Please fix the conflict.

yujincheng08:
close as stale

PR #5274, opened 2022-01-25
Added es, ro & ca missing strings.

ZillionMuffin:


ZillionMuffin:
As I see I've made some accidental deletions, sorry about that.

ArmynC (changes requested):
For the Romanian translation, there should be some changes. As an example, "Patchați" should be replaced by something around "Patchuiţi" for a 'mot-a-mot' way of writing, though there are other ways of doing it.

Next, for the "Modul suspendat pentru că %1$s este activat" you could do "Modulul este suspendat deoarece %1$s este activat" and so on for each iteration.

And finally, for the restore string, the term "ocultă" in this language has rather the meaning of a  paranormal thing/or a secret society. Instead, go for "Aplicaţia ascunsă va fi readusă la varianta iniţială. Sigur vreți să continuați?"

ZillionMuffin:
> For the Romanian translation, there should be some changes. As an example, "Patchați" should be replaced by something around "Patchuiţi" for a 'mot-a-mot' way of writing, though there are other ways of doing it.
> 
> Next, for the "Modul suspendat pentru că %1$s este activat" you could do "Modulul este suspendat deoarece %1$s este activat" and so on, for each iteration.
> 
> And finally, for the restore string. The term "ocultă" in this language has rather the meaning of a paranormal thing/secret society. Instead, go for "Aplicaţia ascunsă va fi readusă la varianta iniţială. Sigur vreți să continuați?"

I agree with that, gonna fix it later. My Romanian stills pretty basic.

ZillionMuffin:
On Spanish shouldn't be any error, only Catalan my have some missing things like l·l or ´,`

vvb2060:
Please fix build

ZillionMuffin:
Right now the xml files should be good. I just fixed what the output says.

app/src/main/res/values-ca/strings.xml
```diff
@@ -32,6 +33,7 @@
     <!--Install-->
     <string name="keep_force_encryption">Mantenir el xifrat forçat</string>
     <string name="keep_dm_verity">Mantenir AVB 2.0/dm-verity</string>
+    <string name="patch_vbmeta">Modifiqueu vbmeta en l'imatge d'inici</string>
```
Madis0, line 36:
```suggestion
    <string name="patch_vbmeta">Modifiqueu vbmeta en l\'imatge d\'inici</string>
```
Apostrophes need to be escaped with `\`, that's probably why your builds are failing. Do the same change for other strings as well.

ZillionMuffin:
"unsupport_system_app_msg" is giving problems, gonna fix it if I have time tomorrow

PR #5288, opened 2022-01-27
Fix crash when zip is malformat

yujincheng08:
`getNextEntry` may also throw `IllegalArgumentException`

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/utils/ZipUtils.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/utils/ZipUtils.kt
@@ -38,11 +38,7 @@ fun InputStream.unzip(folder: File, path: String, junkPath: Boolean) {
             }
             SuFileOutputStream.open(dest).use { out -> zin.copyTo(out) }
         }
-    } catch (e: IOException) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/topjohnwu/magisk/core/utils/ZipUtils.kt
```diff
@@ -41,5 +41,8 @@ fun InputStream.unzip(folder: File, path: String, junkPath: Boolean) {
     } catch (e: IOException) {
         e.printStackTrace()
         throw e
+    } catch (e: Throwable) {
+        e.printStackTrace()
+        throw IOException(e)
     }
```
vvb2060, line 47:
```suggestion
    } catch (e: IOException) {
        throw e
    } catch (e: IllegalArgumentException) {
        throw IOException(e)
    }
```

canyie, line 47:
```suggestion
    } catch (e: IllegalArgumentException) {
        throw IOException(e)
    }
```

topjohnwu (approved):

PR #5291, opened 2022-01-27
Add appcenter

vvb2060:


mywalkb:
I don't think @topjohnwu can accept that magisk send data to microsoft while he works for google. 😀

digitalcircuit:
As @androidacybot mentions, I'm uneasy with analytics on by default, opt out in a tool that's overseeing the root of trust on my device.

### Case study: Fritter

However, an approach similar to Fritter might be nice?
https://github.com/jonjomckay/fritter

Automatic reporting is off by default.  There's a dismissable prompt to send a crash report when errors do happen, with a way to always and never send crash reports.  In settings, you can change your mind later.

There's also a button in settings to "Say hello", which shows the full JSON details that will be sent while confirming if you want to send them.

If there's concern about not enough participation, Magisk in particular could ask about sending a ping on successful/unsuccessful installation (whether patching images, direct install, etc).  As additional options, Periodically Send and Never could be provided (so long as they can be changed later in settings).

### Thoughts on analytics

Ultimately, I suspect there will always be some folks who do not even want analytics integrated at all.  Fritter [has a `--no-analytics` build that's tested as part of the CI pipeline](https://github.com/jonjomckay/fritter/blob/9c5e273f0e46b63fc6a2462d80ac5aa9a32a6058/release.sh#L71 ).  AdAway (root) received a lot of pushback [on the off-by-default telemetry and crash reporting](https://github.com/AdAway/AdAway/wiki/Telemetry).  In a different realm, [Audacity had quite the uproar over telemetry being added](https://duckduckgo.com/?q=Audacity+telemetry ).

On the flip side, I was impressed when I updated the Tap, Tap app and [saw a dialog modeled after Android's permission prompts asking if I wanted to allow Internet access at all](https://github.com/KieronQuinn/TapTap/blob/612eb203b40052f5a42e5b623cfd375e96c154f8/app/src/main/res/values/strings.xml#L36 ), alongside explaining why and how it was used.  Obviously, the app could lie, but it showed a surprising level of concern for these matters.

In general, I suspect many folks are just tired of "tracking without asking is okay."  I personally am happy to send in statistics, I just prefer being asked first, since it acknowledges that this isn't something that's acceptable to do by default.

(Also, how many people use Magisk for systemless hosts ad-blocking?  If someone wants to send in stats, but doesn't want to globally allow centralized analytics platforms, something self hosted would avoid that.)

vvb2060:
If you care about privacy, please review the code and compile it yourself.  If you don't want to upload data, please don't use the canary version.

vvb2060:
You certainly have the right to opt out, please use the ci build, while this guarantees 100% that the apk is identical to the code.

josesilveiraa:
no merges?

PR #5310, opened 2022-01-29
Update italian translation

xDonatello:
A string is missing in the current Italian translation, which is present in the English version.
I added it and translated it.

### DIFF
--- a/app/src/main/res/values-it/strings.xml
+++ b/app/src/main/res/values-it/strings.xml
@@ -33,6 +33,7 @@
     <!--Install-->
     <string name="keep_force_encryption">Mantieni crittografia forzata</string>
     <string name="keep_dm_verity">Mantieni AVB 2.0/dm-verity</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-it/strings.xml
```diff
@@ -33,6 +33,7 @@
     <!--Install-->
     <string name="keep_force_encryption">Mantieni crittografia forzata</string>
     <string name="keep_dm_verity">Mantieni AVB 2.0/dm-verity</string>
+    <string name="patch_vbmeta">Correggi vbmeta nell'immagine di boot</string>
```
Madis0, line 36:
```suggestion
    <string name="patch_vbmeta">Correggi vbmeta nell\'immagine di boot</string>
```
Apostrophes need to be escaped with `\`, this fixes the automatic building as well

xDonatello: D'Oh! I forgot it. Thanks.

topjohnwu (approved):

PR #5318, opened 2022-01-29
Clarify "show OS apps"

Madis0:
~Looking at how the "show OS apps" toggle works on Magisk 24.1, I think this is what you really meant - apps that are specific to a device model, added by the OEM.~
Clarification added by vvb2060, so maybe "updatable" vs "fixed" are better terms - updated the PR as well.

Right now the words "system" and "OS" are too similar and don't really convey the difference...

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -92,7 +92,7 @@
 
     <!--MagiskHide-->
     <string name="show_system_app">Show system apps</string>
-    <string name="show_os_app">Show OS apps</string>
+    <string name="show_os_app">Show device-specific apps</string>
     <string name="hide_filter_hint">Filter by name</string>
     <string name="hide_search">Search</string>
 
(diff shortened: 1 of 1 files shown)


vvb2060:
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/UserHandle.java;l=216

```java
    /**
     * Whether a UID belongs to a system core component or not.
     * @hide
     */
    public static boolean isCore(int uid) {
        if (uid >= 0) {
            final int appId = getAppId(uid);
            return appId < Process.FIRST_APPLICATION_UID;
        } else {
            return false;
        }
    }
```

vvb2060:
It should be "show core apps", but I think OS is more accurate.

Madis0:
Hmm. But what is "system" and what is "core"?

An alternative translation I thought about could be "updatable system apps" vs "fixed system apps" but I'm not sure if that's accurate either.

VD171:
Why the string "show_system_app" keeps working? Apparentely, it does not show nothing else.
Only "show_os_app" is working perfectly, showing every OS app.
Why not merge them?

Nitrovenom:
> It should be "show core apps", but I think OS is more accurate.

I think "Show system apps" will be more accurate

PR #5346, opened 2022-01-31
Fix (very) little typo in French translation

xerta555:


HuskyDG:
Don't see anything different than deleting a line?

xerta555:
> Don't see anything different rather than deleting a line?

Yes, so that it is not confused with the original English translation

PR #5371, opened 2022-02-03
Cleanup preload before unmounting with zygisk

Fox2Code:
This is pull request cleanup the preload register on the Andorid linker before unmount on a zygisk process.

The goal of DenyList is to avoid process tampering, and make zygisk unloads itself.
But the linker often still keep a zygisk `soinfo` reference inside the preload register.
This is not a problem if `zygisk` is still mounted on the device, but **it can be still a problem** if the `zygisk` binary is **not accessible**.

This code may not look the best and may need a bit more cleaning, but the code is working perfectly on my device.
Note: My device is OxygenOS (Stock) Android 11

Inspired from https://github.com/LSPosed/NativeDetector code, and used some elf reader (GPLv3) code.
The elf reader code was striped to the minimum and transformed for performance. (and stability)
(There was no tutorials on how to read elf properly, I wanted to make it my own reader from scratch but I couldn't)

I post this PR as a draft for the code to be revived and improved.
Please be kind as C++ is not my main programing language.

### DIFF
--- a/native/jni/zygisk/deny/deny.hpp
+++ b/native/jni/zygisk/deny/deny.hpp
@@ -21,6 +21,7 @@ void ls_list(int client);
 bool is_deny_target(int uid, std::string_view process);
 
 void revert_unmount();
+void cleanup_preload();
 
 extern std::atomic<bool> denylist_enforced;
 extern std::atomic<int> cached_manager_app_id;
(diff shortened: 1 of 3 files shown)


yujincheng08:
Before we add Zygisk detection to NativeDetector we asked John about how Magisk was going to deal with it. And John's attitude towards it was:

> Denylist is not for hiding so Magisk won't take any effort to deal with hiding.

Fox2Code:
I will fork Magisk then ¯\_(ツ)_/¯

Fox2Code:
Also made this for OEM concern as they might change something that crash hidden processes.

PR #5394, opened 2022-02-07
Fix disappearing of MainActivity in Recents

nie-drug:
Fix disappearing of MainActivity in Recents after triggering SuRequestActivity


vvb2060:
This cannot be solved so simply

PR #5403, opened 2022-02-07
Fix UID tracking

yujincheng08:


native/jni/zygisk/deny/utils.cpp
```diff
@@ -76,34 +76,19 @@ static void rescan_apps() {
...
-        data_path += '/';
-        data_path += pkg;
-        struct stat st{};
-        if (stat(data_path.data(), &st) != 0) {
+    dirent *entry;
+    struct stat st{};
+    char buf[PATH_MAX] = {0};
+    // For each user
+    while ((entry = xreaddir(data_dir.get()))) {
+        snprintf(buf, PATH_MAX, "%s/%s", entry->d_name, pkg.data());
```
topjohnwu, line 90: I prefer `snprintf(buf, sizeof(buf), ...)` over hardcoding the length value

yujincheng08, line 90:
```suggestion
        snprintf(buf, sizeof(buf), "%s/%s", entry->d_name, pkg.data());
```

topjohnwu (approved):

PR #5408, opened 2022-02-08
Proper escape `:` and `\` when binding intent

yujincheng08:


Fox2Code:
Why this is only needed on Android 11+ only? (Just curious)

yujincheng08:
> Why this is only needed on Android 11+ only? (Just curious)

https://cs.android.com/android/_/android/platform/frameworks/base/+/547bf5487d52b93c9fe183aa6d56459c170b17a4

topjohnwu (approved):

Issue #4914, opened 2021-11-09
Magisk fails to handle special sepolicy on Oculus Go


moocow1452:
Device:  Oculus Go (pacific)
Android version: 7.1.2
Magisk version name: 258e89c9
Magisk version code: 23013

Oculus released an unlocked build for the Oculus Go that allows user access to boot and system, as well as adb root, but not vbmeta or recovery. When patching Magisk into boot, it will flash, but on boot, it goes to a black void with a "no command" error on the display for about a minute, then goes to recovery and reboot immediately takes it back to recovery until the original boot.img is reflashed. I have a console-ramoops attached among  other things, but because reflashing boot.img wipes the file, and I cant get into the system without reflashing the good boot.img, I'm not sure how relevant it is to Magisk.

[magisk_install_log_2021-11-08T13.log](https://github.com/topjohnwu/Magisk/files/7500255/magisk_install_log_2021-11-08T13.log)
[console-ramoops.txt](https://github.com/topjohnwu/Magisk/files/7500256/console-ramoops.txt)
[boot.zip](https://github.com/topjohnwu/Magisk/files/7500257/boot.zip)
[proc-mounts.txt](https://github.com/topjohnwu/Magisk/files/7500260/proc-mounts.txt)
[magisk_patched-23013_FemzM.zip](https://github.com/topjohnwu/Magisk/files/7500261/magisk_patched-23013_FemzM.zip)

Thank you for your time.

osm0sis:
Oh wow, it runs on Android 7.1? Rough...

I don't see any mention of Magisk in that console-ramoops, so it doesn't seem like it's from the failed boot, unfortunately.

SARankDirector-Minecraft:
> You'll have to see if there's any way to get proper logs.. since it's so old maybe it has a boot.img ramdisk and you can make it adb insecure using AIK?

why do we need adb insecure if we can do adb root and adb shell

osm0sis:
Yep you're probably right, hadn't reread his post.

moocow1452:
Not sure I can get ADB from the bootloop, whatever crashes happens before adb starts up, or at least adb on my PC cannot find the device after the magisk flash. If reflashing the original boot.img is the only way to bring the device to a bootable state, but it wipes any logging data, is there a recommended way to stick a script in the Magiskified boot that will log kernel messages to /sdcard, so that the logs will survive a reflash, but those changes would not otherwise interfere with the installation attempt?

[...]

PR #5411, opened 2022-02-08
Do not apply rootfs patch on pre oreo devices. (Fix #5124)

### LINKED ISSUE #5124: Repacking kernel causes bootloop on many MT6763 devices
Device: Oppo F5 Youth (CPH1725)
Android version: 7.1.1
Magisk version name: 92546e8a
Magisk version code: 23016

Hello everyone, I tried to root my device using Magisk .But after flashed the patched boot.img, my device bootlooped, it's stuck at Oppo logo, and then resetting itself again. 😭

When I flashed the before Magisk patched boot.img, it returns to normal

Here I uploaded the `/proc/last_kmsg` and `/proc/mounts`. I also included the... non-stock boot.img I used to patch (contexts in the bottom of this post). But I also used the stock boot.img, it didn't work either.

~[last_kmsg.txt](https://github.com/topjohnwu/Magisk/files/7777169/last_kmsg.txt)~ (wrong last_kmsg)
[mounts.txt](https://github.com/topjohnwu/Magisk/files/7777173/mounts.txt)
[boot.zip](https://github.com/topjohnwu/Magisk/files/7777179/boot.zip)

---

Some contexts about this device,
* The firmware that I used is version A41
* I think I accidentally broke the i2c interface? The battery doesn't want to charge, the touchscreen didn't work
* The boot.img I attached was modified before patched by Magisk. But the bootloop is not caused by this, because the stock boot.img also didn't work
  * I moved the system and data partition out from eMMC to external microSD, so it make me easier to browse the filesystem when I don't have the root
  * Modification to force enable the OTG function (it's on "SMANKUSORS defined code" section)

Fox2Code:
This hexpatch was made for Android 10 system-as-root to force it to load `rootfs`. 
So this patch shouldn't be needed under Android 10

I chose to use Andorid 8 as a minimum instead because of project treble that may be weird in some rare cases.
Also Andorid 8 having project treble in mind reduce the chances of this patch causing a boot loop.

Any version under Andorid 8 doesn't support anything else except `rootfs` as a boot mechanism.

Issue #5124 use Android 7, and tested disabling the patch, and it allowed him to boot successfully with Magisk enabled.

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -197,11 +197,15 @@ if [ -f kernel ]; then
   # After:  [mov w2, #-32768]
   ./magiskboot hexpatch kernel 821B8012 E2FF8F12
 
-  # Force kernel to load rootfs
-  # skip_initramfs -> want_initramfs
-  ./magiskboot hexpatch kernel \
-  736B69705F696E697472616D667300 \
-  77616E745F696E697472616D667300
+  # This patch is unneeded on pre-oreo devices
+  # and might boot loop them in some situations
+  if [ $API -ge 26 ]; then
+    # Force kernel to load rootfs
+    # skip_initramfs -> want_initramfs
+    ./magiskboot hexpatch kernel \
+    736B69705F696E697472616D667300 \
+    77616E745F696E697472616D667300
+  fi
 fi
 
 #################


Fox2Code:
Also another reason I chosen Andorid 8 is because of OTA might jump from Andorid 8 to Andorid 9, or to Andorid 9 to Andorid 10.

So I feel my decision of using Andorid 8 as the minimum version for applying this patch is a good idea. 
(Feel free to disagree)

Smankusors:
btw I'm wondering... do Magisk have a minimum Android version? Could it possibly run on Kitkat? 😅

mohammedrqa:
system for my Samsung Galaxy A6 sprint a6eltespr
sm-a600p
I want a modified ROM + TWRP recovery, please

iamshubh69 (approved):

canyie:
> Any version under Andorid 8 doesn't support anything else except rootfs as a boot mechanism.

Not always true. See #4914

canyie:
Hmmm... I think we can check if the device is actually a SAR device instead of checking version? 
```shell
if $SYSTEM_ROOT; then
  patch kernel
end
```

PR #5462, opened 2022-02-15
Add f2fs tuning for unencrypted devices

arter97:
Why tune f2fs here in Magisk?

f2fs being a LFS file-system, needs to perform GC periodically.
Android handles this via vold's IdleMaint service.

The kernel throttles this behavior as it can affect foreground performance
via gc_urgent_sleep_time, which is set to 500ms by default.

Android overrides this from init.rc to 50ms, as vold is pretty clever when
it comes to idle detection and 500ms takes too long to GC all invalid
segments in time, which results in long-term performance degradation.

Unfortunately, due to how Android init works, the logic to write 50 to
gc_urgent_sleep_time breaks when the device is unencrypted.
(The fix [1] for this is expected to ship with Android 13.)

Moreover, f2fs-specific optimizations such as cp_interval and discard
limit also breaks.

When Android added support for f2fs platform-wise, encryption was enforced
to OEMs, so while this isn't a problem for production devices running
unmodified OS, it is for rooted users.

Many people end up decrypting their device (some are forced to do so as
unlocked bootloader forces users to use unencrypted userdata), and they
will be met with storage performance going noticeably slower as the time
goes on (typically within 1-2 months).

As Magisk is the go-to solution for rooting your device, it makes the best
sense to implement a workaround for this in Magisk itself without relying
on a separate module.

The values were taken directly from AOSP and weren't changed in years.

[1] https://android-review.googlesource.com/c/platform/system/core/+/1717435
Signed-off-by: Juhyung Park <qkrwngud825@gmail.com>

### DIFF
--- a/native/jni/core/bootstages.cpp
+++ b/native/jni/core/bootstages.cpp
@@ -258,6 +258,85 @@ static bool check_key_combo() {
     return true;
 }
 
+#define F2FS_SYSFS_PATH "/sys/fs/f2fs"
+#define F2FS_DEF_CP_INTERVAL "60"
+#define F2FS_TUNE_CP_INTERVAL "200"
+#define F2FS_DEF_GC_THREAD_URGENT_SLEEP_TIME "500"
+#define F2FS_TUNE_GC_THREAD_URGENT_SLEEP_TIME "50"
+#define BLOCK_SYSFS_PATH "/sys/block"
+#define TUNE_DISCARD_MAX_BYTES "134217728"
+
+static inline bool tune_f2fs_target(const char *device) {
+    // Tune only SCSI (UFS), eMMC, NVMe and virtual devices
+    return !strncmp(device, "sd", 2) ||
+           !strncmp(device, "mmcblk", 6) ||
+           !strncmp(device, "nvme", 4) ||
+           !strncmp(device, "vd", 2) ||
+           !strncmp(device, "xvd", 3);
+}
+
+static void __tune_f2fs(const char *dir, const char *device, const char *node,
+                        const char *def, const char *val, bool wr_only) {
+    char path[128], buf[32];
+    int flags = F_OK | R_OK | W_OK;
+
+    sprintf(path, "%s/%s/%s", dir, device, node);
+
+    if (wr_only)
+        flags &= ~R_OK;
+    if (access(path, flags) != 0)
+        return;
+
+    int fd = xopen(path, wr_only ? O_WRONLY : O_RDWR);
+    if (fd < 0)
+        return;
+
+    if (!wr_only) {
+        ssize_t len;
+        len = xread(fd, buf, sizeof(buf));
+        if (buf[len - 1] == '\n')
+            buf[len - 1] = '\0';
+        if (strncmp(buf, def, len)) {
+            // Something else changed this node from the kernel's default.
+            // Pass.
+            LOGI("node %s unnecessary for tuning\n", node);
+            close(fd);
+            return;
+        }
+    }
+
+    xwrite(fd, val, strlen(val));
+    close(fd);
+
+    LOGI("node %s tuned to %s\n", path, val);
+}
+
+static void tune_f2fs() {
+    // Tune f2fs sysfs node
+    if (auto dir = xopen_dir(F2FS_SYSFS_PATH); dir) {
+        for (dirent *entry; (entry = readdir(dir.get()));) {
+            if (entry->d_name == "."sv || entry->d_name == ".."sv || !tune_f2fs_target(entry->d_name))
+                continue;
+
+            __tune_f2fs(F2FS_SYSFS_PATH, entry->d_name, "cp_interval",
+                F2FS_DEF_CP_INTERVAL, F2FS_TUNE_CP_INTERVAL, false);
+            __tune_f2fs(F2FS_SYSFS_PATH, entry->d_name, "gc_urgent_sleep_time",
+                F2FS_DEF_GC_THREAD_URGENT_SLEEP_TIME, F2FS_TUNE_GC_THREAD_URGENT_SLEEP_TIME, false);
+        }
+    }
+
+    // Tune block discard limit
+    if (auto dir = xopen_dir(BLOCK_SYSFS_PATH); dir) {
+        for (dirent *entry; (entry = readdir(dir.get()));) {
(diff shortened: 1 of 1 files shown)


arter97:
Fixed `!strncmp(device, "vd", 4)` to `!strncmp(device, "vd", 2)`.

vvb2060:
I think this should be handled by module

arter97:
> I think this should be handled by module

I gave a few reasons as to why I think it makes sense to do it in Magisk directly. What's your reason?

vvb2060:
Obviously, inside magisk is not the place to do this kind of thing, so we have provided module functions specifically for this case.

vvb2060:
Unless it's to fix a problem that affects magisk itself, like procfs hidepid
https://github.com/topjohnwu/Magisk/blob/4c80808997e392667d80bcc53da9e59d547913a7/native/jni/core/bootstages.cpp#L85-L88

arter97:
> Obviously, inside magisk is not the place to do this kind of thing, so we have provided module functions specifically for this case.

As I mentioned, some people may be forced to lose encryption upon rooting their device. I've seen several Samsung devices fall under this scenario.

If we were to put this workaround to a module, the user must be aware of it to avoid performance degradation.

Magisk has been doing device-specific workaround for ages. I don't see why this is any different (if any, this should be prioritized as this affects all Android versions).

https://twitter.com/arter97/status/1459479883805970433
This is a wide-spread issue. (Everyone who answered 500 in the replies are subject to this degradation.)

> Unless it's to fix a problem that affects magisk itself, like procfs hidepid

Yes, this doesn't affect Magisk directly, but it is very closely related to rooting itself.
Implementing a fix in Magisk is obviously not a must and definitely not its responsibility, doing so provides a better experience for people who would have gotten subjected to this issue.

vvb2060:
> Magisk has been doing device-specific workaround for ages. 

Only for stock ROMs, in order for magisk itself to work, not to help users fix bugs.

arter97:
> > Magisk has been doing device-specific workaround for ages.
> 
> Only for stock ROMs, in order for magisk itself to work, not to help users fix bugs.

This can happen on stock ROMs:
> As I mentioned, some people may be forced to lose encryption upon rooting their device. I've seen several Samsung devices fall under this scenario.

arter97:
eh.

TadiT7:
This deserves at least a warning in Magisk Manager.

daxiaamu:
> This deserves at least a warning in Magisk Manager.

indeed

ayanamist:
There is a module made by someone (not me) which you can try out
https://t.me/VaultAE/574

assembleDebug:
> There is a module made by someone (not me) which you can try out https://t.me/VaultAE/574

Doesn't seems to apply the values after boot.

daiaji:
> There is a module made by someone (not me) which you can try out https://t.me/VaultAE/574

Seems to be equally ineffective。

daiaji:
> Doesn't seems to apply the values after boot.

Since I couldn't contact the original author, I modified this module by myself, the bug seems to be due to a syntax error.
[F2FS_Tune.zip](https://github.com/topjohnwu/Magisk/files/8162056/F2FS_Tune.zip)

Stillhard:
> > Doesn't seems to apply the values after boot.
> 
> Since I couldn't contact the original author, I modified this module by myself, the bug seems to be due to a syntax error. [F2FS_Tune.zip](https://github.com/topjohnwu/Magisk/files/8162056/F2FS_Tune.zip)
Have you check/test the v2 released by the author [here](https://t.me/VaultAE/581)?

daiaji:
> Have you check/test the v2 released by the author [here](https://t.me/VaultAE/581)?

Honestly, just adding a delay to the command doesn't help.

Stillhard:
> > Have you check/test the v2 released by the author [here](https://t.me/VaultAE/581)?
> 
> Honestly, just adding a delay to the command doesn't help.

```
GoogleIsNotYourGod:/ $ su
GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/cp_interval
200
GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/gc_urgent_sleep_time
50
GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/iostat_enable
1
GoogleIsNotYourGod:/ # cat /sys/block/sda/queue/discard_max_bytes
134217728
GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/gc_urgent_sleep_time
50
GoogleIsNotYourGod:/ # ls -la /sys/fs/f2fs/*/gc_urgent_sleep_time
-r--r--r-- 1 root root 4096 2022-03-02 15:05 /sys/fs/f2fs/dm-7/gc_urgent_sleep_time
GoogleIsNotYourGod:/ #
```

Stillhard:
> > > > Have you check/test the v2 released by the author [here](https://t.me/VaultAE/581)?
> > > 
> > > 
> > > Honestly, just adding a delay to the command doesn't help.
> > 
> > 
> > ```
> > GoogleIsNotYourGod:/ $ su
> > GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/cp_interval
> > 200
> > GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/gc_urgent_sleep_time
> > 50
> > GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/iostat_enable
> > 1
> > GoogleIsNotYourGod:/ # cat /sys/block/sda/queue/discard_max_bytes
> > 134217728
> > GoogleIsNotYourGod:/ # cat /sys/fs/f2fs/*/gc_urgent_sleep_time
> > 50
> > GoogleIsNotYourGod:/ # ls -la /sys/fs/f2fs/*/gc_urgent_sleep_time
> > -r--r--r-- 1 root root 4096 2022-03-02 15:05 /sys/fs/f2fs/dm-7/gc_urgent_sleep_time
> > GoogleIsNotYourGod:/ #
> > ```
> 
> The essence of the problem is that the script syntax in this module is for bash, not the shell used by magisk.
> 
> At first, I struggled with the compatibility issue until Shellcheck suggested a change.
> 
> https://github.com/koalaman/shellcheck/wiki/SC3031

The original script(s) (v1 and v2) works in my test, but if your version is suitable for your system, it's not problem too, just don't say that it's:

> > There is a module made by someone (not me) which you can try out https://t.me/VaultAE/574
> 
> Seems to be equally ineffective.

PR #5486, opened 2022-02-23
Fallback to rw if mounting as ro fails

### LINKED ISSUE #5481: i can't use any module
Device: sm g130h
Android version: 5.1.1 lollipop
Magisk version name: 1a1b346c (24102)
Magisk version code: 1a1b346c (24102)
i try to install module example : busybox magisk and it success installed but i can't use it  even after reboot and i try check version busybox but still version 1.22.1
[my boot img](https://download1645.mediafire.com/9v51i3g475tg/5i59wjc89o1768n/magisk_patched-24102_cIioS.img) 
[magisk_log_2022-02-21T13_31_54.log](https://github.com/topjohnwu/Magisk/files/8106861/magisk_log_2022-02-21T13_31_54.log)

canyie:
If the source is used to mount read-write partitions, reusing the source to mount as read-only will fail with "Device or resource busy".

Fix #5481

### DIFF
--- a/native/jni/utils/xwrap.cpp
+++ b/native/jni/utils/xwrap.cpp
@@ -393,6 +393,11 @@ int xmount(const char *source, const char *target,
     const char *filesystemtype, unsigned long mountflags,
     const void *data) {
     int ret = mount(source, target, filesystemtype, mountflags, data);
+    if (ret < 0 && errno == EBUSY && (mountflags & MS_RDONLY) != 0) {
+        // This may happen when source is used to mount read-write partitions
+        LOGW("source %s cannot be mounted as read-only, fallback to read-write\n", source);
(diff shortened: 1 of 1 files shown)


canyie:
```
E : mount /sbin/.magisk/block/system->/sbin/.magisk/mirror/system failed with 16: Device or resource busy
E : lstat /sbin/.magisk/mirror/system/etc/hosts failed with 2: No such file or directory
SELinux:  Context x�ӷ  is not valid (left unmapped).
```
Btw, if getattr() failed, the `fileattr` structure is uninitialized and further use of it is undefined behavior. In #5481 the file's owner, permission and context was messed up. I think we should at least prevent this from happening?

HuskyDG:
Same as Android-x86 (BlissOS, PrimeOS, etc), its system partition is mounted read-write by default and Magisk fails to mount mirror then get `Device or resource busy` error

PR #5517, opened 2022-03-02
Update RU strings

Displax:


z8512:
Ho Russian. Why putin is invading ukraine

topjohnwu (approved):

PR #5546, opened 2022-03-05
Migrate to using GitHub issue forms

pixincreate:
When I was surfing around the Magisk on GitHub, I found that some of the issues have had ignored the issue template. It's really annoying when they just don't read what's written and blatantly write what ever they wish. 

And hence, I decided to re-write the existing BUG_REPORT.MD into issue forms which is a new feature which is currently in Beta. By deploying this, it'll force the user into reading / interact with the template. They cannot delete  or modify the content either. y employing this template, this would mean you no longer need to use the bot to automatically close the issue as the user is now forced to interact with. And this would also guarantee to reduce the number of spams and stuffs that an average user creates without reading anything. 

I've added a few more information that on top of the existing issue template to bring the issue into spotlight which would also help us find the exact problem.

You can see live demo on my fork _**[here](https://github.com/pixincreate/Magisk/issues/new/choose)**_.

Some other info about this issue forms:
- https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/
- https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository


vvb2060:
https://github.com/topjohnwu/Magisk/issues/new

pixincreate:
> https://github.com/topjohnwu/Magisk/issues/new

Should I create a new issue for this?

canyie:
I can still submit a spam issue without using the issues forms (https://github.com/pixincreate/Magisk/issues/1 ) so i think we still need the cute bot :(

pixincreate:
In case if you've used CLI to create an issue [here](https://github.com/pixincreate/Magisk/issues/1), I believe that is something what needs to be taken care of. (I don't have much experience with Git or GitHub).

canyie:
Actually I use an android app (OpenHub) to create it, but I believe CLI can do it too.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,116 @@
...
+        - 8.0
+        - 7.1
+        - 7.0
+        - 6
+        - 5.1
+        - 5.0
+        - 4.4
+        - 4.3
+        - 4.2
+        - 4.1
```
canyie, line 67: I think we can remove 4.1~4.4 as Magisk only supports Android 5.0+ since v23.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,116 @@
...
+    attributes:
+      label: Device architecture
+      multiple: false
+      options:
+        - arm
+        - arm64
+        - arm64-v8a
+        - armeabi-v7a
+        - x86
+        - x64
```
canyie, line 35: Aren't arm64 and arm64-v8a the same? Also, I think arm is actually armeabi-v7a because Magisk doesn't support armv5 which does not even have thumb-2 and hard float support.

canyie (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
pixincreate: Sure I'll do those if required.

pixincreate:
My intention is to have restrictive and unidirectional issue creation approach as every other issue that's created will be of same type of which I turn make it easy for everyone to read and understand the issue.


And I'm ready to apply the changes suggested by @canyie if you(@vvb2060) if you ever wish to merge the same in future.


NVM, thank you for showing some interest.

PR #5313, opened 2022-01-29
Remove temporary note about OTA update no longer working on Pixel devices

capntrips:
this option was restored in #5204

PR #5564, opened 2022-03-07
Update ota.md

V0latyle:
Update to reflect that "install to inactive slot" has been re-enabled as of Canary 23017.   Additional minor grammar correction.


osm0sis:
#5313

PR #5599, opened 2022-03-17
Use Shell cmd instead of systemReboot()

SageSyS:
This skips the system reboot animation and it avoids some of the problems I personally had on my surya (on multiple ROMs, firmwares and kernels) where my phone would just randomly freeze instead of rebooting, much appreciated.. 🥰

p.s. it's system reboot's fault, not magisk's, for some reason it's stabler to just call the reboot from the shell


vvb2060:
This is the same reboot method as the system framework, so why is there a problem?

PR #5602, opened 2022-03-17
Fix multiple modules adding the same subdirectory

### LINKED ISSUE #5511: Multiple modules adding the same subdirectory will fail to be added to the tree
Device: OnePlus 8T (kebab)
Android version: 11
Magisk version name: 7cfce9ff
Magisk version code: 24201

Discovered an interesting bug back in August which I reported internally and @topjohnwu said might be "an algorithm issue", but I presume it's since been forgotten so wanted to open it up to everyone to take a look since development has been much more collaborative lately (🎉)!

Basically, if 2+ modules have changes in the same newly added subdirectory (i.e. the subdirectory did not exist before either module) only the first loaded module's changes (alphabetical at boot) to that new subdirectory get added.

My example of this can be seen with /system/usr/libexec in the Git and SSH modules attached, but I have confirmed with @Didgeridoohan it can happen anywhere, and so is a decently large flaw in the Magisk skeleton tree creation process.

[magisk_log_2022-03-02T09_14_07.log](https://github.com/topjohnwu/Magisk/files/8169787/magisk_log_2022-03-02T09_14_07.log)

[Git_for_Android-2.23.0(2230).zip](https://github.com/topjohnwu/Magisk/files/8169526/Git_for_Android-2.23.0.2230.zip)
[SSH_for_Magisk-1.0(100)-osmod.zip](https://github.com/topjohnwu/Magisk/files/8169547/SSH_for_Magisk-1.0.100.-osmod.zip)

yujincheng08:
Fix #5511

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -28,6 +28,6 @@ android.nonTransitiveRClass=true
 
 # Magisk
 magisk.stubVersion=27
-magisk.versionCode=24302
+magisk.versionCode=24303
 magisk.ndkVersion=23b
 magisk.fullNdkVersion=23.1.7779620


osm0sis:
Seems to break a lot of modules 🤔

PR #5615, opened 2022-03-19
Device using split policy can still have monolithic sepolicy file

### LINKED ISSUE #5609: last canary won't work on samsung j5 2016
Device: Samsung J5 2016
Android version: 10 Los 17.1
Magisk version name: canary
Magisk version code:24302

Screenshot_20220318-120142_Magisk
[magisk_install_log_2022-03-18T11:53:43.log](https://github.com/topjohnwu/Magisk/files/8304005/magisk_install_log_2022-03-18T11.53.43.log)
[dmesg.log](https://github.com/topjohnwu/Magisk/files/8304012/dmesg.log)
[recovery.log](https://github.com/topjohnwu/Magisk/files/8304013/recovery.log)

canyie:
Fix #5609

### DIFF
--- a/native/jni/init/rootdir.cpp
+++ b/native/jni/init/rootdir.cpp
@@ -235,7 +235,7 @@ void SARBase::patch_ro_root() {
         }
     }
 
-    if (access("/sepolicy", F_OK) == 0) {
+    if (access(SPLIT_PLAT_CIL, F_OK) != 0 && access("/sepolicy", F_OK) == 0) {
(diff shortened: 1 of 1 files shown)


yujincheng08 (approved):

canyie:
Maybe rootfs devices also need this patch? I think we need to add early mounting back, but we don't want to do this. Perhaps we can always use hijack method even if monolithic sepolicy is present.

PR #5619, opened 2022-03-19
Don't always mock selinux enforce as "0"

### LINKED ISSUE #5607: Magiskinit causes `security_setenforce` fails with Invalid argument
Device: Realme 7 5G
Android version: 11
Magisk version name: Canary
Magisk version code: 24302

Upgrade and patch with or without vbmeta patch stock boot.img same error.
24301 was perfect
[magisk_install_log_2022-03-18T11_15_06[1].log](https://github.com/topjohnwu/Magisk/files/8303829/magisk_install_log_2022-03-18T11_15_06.1.log)
[magisk_install_log_2022-03-18T11_22_22[1].log](https://github.com/topjohnwu/Magisk/files/8303830/magisk_install_log_2022-03-18T11_22_22.1.log)

yujincheng08:
Some devices' SELinux will become enforced as soon as it loads sepolicy. And set it to enforcing again will crash `init`. So read the content again after sepolicy loaded and mock that result instead.

Fix #5607

### DIFF
--- a/native/jni/init/selinux.cpp
+++ b/native/jni/init/selinux.cpp
@@ -70,7 +70,6 @@ void MagiskInit::hijack_sepolicy() {
     } else {
         // We block using the "enforce" node
         blocking_target = SELINUX_ENFORCE;
-        actual_content = "0";
     }
 
     // Hijack the "load" and "enforce" node in selinuxfs to manipulate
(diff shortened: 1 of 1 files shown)


native/jni/init/selinux.cpp
```diff
@@ -169,6 +168,10 @@ void MagiskInit::hijack_sepolicy() {
     xmount("selinuxfs", REAL_SELINUXFS, "selinuxfs", 0, nullptr);
     sepol->to_file(REAL_SELINUXFS "/load");
 
+    if (strcmp(blocking_target, SELINUX_ENFORCE) == 0) {
```
topjohnwu, line 171: Can you move this into the hijack lambda?

topjohnwu (approved):

PR #5620, opened 2022-03-19
Replace `realpath` to BSD implementation

### LINKED ISSUE #5530: `readlink -f` from busybox is broken for kernel version < 3.6
Device: Nexus 7 (2013) WiFi (flo)
Android version: Android 11 (lineage 18.1 - lineage-18.1-20220228-nightly-flox-signed)
Magisk version name: 7cfce9f
Magisk version code: 24201

I patched the boot.img and got an magisk patched boot file. I booted into TWRP (twrp-3.6.0_9-0-flo) and installed the patched boot image. No errors. After booting the device freeze. I have to re-install org boot.img to make it work again.
I took magisk.apk and renamed it to magisk.zip. I trying to install magisk.zip with TWRP.
ERROR!

Magisk 24.2 Installer

! Cannot mount /system
'- Unmount partition
Prozess Updater endet mit FEHLER: 1
Fehler beim Installieren von ZIP '/sdcard/Download/magisk-v24.2.zip'
Partitions-Informationen werden aktualisiert...
... Fertig

yujincheng08:
Fix #5530

The thing is, ndk changes its `realpath` implementation to a new one that does not support kernel with version < 3.6. So, on some old devices or TWRP, `readlink -f` will fail. This replaces the `realpath` implementation to support older kernel versions.

@osm0sis Could you help me make this change to ndk-kitchen and co-author me? Thx.


yujincheng08:
This issue is introduced by NDK, see: https://github.com/android/ndk/issues/1260

And NDK seems not going to fix it. So sad.

osm0sis:
Not a ton of time with a 1-month old here; basically limited to stuff I can do on my phone in between/during feedings. 🥲😴😅

osm0sis:
Might be best to do this in the busybox source, similar to this: https://github.com/osm0sis/android-busybox-ndk/blob/master/patches/024-a-hush-platform.patch

It's how busybox handles broken/incomplete platforms.

osm0sis (changes requested):
Might be best to do this in the busybox source, similar to this: https://github.com/osm0sis/android-busybox-ndk/blob/master/patches/024-a-hush-platform.patch

It's how busybox handles broken/incomplete platforms.

yujincheng08:
> Might be best to do this in the busybox source, similar to this: https://github.com/osm0sis/android-busybox-ndk/blob/master/patches/024-a-hush-platform.patch
> 
> It's how busybox handles broken/incomplete platforms.

No. That patch is for undefined symbols but mine is for **replace** the bionic implementation. So doing something like your patch will raise duplicated symbols when linking.

osm0sis:
platform.h/.c it can be used to wrap and add various functions and calls, sometimes replacing ones from the libc used on a particular platform by defining a bb_ version instead.

yujincheng08:
https://github.com/mirror/busybox/blob/fc7868602ecf0d761a9a877141add4a9b6918d02/include/platform.h#L600
you mean this?

osm0sis:
Sort of? I might be also thinking about how I took the bb_ functions for ttyname and used them in my nano builds by defining the replacement for the useless NDK stub version of it: https://forum.xda-developers.com/t/tools-zips-scripts-osm0sis-odds-and-ends-multiple-devices-platforms.2239421/page-34#post-66486572

But busybox platform.h/.c should allow for something similar.

yujincheng08:
Ok, I see now. Instead of `-Wl,--wrap` them, busybox prefers to use macro `#define bb_x x` to replace `x` function. Well, as long as it can cover all source code, it's okay. I use `wrap` because I am afraid some source files don't include `platform.h` and use the original function from `libc.a`.

osm0sis:
Nah in busybox platform is in all since it has all the platform hacks.

osm0sis:
Sounds good. Can you let me know the first SDK version or NDK version where this issue started? I'll look into guarding it in a patch to platform.h with that, and upstream it in busybox if they'll accept it.

yujincheng08:
ndk 21 and ndk 22 have no problem. ndk 23 introduces the problem.

osm0sis:
Could this break anything else in Magisk? Should the used NDK perhaps be downgraded for now? I know @topjohnwu has done that in the past.

yujincheng08:
Only busybox is affected.

topjohnwu:
@osm0sis we shouldn't upstream this to BusyBox, as this is an issue with Bionic updating `realpath` to an implementation that requires higher Linux versions. Bionic does not really support static executables, so no considerations has been made when this Bionic change is pushed. If we build `busybox` as a dynamically linked executable (which 99.9% of people should, we only build busybox statically for no reason other that supporting recovery installations), there would be no issues at all.

osm0sis:
I mean, busybox is about supporting broken platforms in places so it's not entirely outside their usual, but 👍

PR #5626, opened 2022-03-21
Update Vietnamese translation

CDzungx:
Quick fix: Yes - "Đồng ý" -> "Có" to be more versatile.
image
Ramdisk isn't something to "agree" with.


abditc (commented):
AD ABDO

topjohnwu (approved):

PR #5637, opened 2022-03-23
Use svc for reboot to recovery

vvb2060:


osm0sis:
We intentionally had it not use the svc since it triggered some weird ROM dialogs that made people think their device was rebooting to be factory reset: https://github.com/topjohnwu/Magisk/commit/b1e63f0f140e9ef88e7bf19bfb266c95e4ae335d

vvb2060:
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/power/ShutdownThread.java;l=268;drc=c6c60534461c03f919a7196cb0161245b93b2126

osm0sis:
The issue wasn't with recovery mode users, it was with stock ROMs showing "factory reset" when triggering a reboot to recovery using svc.

osm0sis:
I'll see if I can reproduce the issue on any of my devices. I recall it used to be on OnePlus 5 era OOS, and I think MIUI Poco X2.

osm0sis (approved):
Found one device of mine that still has the issue - interestingly enough my Nexus 7 2013 on official flox Lineage 18.1.

Great idea for a workaround forcing the screen off. Works well everywhere I have tried. 👍

yujincheng08:
Reopen it to trigger GitHub action.

PR #5638, opened 2022-03-23
Verify app signature

vvb2060:


osm0sis:
Removing the .img files should probably remain indefinitely, since it'll clean up devices migrating from older Magisk versions.

osm0sis (changes requested):
Removing the .img files should probably remain in the code indefinitely, since it'll clean up any devices migrating from older Magisk versions.

vvb2060:
find_magisk_apk is only used for addon.d, when the new code is read, the magisk app version will not be lower than it is now.

I don't think it makes sense to support older versions from 3 years ago. Since the magisk file repository was modified, upgrading via in-app would first get to an intermediate version, by which time img would have been removed.

osm0sis:
Someone could directly update using the zip/apk...

Fox2Code:
The pull request should be converted to a draft to mark it as not finished.
The actual code get the certificate but doesn't check that the certificate match with the APK.

Also this code should be disabled on Debug builds because it would make harder to develop Magisk with this enabled.
Magisk developers should not waste their time on working around signature verification systems for debug builds.

Issue #4624, opened 2021-09-02
Magisk cannot detect key events correctly


pndwal:
Device: Redmi Note 8T
Android version: 10
Magisk version name: f822ca5
Magisk version code: 23001

**Key Combo Timing for Safe Mode on MIUI / Xiaomi**

I thought I'd try safe mode on my (Xiaomi) devices... (for my own safety).

I have Mi Pad 4 (LTE) w/ LOS18.1 (Android 11), and Redmi Note 8T w/ stock MIUI (Android 10).

[I have not been able to access Safe Mode from RN8T / MIUI from **booted system** any which way 🤨, but no problem on Mi Pad 4 / LOS.]

From **device powered down** I also had issues with timing on both Mi Pad 4 / LOS and RN8T / MIUI.

Holding the pwr button down for more than ~6 seconds after MI / REDMI logos appear causes black screen and fresh reboot, so it is necessary to release pwr after about 4 seconds.

[Releasing pwr then pressing vol- immediately after MI / REDMI logos appear doesn't trigger Safe Mode.]

After releasing pwr at ~ 4 seconds, then waiting (about 10 seconds more) for boot animation (Lineage / MIUI animated logos), then holding vol- until booted, I get Safe Mode but _Magisk has not detected button combo_, and MagiskHide / modules are still enabled.

**Working method:**

However, by holding the pwr button down and releasing it after about 4 seconds, then holding vol- during these logos as well as boot animation logos until booted I get safe mode and Magisk detects / disables MagiskHide and modules.

So clearly Magisk's _button detection is not reliable_ and device may boot to Safe Mode without Magisk detecting it.

It seems there is a critical timing sweet spot, at least for Xiaomi, where Safe Mode is triggered and Magisk also detects the key combination.

osm0sis:
May already be fixed upstream from the last Canary? https://github.com/topjohnwu/Magisk/commit/117d1ed0801c0b9fa144d195c0a89cabc488d176

pndwal:
> May already be fixed upstream from the last Canary? [117d1ed](https://github.com/topjohnwu/Magisk/commit/117d1ed0801c0b9fa144d195c0a89cabc488d176)

No, that must be something else - Entering Safe Mode is not the real issue; Magisk not detecting that this has occurred is.

I'm actually using @vvb2060's latest Alpha w/ TJW commits to 30 August already merged and this remains the same.

[...]

PR #5642, opened 2022-03-24
Add new way to trigger Magisk Safe Mode

HuskyDG:
On some devices, Magisk can't detect Safe mode, can't detect combo key or Safe mode doesn't work.
So we add a simple way to that if Magisk find specific files we created in Custom Recovery, it will trigger Magisk Safe Mode.


vvb2060:
https://github.com/topjohnwu/Magisk/issues/4624
We should fix the combo key detection, as this also used to detect recovery mode  users boot to true recovery.

PR #5658, opened 2022-03-29
Fix "magiskpolicy" missing when starting emulator

mirfatif:


canyie:
Already done in #5636 I think?

canyie:
Already done in #5636 I think?

PR #5670, opened 2022-03-31
Fix Notify & Logging snackbar

VladWinner:


vvb2060:
`it.notifyPropertyChanged`

Issue #5335, opened 2022-01-30
Cannot find vbmeta partition, patch vbmeta in boot image


ChuongPham:
Hi,

Device: 
Android version: Android 8.0
Magisk version name:  v24
Magisk version code: 24101
Magisk release downloaded from: https://github.com/topjohnwu/Magisk/releases

I encountered the following error when trying to upgrade Magisk v24.0 from Magisk v23.0 onto a Samsung stock-ROM Android 8.0 mobile device:

_**Cannot find vbmeta partition, patch vbmeta in boot image**_

Also, try a clean install of Magisk v24.0 [by uninstalling Magisk v23.0 first], still same error occurred.

I have rolled back to Magisk v23.0 as this version works on my mobile device.

Anyway, can you please check the error?

If you need me to provide image or files from the Samsung stock ROM, please let me know.

Thanks,
Chuong
[magisk_install_log_2022-01-30T213523.log](https://github.com/topjohnwu/Magisk/files/7965800/magisk_install_log_2022-01-30T213523.log)

MagiskBot:
Invalid bug report, automatically closed.
Please report issues using the latest canary Magisk version (version code: 24101).

osm0sis:
Provide the images, yeah. And output of `cat /proc/mounts`

ChuongPham:
@osm0sis: Thanks, for re-opening this issue.

As requested, please find attached a JPEG image showing the error mentioned in the OP and a dump of the command _cat /proc/mounts_ from my Android 8.0 Samsung mobile device.

Please let me know if you need to further information.

Kind regards,
Chuon
![magisk_vbmeta_error](https://user-images.githubusercontent.com/7285620/151775469-41fdc50a-1416-407b-9fd3-3f8194769792.jpg)
[mounts.txt](https://github.com/topjohnwu/Magisk/files/7969932/mounts.txt)
g

osm0sis:
No the boot.img ...

ChuongPham:
I'm new to Magisk.

Can you please tell me how I can provide the boot.img and the magisk_patched image from the latest Canary. I'll probably need instructions on how to accomplish this task.

The device boots fine, but Magisk is read-only - i.e. I'm unable to even install the Magisk Manager app, won't allow me to do this and keeps saying "installation failed".

osm0sis:
Flash the Canary from the link in the README.

ChuongPham:
Have just flashed Canary build from:

[...]

PR #5697, opened 2022-04-07
More friendly info

vvb2060:
#5335


osm0sis:
Actually this would match the other get_flags outputs best:

"- No vbmeta partition, patch vbmeta in boot image"

topjohnwu (approved):

PR #5759, opened 2022-04-22
Some fixes in readme

Nitrovenom:


HuskyDG:
`applications'` is actually short for `applications's` but because in English, `plural noun + ’` not `plural noun + ’s`

Nitrovenom:
> `applications'` is actually short for `applications's` but because in English, `plural noun + ’` not `plural noun + ’s`

Since there was every, I don't think there is need to put application in prural form.

PR #5774, opened 2022-04-30
Add info about android 11 recovery entering

nettnikl:
This took me a good time to understand, so this info might help anyone who tries to flash an updated magisk to an old device.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -76,7 +76,7 @@ As a summary, after installing Magisk in recovery **(starting from power off)**:
 - **(Recovery Key Combo) → (Splash screen) → (Long press volume up) → (Recovery Mode)**
 
 (Note: You **CANNOT** use custom recoveries to install or upgrade Magisk in this case!!)
-(Note: In Android version 11 and above, you have to have your phone conncted to a computer to start into recovery, and therefore also to start into Magsik!)
+(Note: Newer versions of Android (reportedly version 11 and above), do not allow to boot into Recovery mode the known way. To do so, you have to have your phone connected to a computer **and** hold the recovery button combination to start into recovery, (and therefore now to start into Magisk)!)
 
 ## Samsung (System-as-root)
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -76,6 +76,7 @@ As a summary, after installing Magisk in recovery **(starting from power off)**:
 - **(Recovery Key Combo) → (Splash screen) → (Long press volume up) → (Recovery Mode)**
 
 (Note: You **CANNOT** use custom recoveries to install or upgrade Magisk in this case!!)
+(Note: In Android version 11 and above, you have to have your phone conncted to a computer to start into recovery, and therefore also to start into Magsik!)
 
 ## Samsung (System-as-root)
 


nin0-dev:
You wrote "Magsik", but it's "Magisk". (Line 79)

vvb2060:
I don't understand what you mean…

nettnikl:
> You wrote "Magsik", but it's "Magisk". (Line 79)

Thanks for spotting!

nettnikl:
> I don't understand what you mean…

Hm, maybe i have to clarify more then. If you flash Android 11 or later, you cannot enter recovery mode without connecting the phone to a pc. So, to enter magisk, you have to do that, if you installed magisk as recovery.

vvb2060:
Can you explain why? What was done on the phone and why it didn't work, and what needs to be done on the PC.

nettnikl:
No, i cannot - this is a change in android, and i'm not a google engineer, so i don't understand their reason as well.
You don't have to do anything on PC, it's literally just connecting a power source to your phone, waiting for the charge symbol, then proceeding with starting your phone into recovery. You can test by yourself with a rooted Galaxy S10 (as you have to install Magisk to recovery here), if you have one at disposal.

vvb2060:
What did you do on the phone to try to reboot to recovery and what was the result.

nettnikl:
The "old" way also works on older Android versions, just not in 11 (or so i read) and above (i tested, and can confirm for 12).

nettnikl:
As a summary, after installing Magisk in recovery (starting from power off):

(Recovery Key Combo) → (Splash screen) → (Release all buttons) → (System with Magisk)

Does not work if the base system is based on Android 12.

The fix is described in my commit. The same info can be found all over the internet.

vvb2060:
what was the result？

without PC
(Recovery Key Combo) → (Splash screen) → (Release all buttons) → ???

without PC
(Recovery Key Combo) → (Splash screen) → (Long press volume up) → ???

connect PC
(Recovery Key Combo) → (Splash screen) → (Release all buttons) → ???

connect PC
(Recovery Key Combo) → (Splash screen) → (Long press volume up) → ???


Recovery Mode? 
Or System with Magisk? 
Or System with NO Magisk?

nettnikl:
With Android 11 or later:

without PC
(Recovery Key Combo) → (Splash screen) → (Release all buttons) → System with NO Magisk

without PC
(Recovery Key Combo) → (Splash screen) → (Long press volume up) → System with NO Magisk

connect PC
(Recovery Key Combo) → (Splash screen) → (Release all buttons) → System with Magisk

connect PC
(Recovery Key Combo) → (Splash screen) → (Long press volume up) → Recovery Mode

I can repeat the cases once again, when I'm back at my computer.

vvb2060:
Thanks for the explanation, it looks like the issue is related to #4624.

nettnikl:
Might be, i cannot test, as i don't own this device - but isn't this about a Android 10 phone?

vvb2060:
Maybe we should compare the kernel version instead of the Android version.

nettnikl:
Oh, yeah, sounds reasonable. I thought it is caused by Android 11, because i read that online. Let's verify.

Mine is 4.14.113-23727351

pndwal:
> Thanks for the explanation, it looks like the issue is related to #4624.

I don't think so... That issue is about Magisk Daemon failing to detect safe mode key combo presses due to timing inconsistencies...

This PR is simply to alert users to Samsung's new key combo requirement that usb is connected (even a headset apparently) for devices (seems may affect Exynos socs only) running One UI 3+ / Android 11+ for recovery and download modes.
https://www.naldotech.com/how-to-boot-into-recovery-mode-on-samsung-one-ui-3-0-new-method/

I'm guessing this was done to avoid accidental booting to special modes...

As key combos vary between devices, and even between Samsung models (eg. some use Bixby key while others don't have it ...), I'm not sure Magisk needs to make users aware of device specific requirements for entering special modes...

nettnikl:
Well, in this case, it is actually the example case, which does not work as described, which was very confusing for me (Samsung S10 is the example device mentioned here). Especially as it is already exeptional to have magisk installed on recocery nowadays (so there are not many devices this does not apply too which are installed on recovery at the same time).

At the very least, we might want to suggest that there are differences between Android versions, and change the example device mentioned - the way it currently stands, the example seems incorrect.

topjohnwu:
How to reboot into recovery varies device to device. Not all devices running Android 11+ requires connecting to a PC to reboot to recovery, and the guide should not specify any device specific stuffs.

nettnikl:
Then please remove Samsung Galaxy S10 as example. As mentioned, this example is now plainly wrong for all recent Android version.

PR #5778, opened 2022-04-30
Fix preload happens on main thread

canyie:
Since we now pre-heat Shell in Application.onCreate(), the GetShellCallback will get executed on the caller thread (in our case it's the main thread) immediately if we already have an available Shell, which will cause ANR and splash screen not showing.

Fix topjohnwu#5662


canyie:
Fixed in https://github.com/topjohnwu/Magisk/commit/cc79a96fa31c3c7b9d05177726465a0f126880dc

PR #5794, opened 2022-05-06
Fix sepolicy rules dir is not found in recovery

reiryuki:
- /persist and /metadata are not mounted in recovery by default causing sepolicy rules dir to not be detected
- recovery mounting appears to have been completely removed in error from util_functions.sh with https://github.com/topjohnwu/Magisk/commit/16e4c67992ead7705d6c8b10bd200f300aad451a#diff-f8cb6ae933f02a7a542c3f84299aaa23d8241aca56580fd7dcb64a92bce47692
- bring recovery mounts of sepolicy rules dir locations to mount_partitions from the uninstaller

PR #5799, opened 2022-05-07
Remove mounts after it's already handled by mount_partitions

reiryuki:
Remove mounts after it's already handled by mount_partitions https://github.com/topjohnwu/Magisk/pull/5794/commits


osm0sis:
Ah, here it is. Put this in your other pull request please, no need for a second.

Issue #5886, opened 2022-05-31
BUG: approve grant / deny root access can't be done on Android TV API 25 and FireOS 7.1.2 with D-PAD / Arrows


tsynik:
All Magisk Managers after Magisk Manager 7.5.1 have this bug. User can't press Allow or Deny buttons on root request dialogue. It happens on Android 7. It worked fine with Magisk 20.4 and Magisk Manager 7.5.1.

2nd D-PAD issue with DenyList - user can't choose app processes switches with D-PAD

Third D-PAD issue - user can't scroll magisk and superuser logs on 1st access with D-PAD

You can reproduce it in Studio emulator with Android TV image api25

https://user-images.githubusercontent.com/1551446/171085532-b1498a89-adcc-41e0-9857-0b6049d885b3.mov

https://user-images.githubusercontent.com/1551446/171085571-b7e1a7ac-8b46-4d0e-b3af-762dafb27da4.mov

Device: Android TV
Android version: 7.1 API25
Magisk version name: 24314
Magisk version code: 24314

vvb2060:
#5834

Issue #5834, opened 2022-05-14
Can't approve grant / deny root access with D-PAD on SDK <= 25


tsynik:
All Magisk Managers after Magisk Manager 7.5.1 have this bug. User can't press Allow or Deny buttons on root request dialogue. It happens on Android 7.x (API 25) and probably less. It worked fine with Magisk 20.4 and Magisk Manager 7.5.1.

Device: Fire TV 4K
Android version: 7.1.2
Magisk version name: 24310
Magisk version code: 24310

yujincheng08:
@RikkaW https://developer.android.com/training/tv/start/controllers

tsynik:
On Android 9+ Allow button gain focus after dialog window appears (and can be pressed or switched to Deny or list), but on 7.1.2 and less it doesn't, and pressing any D-PAD arrows or SELECT doesn't help at all.

tsynik:
[0001-fix-request-dialog-module-log-and-denylist-navigation.patch.zip](https://github.com/topjohnwu/Magisk/files/8736211/0001-fix-request-dialog-module-log-and-denylist-navigation.patch.zip)

@topjohnwu this patch fixes request dialog, module, log and denylist navigation with d-pad

vvb2060:
Can't reproduce
https://user-images.githubusercontent.com/26996262/170833336-4583e524-f7e7-4c03-8ab5-2adcc967a407.mp4

tsynik:
> Can't reproduce

This issue is about deny / grant root permission request _dialogue_, not that list.

vvb2060:
1:52

tsynik:
> 1:52

Device API level: 27? this happen on < 26

tsynik:
And just adding RequestFocus in acivity_request.xml 
![rf](https://user-images.githubusercontent.com/1551446/171072663-6797decf-0227-4324-8730-c18cbaf68066.png)
fixes this behavior for it

vvb2060:
> on Android < 9 
？

tsynik:
> ？

< != 9 but actually according emulator tests focus exists on api 26 and up and doesn't on api 25

PR #5893, opened 2022-05-31
fix D-PAD nav on API <=25

tsynik:
Described in #5886


vvb2060:
I don't know why so many modifications are needed #5890

tsynik:
> don't know why so many modifications are needed

some of them are Android Studio xml format, for log window there is initial focus on log frame for able o scroll, for superuser log windows there is focus attribute to able to scroll and o access Vack arrow button. For Modules there is clickable aand focusable options to allow select full block not only eelements on right with action to on/off module (bu you can just off) and focusRight to switches

vvb2060:
For log, it should be consistent with other tabs and not move the focus.

vvb2060:
For other, keep consistent with existing API > 26 behavior

tsynik:
> For log, it should be consistent with other tabs and not move the focus.

Where is initial focus when you click Log in main toolbar? It switches to Button and toolbar, not to actual viewed log on API 25, but if you click floating button and go to Superuser log and Back - Magisk log will be scrollable. You can't scroll it initialy.

tsynik:
> or other, keep consistent with existing API > 26 behavior

This behavior doesn't changed at all, all elements still do what they did. But in other parts of App you can navigate other full blocks, but not in modules. The same for DenyList and that's why you can't access process switches (full block not clickable so no expanded). Add full block on Modules clickable and focuable make it consistent with others.

vvb2060:
keep sync with https://github.com/topjohnwu/Magisk/issues/5834#issuecomment-1140289537

tsynik:
> For log, it should be consistent with other tabs and not move the focus.

here is video where you can see user unable to return from 2nd log window at all without mouse - https://user-images.githubusercontent.com/1551446/171092370-54b667eb-939a-4243-b1e4-00ca05289077.mov

vvb2060:
tv nav controller

tsynik:
> You just need to tap the back button

I pressed Back button on remote to return from Superuser log to Magisk log an now I can't navigate at all on real device with AndroidTV 11 (API 31). So this suggestion do Back but breaks navigation at all.

tsynik:
> For other, keep consistent with existing API > 26 behavior

In earlier Magisk releases Modules module block was selectable and clickable and it pops full module description onClick and highlighted on select as I remember.

app/src/main/res/layout/item_module_md2.xml
```diff
@@ -26,9 +26,12 @@
         <com.google.android.material.card.MaterialCardView
             style="@style/WidgetFoundation.Card"
             isEnabled="@{!item.removed &amp;&amp; item.enabled &amp;&amp; !item.showNotice}"
+            android:clickable="true"
+            android:focusable="true"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:nextFocusRight="@id/module_indicator"
```
tsynik, line 33: Note here is still nextFocusRight

yujincheng08:
close because #5890 has merged

tsynik:
> close because #5890 has merged

Log navigation issue still not fixed as modules inconsisency

Issue #5915, opened 2022-06-04
Magisk Manager: log window cant be scrolled with TV remote, D-PAD on Android TV 7, Fire OS 6 and ohers


tsynik:
Here is example video - https://user-images.githubusercontent.com/1551446/171092370-54b667eb-939a-4243-b1e4-00ca05289077.mov

Device: Fire TV 4K
Android version: 7.1.2
Magisk version name: 24315
Magisk version code: 24315

tsynik:
possible solution in #5917

yujincheng08:
I am merging Android TV related issues into one #5916.

PR #5917, opened 2022-06-05
other fix for #5915

tsynik:
This PR add D-PAD navigation to logs Manager section and corrects LogFragment var meaning.
If requestFocus will also be added to HorizontalScrollView magisk log will scroll without toggle superuser log first.


alnabeel11 (commented):
شكرا على المساعدة

PR #5960, opened 2022-06-10
Fix #5673

### LINKED ISSUE #5673: Failed to update in recovery patch mode
Device:  Galaxy S10
Android version: 12 One UI 4.1
Magisk version name:  canary https://github.com/topjohnwu/Magisk/commit/704f91545e8433c382af25c136d06bee07684d7e
Magisk version code:  24305 

First the installer is not detecting that is installed in recovery mode and not marking the check box

photo_2022-03-31_23-32-24

So I'm marking it and selecting direct install here are the logs

[magisk_install_log_2022-03-31T23_35_19.log](https://github.com/topjohnwu/Magisk/files/8394516/magisk_install_log_2022-03-31T23_35_19.log)

After reboot the device stills in same old version and shows the following prompt

photo_2022-03-31_23-32-15

Wich gets me to the first screenshoot  screen after accepting

Here are the current flashed boot.img and recovery,img, extracted from AP.tar patched  file using canary 24304 cause stable 24.3 was giving me vmbeta download mode error.

[recovery.img-boot.img-magisk_patched-24304.zip](https://github.com/topjohnwu/Magisk/files/8401039/recovery.img-boot.img-magisk_patched-24304.zip)

canyie:
util_functions.sh overrides `get_flags` function (defined in manager.sh), which sets `RECOVERYMODE` and causes `check_boot_ramdisk` not overriding the incorrect value.

- The "Install to recovery" checkbox does not take effect, needs more investigations.

Fix #5673

### DIFF
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -217,10 +217,11 @@ grep_prop() { return; }
 
 app_init() {
   mount_partitions
+  RAMDISKEXIST=false
+  check_boot_ramdisk && RAMDISKEXIST=true
   get_flags
   run_migrations
   SHA1=$(grep_prop SHA1 $MAGISKTMP/config)
-  check_boot_ramdisk && RAMDISKEXIST=true || RAMDISKEXIST=false
   check_encryption
   # Make sure RECOVERYMODE has value
   [ -z $RECOVERYMODE ] && RECOVERYMODE=false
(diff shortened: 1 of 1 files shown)


osm0sis:
&& || should be avoided, I'd suggest:
```
RAMDISKEXIST=false
check_boot_ramdisk && RAMDISKEXIST=true
```

osm0sis (approved):

topjohnwu (approved):

PR #6016, opened 2022-06-22
Update russian translation

LNDEVIL:
The line "Superuser" was changed and shortened and, since on devices with the Russian language it was too long and did not fit (because of this it showed ellipsis)

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -2,7 +2,7 @@
 
     <!--Sections-->
     <string name="modules">Модули</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-ru/strings.xml
```diff
@@ -2,7 +2,7 @@
 
     <!--Sections-->
     <string name="modules">Модули</string>
-    <string name="superuser">Суперпользователь</string>
+    <string name="superuser">Суперюзер</string>
```
Displax, line 5: In this case must be untranslatable. Please change to "Superuser".

topjohnwu (approved):

PR #6083, opened 2022-07-09
Update german translation

Henry2o1o:


meiser79:
There's a typo in the translation.

`<string name="home_follow_title">Folge Sie uns</string>`
should be
`<string name="home_follow_title">Folgen Sie uns</string>`

Henry2o1o:
> There's a typo in the translation.
> 
> `<string name="home_follow_title">Folge Sie uns</string>` should be `<string name="home_follow_title">Folgen Sie uns</string>`

thanks, edited

PR #6113, opened 2022-07-21
Support 32-bit emulators

canyie:
This would be helpful when debugging arch-related bugs, there is no reason to prevent it.


vvb2060:
https://github.com/topjohnwu/Magisk/blob/4264ae49c048d1778107cf688a03b4000643c931/scripts/avd_magisk.sh#L57

topjohnwu (approved):

PR #6173, opened 2022-08-07
Use `ANDROID_DLEXT_FORCE_LOAD` to load second stage if possible

### LINKED ISSUE #6095: Zygisk kills root
Device: Xiaomi Redmi Note 4
Android version: 7, MIUI Global 11.0.2
Magisk version name: 25.1
Magisk version code: 25101

Update to 25.1 from Magisk 24.2 (hide app faild) works.
After activating Zygisk, root disapears.

I had to install Magisk via TWRP because installing via fastboot didn't work.

Can you help?

[magisk_log_2022-07-12T01.15.36.log](https://github.com/topjohnwu/Magisk/files/9105688/magisk_log_2022-07-12T01.15.36.log)

canyie:
Fix #6095

The thing is, for some reason Zygisk 2nd stage fails to be triggered, leaves `MAGISK_INJ_2` in process environment. When app tries to exec `su` or any applet of `magisk`, its constructor detects that environment var and tries to do Zygisk 2nd stage setup. `magiskd` rejects request from the process because context mismatch, and makes the process crashes. (Fix is in #6172)

### DIFF
--- a/native/src/zygisk/entry.cpp
+++ b/native/src/zygisk/entry.cpp
@@ -51,16 +51,20 @@ static void *unload_first_stage(void *) {
     return nullptr;
 }
 
+#if defined(__LP64__)
+// Use symlink to workaround linker bug on old broken Android
+// https://issuetracker.google.com/issues/36914295
+#define SECOND_STAGE_PATH "/system/bin/app_process"
+#else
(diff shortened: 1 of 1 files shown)


canyie:
Build failed?? I didn't touch it??

PR #6252, opened 2022-09-07
Delete invalid deny list

rawer886:
If app reduces some processes after updated, then these processes cannot be removed from the denylist

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/ui/deny/AppProcessInfo.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/ui/deny/AppProcessInfo.kt
@@ -43,6 +43,14 @@ class AppProcessInfo(
     val packageName: String get() = info.packageName
     val processes = fetchProcesses(pm)
 
+    val unusedDenyList: MutableList<CmdlineListItem> by lazy {
+        this.denyList.filter {
+            processes.none { processInfo ->
+                it.packageName == processInfo.packageName && it.process == processInfo.name
+            }
+        }.toMutableList()
+    }
+
     override fun compareTo(other: AppProcessInfo) = comparator.compare(this, other)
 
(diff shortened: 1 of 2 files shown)


vvb2060:
I think invalid processes should only be deleted when the user updates the deny state of the app

vvb2060:
https://github.com/topjohnwu/Magisk/blob/6e299018a432deae49df454a9e5b5a2f9d57b9ce/native/src/zygisk/deny/utils.cpp#L287-L295

You can delete the process name parameter when it is not expanded.

rawer886:
> https://github.com/topjohnwu/Magisk/blob/6e299018a432deae49df454a9e5b5a2f9d57b9ce/native/src/zygisk/deny/utils.cpp#L287-L295
> 
> You can delete the process name parameter when list is not expanded.

The above code provide the ability to delete all processes. 
But in magisk app, you can only delete all the **current processes** of the application. If you want to delete the processes that existed before, you must execute "magisk --denylist rm [packagename]" in cli.

rawer886:
https://github.com/topjohnwu/Magisk/blob/928a16d8cc4b93bd10cd0568a9661591d623c79e/app/src/main/java/com/topjohnwu/magisk/ui/deny/AppProcessInfo.kt#L90-L95

@vvb2060 

E.g:  There are two records in denylist
```
com.xxx.demo|com.xxx.demo
com.xxx.demo|com.xxx.demo:abc
```
The `com.xxx.demo:abc` process is deleted after Demo upgrade. When I want to remove the Demo App from Denylist, I can only remove the process `com.xxx.demo|com.xxx.demo`, but `com.xxx.demo|com.xxx.demo:abc` cannot be operated.

yujincheng08:
afaik, this is expected

PR #6279, opened 2022-09-21
Create strings.xml

theunknownKiran:
Added Malayalam translation


Nitrovenom (commented):
Looks like a incomplete translation

theunknownKiran:
> Looks like a incomplete translation

The strings.xml under stub directory does have only 4 strings to translate. Am I missing something?
https://github.com/topjohnwu/Magisk/blob/master/stub/src/main/res/values/strings.xml

topjohnwu (approved):

Issue #6093, opened 2022-07-13
/overlay.d/custom.rc isnt injected!


Dima-diep:
Device Info:

Device: Xiaomi Mi Note 3
Android version: 10
Magisk version name: 25.1:MAGISK:R
Magisk version code: 25100

My ramdisk is contain "/overlay.d/custom.rc". But there is no inject init.rc, my services does not work!! Also, I can't find magisk-patched init.rc. In old versions, I could find modified init.rc in "/sbin/.magisk/rootdir/init.rc". But I can't find modified init.rc in current Magisk version! I think there is no any initrc-inject...

MagiskBot:
Invalid bug report, automatically closed.
Please report issues using the latest debug Magisk build (version code: 25101).

HuskyDG:
> In old versions, I could find modified init.rc in "/sbin/.magisk/rootdir/init.rc". But I can't find modified init.rc in current Magisk version! I think there is no any initrc-inject...

I knew why
https://github.com/topjohnwu/Magisk/blob/9474750bdf9a92e6d9154c2010dc00583da9e9b0/native/src/core/daemon.cpp#L370

Issue #6270, opened 2022-09-15
overlay.d not merging rc files


c0decafe:
Device: moto g200 5G (xpeng_p)
Android version: 12 (2022-07-01)
Magisk version name: 38ab6858
Magisk version code: 25203

essentially similar to #6093, contents of my custom rc file do not get merged into the system init nor any files within `/system/etc/init/*/*`.

i tested scripts under ./sbin as well and they get installed correctly, so issue seems to be specific to the rc handling.

the boot image layout is per the [docs](https://topjohnwu.github.io/Magisk/guides.html), verified by extracting it via dd and unpacking.

Is that a known problem (related the permission denied errors below?) or am i missing something?

``` console
xpeng:/sdcard/Download/out/ramdisk # find overlay.d
overlay.d/
overlay.d/custom.rc
overlay.d/sbin
overlay.d/sbin/magisk64.xz
overlay.d/sbin/magisk32.xz
```

``` console
xpeng:/sdcard/Download/out/ramdisk # cat overlay.d/custom.rc 
on pre-init
    mkdir /test 0775 root system
```

``` console
xpeng:/ #  dmesg | grep -i magisk
[   14.428464] init: starting service 'exec 48 (/dev/Hu7i/magisk --boot-complete)'...
[   14.430029] init: SVC_EXEC service 'exec 48 (/dev/Hu7i/magisk --boot-complete)' pid 3433 (uid 0 gid 0+0 context default) started; waiting...
[   14.435326] type=1400 audit(1663259959.387:117): avc: denied { read } for comm="magisk" name="bin" dev="dm-2" ino=44 scontext=u:r:magisk_client:s0 tcontext=u:object_r:vendor_file:s0 tclass=dir permissive=0
[   14.435360] type=1400 audit(1663259959.387:118): avc: denied { search } for comm="magisk" name="tests" dev="dm-8" ino=107 scontext=u:r:magisk_client:s0 tcontext=u:object_r:shell_test_data_file:s0 tclass=dir permissive=0
[   14.435375] type=1400 audit(1663259959.387:119): avc: denied { search } for comm="magisk" name="tests" dev="dm-8" ino=107 scontext=u:r:magisk_client:s0 tcontext=u:object_r:shell_test_data_file:s0 tclass=dir permissive=0
[...]

PR #6284, opened 2022-09-23
Do not delete $MAGISKTMP/.magisk/rootdir

HuskyDG:
Why does Magisk delete `$MAGISKTMP/.magisk/rootdir` while it is good for some case such as debugging overlay.d :)

https://github.com/topjohnwu/Magisk/issues/6093

### DIFF
--- a/native/src/core/daemon.cpp
+++ b/native/src/core/daemon.cpp
@@ -367,7 +367,6 @@ static void daemon_entry() {
             return true;
         });
     }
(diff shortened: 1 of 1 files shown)


vvb2060:
> But now that i can inspect the output init.rc i can confirm that is not an rc merge issue, but most likely a permissions issue.

https://github.com/topjohnwu/Magisk/issues/6270#issuecomment-1253527445

PR #6311, opened 2022-10-07
Support for installation of Github action artifacts

siavash79:
Github actions delivers the artifact files as zip files, even if they're zipped already.

This makes installation of the modules downloaded from Github actions confusing for users when they receive the "Unable to extract zip file" error.

This commit checks whether the zip file contains a nested zip, and uses that nested file as the installable module. The rest of the process is untouched


vvb2060:
We have an explicit module format specification and do not plan to support other formats.

PR #6336, opened 2022-10-17
CI: update github actions

armujahid:
This fixes this warning 
`
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-java, actions/setup-python, actions/cache, actions/cache, actions/upload-artifact, actions/cache, actions/cache, actions/setup-python, actions/setup-java, actions/checkout
`

Actions output on my fork https://github.com/armujahid/Magisk/actions (Everything is green)

1. Workflow warnings have been reduced from 58 to 10 after this change (You can compare https://github.com/armujahid/Magisk/actions/runs/3268381813 with https://github.com/topjohnwu/Magisk/actions/runs/3237568716
1. Build artifact of 28.6 MB is produced (This size remains same before and after my change indicating that everything is OK)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,8 +37,9 @@ jobs:
           fetch-depth: 0
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v3
         with:
+          distribution: 'zulu'
           java-version: '11'
 
       - name: Set up Python 3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,7 +94,7 @@ jobs:
       # Only upload artifacts built on Linux
       - name: Upload build artifact
         if: runner.os == 'Linux'
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -31,18 +31,19 @@ jobs:
...
+        uses: actions/checkout@v3
         with:
           submodules: 'recursive'
           fetch-depth: 0
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v3
         with:
+          distribution: 'zulu'
```
armujahid, line 42: zulu was implicitly the default distro in v1. But this distribution input is now mandatory
https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md

vvb2060:
distribution: 'temurin'
    java-version: '17'
    cache: 'gradle'

armujahid:
Done JDK 17 changes suggested by @vvb2060

PR #6340, opened 2022-10-19
Use stable random number seed

vvb2060:


HuskyDG:
After adding your commit, some people said that they couldn't flash custom kernel

HuskyDG:
After applying your commit, some people said that they failed to flash AnyKernel mod. I guess because of this
https://github.com/topjohnwu/Magisk/blob/6106a2298eeff60fe19b4496a44ac4392bc649bf/scripts/boot_patch.sh#L150
Revert your commit, the problem no longer occurs

IMG_20221022_095758_647
IMG_20221022_011730_714

vvb2060:
`config` is magisk internal file

vvb2060:
@osm0sis AnyKernel use this file?

HuskyDG:
https://github.com/osm0sis/AnyKernel3/blob/cf1a93136c74a45183bfbaba5860bf766b83a9b5/tools/ak3-core.sh#L342

osm0sis:
We don't do anything with the config file, no, we just check the boot partition ramdisk for backup/.magisk and extract it to repatch kernel binary/dtb if needed.

HuskyDG:
```
:/ $ su
:/ # cd /cache/backup
:/cache/backup # dd if=/dev/random of=config bs=8 count=1 2>/dev/null
:/cache/backup # echo -e "\n" >>config
:/cache/backup # cat << EOF >>config
> A=1
> B=2
> EOF
:/cache/backup # export $(cat config)
/system/bin/sh: export: �p]9e�: is not an identifier
1|:/cache/backup #
```

osm0sis:
Ah, I see, /dev/random can kick out some crazy unprintable characters.

vvb2060:
> /dev/random can kick out some crazy unprintable characters

Yes, so RANDOMSEED=XXX doesn't work. base64 would introduce unnecessary complexity, so I took this implementation.

osm0sis:
Maybe would be better just to send it > randomseed instead of config at all.

vvb2060:
I'm not sure if it should add new file, can you parse `config` like magisk?
https://github.com/topjohnwu/Magisk/blob/916e373edba1ca18b3356e51b7e0b6be9c18c5e1/scripts/util_functions.sh#L30-L36

osm0sis:
I don't pull specific vars, I just export the entire contents, less upkeep that way.

vvb2060:
I added `\n` after random stuff, so both `sed` and `grep` work fine.  Each prop name is longer than 8 bytes, there is no possibility of confusion.  
Since the length is fixed, the first 10 bytes can be skipped directly.

osm0sis:
I can't just start blanket skipping 10 bytes since AK3 can't know what version of Magisk the boot partition was patched by in advance.

vvb2060:
grep -q "RANDOMSEED=" config && dd if=config ibs=20 skip=1

PR #6370, opened 2022-11-04
Updated Russian Translation

Quanor:
The word "Superuser" has not been translated. I understand that its meaning will be clear to any Russian user who is able to install Magisk, but if you decided to translate the application into Russian, you should do it to the end


Displax:
https://github.com/topjohnwu/Magisk/pull/6016

PR #6374, opened 2022-11-08
Update ota.md

pritster5:
Specified the steps of the OTA install process so there is less confusion about what step 1 and 2 entail.


Bighead9971:
Submitted section lab for 6

PR #6396, opened 2022-11-21
Fix zygisk v4 ApiTable abi

yujincheng08:
Also, refactor some codes to let the compiler check the abi.


theenice13:
717-341-7730 text back

topjohnwu (approved):

PR #6398, opened 2022-11-21
Add inode plt hook APIs

yujincheng08:
Since xhook originally hooks by using regex which may cause confusion, we add this API for more specific hooks.


yujincheng08:
@topjohnwu Let's merge #6402 and #6396 first, and I will then update this PR.

PR #6445, opened 2022-12-14
Refactor magic mount by unsharing magiskd

### LINKED ISSUE #2359: Magisk Modules Incompatible with overlayfs
I'm running OxygenOS 10.3.0 on a OnePlus 7 Pro, and the Debloat module is not replacing files in /product correctly.

The `find` command indicates it's doing what it should (eg $MODPATH/product/app/badapp/.replace) and magisk logs indicate it was replaced.

However, the app is still installed as a system app and using a root file manager to examine the folders, the actual contents (not what should be magic mounted) shows up. 

Logs to follow.

_Sent from my OnePlus 7 Pro using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

### LINKED ISSUE #3338: Request for root directory partition support: optics and prism (Samsung), my_* (OPPO/OnePlus)
Hi - I am following on from Didgeridoohan's suggestion that i should raise a request to see if we can get Optics and Prism partitions mirrored on Samsung devices. The CSC features and the accompanying files which can activate hidden features are in these two partitions for most 2020 Samsung devices .

Can this be done? thread on xda here.......

https://forum.xda-developers.com/apps/magisk/getting-magisk-to-mirror-optics-prism-t4155533

yujincheng08:
Previously, magic mount creates its own mirror devices and mount mirror mount points. With these mirror mount points, magic mount can get the original files and directory trees. However, some devices use overlayfs to modify some mount points, and thus after magic mount, the overlayed files are missing because the mirror mount points do not contain the overlayed files. To address this issue and make magic mount more compatible, this patch refactors how magic mount works.

The new workflows are as follows:
1. unshare magiskd so that we can create some private mount points
2. make MAGISKTMP a private mount point so that we can create the private mount points there
3. for mirror mount points, we instead of creating our own mirror devices and mount the mirror mount points, we "copy" the original mount points by recursively mounting /
4. to prevent magic mount affecting the mirror mount points, we recursively set the mirror mount points private
5. to trace the mount points we created for reverting mounts, we again make the mirror mount points shared, and by this way we create a new peer group for each mirror mount points
6. as for tracing the newly created tmpfs mount point by magic mount, we create a dedicated tmpfs mount point for them, namely worker mount point, and obviously, it is shared as in a newly created peer group for tracing
7. when reverting mount points by magic mount, we can then trace the peer group id and unmount the mount points whose peer group ids are created by us

The advantages are as follows:
1. it is more compatible, (e.g., with overlayfs, fix #2359)
2. it can mount more partitions for which previous implementation cannot create mirror mount points (fix #3338)

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/utils/ShellInit.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/utils/ShellInit.kt
@@ -18,11 +18,6 @@ import java.util.jar.JarFile
 
 class ShellInit : Shell.Initializer() {
     override fun onInit(context: Context, shell: Shell): Boolean {
-        if (shell.isRoot) {
-            Info.isRooted = true
-            RootUtils.bindTask?.let { shell.execTask(it) }
-            RootUtils.bindTask = null
-        }
         shell.newJob().apply {
             add("export ASH_STANDALONE=1")
 
@@ -46,20 +41,22 @@ class ShellInit : Shell.Initializer() {
                 Info.noDataExec = !shell.newJob().add("$localBB sh -c \"$localBB true\"").exec().isSuccess
             }
 
+            val entry = if (shell.isRoot) "nsenter -t \$PPID -m sh" else "sh"
+
             if (Info.noDataExec) {
                 // Copy it out of /data to workaround Samsung bullshit
                 add(
                     "if [ -x \$MAGISKTMP/busybox/busybox ]; then",
                     "  cp -af $localBB \$MAGISKTMP/busybox/busybox",
-                    "  exec \$MAGISKTMP/busybox/busybox sh",
+                    "  exec \$MAGISKTMP/busybox/busybox $entry",
                     "else",
                     "  cp -af $localBB /dev/.busybox",
-                    "  exec /dev/.busybox sh",
+                    "  exec /dev/.busybox $entry",
                     "fi"
                 )
             } else {
                 // Directly execute the file
-                add("exec $localBB sh")
+                add("exec $localBB $entry")
             }
 
             add(context.rawResource(R.raw.manager))
@@ -68,6 +65,12 @@ class ShellInit : Shell.Initializer() {
             }
             add("app_init")
         }.exec()
+
+        if (shell.isRoot) {
+            Info.isRooted = true
+            RootUtils.bindTask?.let { shell.execTask(it) }
+            RootUtils.bindTask = null
+        }
 
         fun fastCmd(cmd: String) = ShellUtils.fastCmd(shell, cmd)
         fun getVar(name: String) = fastCmd("echo \$$name")
--- a/native/src/core/bootstages.cpp
+++ b/native/src/core/bootstages.cpp
@@ -1,6 +1,7 @@
 #include <sys/mount.h>
 #include <sys/wait.h>
 #include <sys/sysmacros.h>
+#include <sys/statvfs.h>
 #include <linux/input.h>
 #include <libgen.h>
 #include <vector>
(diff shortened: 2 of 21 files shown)


Displax:
Should we also add "system_dlkm", "odm_dlkm"?

vvb2060:
My opinion:
1. Don't parse mountinfo across processes.
2. Don't switch to init's ns, keep it in magiskd. magiskd's ns is still global and shared among all su processes. And busybox has tool to set ns.

native/src/init/mount.cpp
```diff
@@ -197,20 +196,11 @@ void MagiskInit::mount_rules_dir() {
     strcpy(p, "/persist");
     if (setup_block() < 0 || !do_mount("ext4"))
         return;
-    custom_rules_dir = path + "/magisk"s;
+    strcat(p, "/magisk");
 
 success:
     // Create symlinks so we don't need to go through this logic again
-    strcpy(p, "/sepolicy.rules");
```
vvb2060, line 204: https://github.com/topjohnwu/Magisk/commit/910a36fdc19eab87cf545ec513af54f4d0d9212d

app/src/main/java/com/topjohnwu/magisk/core/utils/ShellInit.kt
```diff
@@ -46,20 +41,22 @@ class ShellInit : Shell.Initializer() {
                 Info.noDataExec = !shell.newJob().add("$localBB sh -c \"$localBB true\"").exec().isSuccess
             }
 
+            val entry = if (shell.isRoot) "nsenter -t \$(pidof magiskd) -m sh" else "sh"
```
canyie, line 44: I guess this will be broken if someone (apps) change their process name to "magiskd"?

native/src/core/bootstages.cpp
```diff
@@ -24,97 +24,37 @@ bool zygisk_enabled = false;
...
-        LOGI("link: %s\n", buf1);
-        parse_mnt("/proc/mounts", [&](mntent *me) {
-            struct stat st;
-            if (MNT_DIR_IS("/") && me->mnt_type != "rootfs"sv && stat("/", &st) == 0) {
-                do_mount_mirror(system_root)
-                return false;
-            }
-            return true;
-        });
+    // first af all, try to setup module mount in the global namespace
```
canyie, line 28:
```suggestion
    // first of all, try to setup module mount in the global namespace
```

native/src/core/bootstages.cpp
```diff
@@ -24,97 +24,37 @@ bool zygisk_enabled = false;
...
+        chmod(SECURE_DIR, 0700);
     }
-    link_mirror(vendor)
-    link_mirror(product)
-    link_mirror(system_ext)
+
+    // then unshare for private mount points
+    xunshare(CLONE_NEWNS);
+
+    // make magisktmp a slave so that the mirror wont be visible
```
canyie, line 44:
```suggestion
    // make magisktmp a slave so that the mirror won't be visible
```

HuskyDG:
Did you test this on rootfs devices? It seems rootfs device cannot bind mount "/"

> mount / -> /sbin/.magisk/mirror failed with 22: Invalid argument

amosbird:
I've verified this PR on my RedMi K50. It successfully fixes the NFC issue on MIUI 14. Good job!

HuskyDG:
should this recursively bind mount?
https://github.com/topjohnwu/Magisk/blob/62042e61c1ea5eb12748fdd5699ae06a7696b4c2/native/src/core/module.cpp#L35

yujincheng08:
As old kernels (3.8- used by Android 5) has no `/proc/[pid]/ns/mnt`, nsenter and switch_ns won't work.

Issue #3820, opened 2021-02-04
Nothing installed in Lineage /system/addon.d/ via Magisk app Direct Install + fix_env


mdoggydog:
I installed Magisk by using the "patch boot.img and flash with fastboot" method, since the installation instructions say that using the installer zip via a custom recovery is no longer recommended.  On the next LineageOS OTA update, Magisk was not preserved, and I had to manually patch the boot.img and flash with fastboot again.

I discovered that `/system/addon.d/99-magisk.sh` was never installed, thus the OTA update would have no way to preserve Magisk.

Digging through the code, it appears that `/system/addon.d/99-magisk.h` is only installed when a Magisk installer zip is used via a custom recovery.

A related problem:  even if Magisk is initially installed via installer-zip/custom-recovery, it appears that updating Magisk via the MagiskManager app will never update an existing `/system/addon.d/99-magisk.sh` either.

This issue is basically a duplicate of #3782, which I think got prematurely closed because I did not explain the problem well enough the first time. Please see my last comment in #3782 for more analysis of this problem.

osm0sis:
It's supposed to install at some point via the app Direct Install too: https://github.com/topjohnwu/Magisk/blob/master/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt#L122

There was a plan to remove the unneeded installer code duplication and have everything run and maintained in one place via the shell scripts, but I'm not sure where that's at now.

mdoggydog:
Yep... all the bits are there.  This probably needs just a one- or two-line fix `MagiskInstaller.kt`, along the lines of `cp /data/adb/addon.d.sh /system/addon.d/99-magisk.sh` spelled the correct way (well, after testing if the `/system/addon.d/` directory exists).  But maybe there is some tricky subtlety involving security contexts or file ownership or something; this android system-level stuff is all new to me.

Half of the lines of code in https://github.com/topjohnwu/Magisk/blob/b6643b7bfcb05b81fc10dc786bc7ad073e581d69/scripts/flash_script.sh#L78-L85 are just ensuring that the target directory exists and is writable; I don't know if the writeability is an issue outside of the recovery environment.

osm0sis:
So looks like this was intentional. John said "I do not want to involve system rw at anytime during boot. This just opens up a whole can of worms..."

I won't close this as #wontfix since he may still change his mind. It would just need to use Magisk's own system mirror and rw it if the addon.d directory exists. 🤞

[...]

PR #6453, opened 2022-12-21
scripts: fix root loss until reboot after Magisk addon.d-v2

osm0sis:
- /system/bin/su was being removed in error from the live system so update remove_system_su to be aware of a running A/B OTA and generalize/simplify removal logic with relative paths to correctly run on the updated system slot

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -513,34 +513,39 @@ sign_chromeos() {
 }
 
 remove_system_su() {
-  if [ -f /system/bin/su -o -f /system/xbin/su ] && [ ! -f /su/bin/su ]; then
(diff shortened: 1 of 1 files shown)


pixincreate:
Does this fix #3820?

osm0sis:
No. That one will need some app changes.

This is an ongoing problem I noticed that nobody's made an issue for yet.

pixincreate:
Thanks for acknowledging the issue. That still exists sadly. If an OTA arrives, I've to keep Magisk app running else, it'll lose root access and doesn't let me install to the other slot on A/B device

osm0sis:
You simply have to flash Magisk in recovery or manually place the addon.d script as a workaround for the other issue.

topjohnwu (approved):

PR #6461, opened 2022-12-24
Add cmdline arg to chose architecture when building native binaries

xclusivor:


vvb2060:
We only support full ABI

PR #6470, opened 2022-12-26
Purify codes for module uninstallation

yujincheng08:


topjohnwu:
This is incorrect with several reasons:
1. `exec_module_scripts(stage, modules)` run scripts in parallel. We need to make sure `uninstall.sh` is done before we remove the module folder
2. `exec_module_scripts(stage)` only goes through active modules. With `remove_modules`, we want to actually remove every single existing module in the filesystem, which means using `foreach_module`

PR #6476, opened 2022-12-26
Use stable random seed

vvb2060:
Use only lowercase letters to match normal strings


yujincheng08 (changes requested):
There's no `magisk_cfg` anymore

yujincheng08 (approved):

canyie (approved):

PR #6480, opened 2022-12-27
Refactor sepolicy.rules resolve

yujincheng08:
We resolve available partitions for sepolicy.rules when patching boot and bind mount the partition by magiskinit.

For older devices, the previous logic won't work because the part name is never readable.

### DIFF
--- a/native/src/init/lib.rs
+++ b/native/src/init/lib.rs
@@ -1,8 +1,9 @@
+use std::env;
+use std::path::Path;
+
 pub use base;
 use base::read_lines;
 pub use logging::*;
-use std::env;
-use std::path::Path;
 
 mod logging;
 
@@ -10,32 +11,30 @@ mod logging;
 pub mod ffi2 {
     extern "Rust" {
         fn setup_klog();
-
-        fn rules_device_config() -> i32;
+        fn print_rules_device() -> i32;
     }
 }
 
-pub fn rules_device_config() -> i32 {
+pub fn print_rules_device() -> i32 {
     const UNKNOWN: i32 = 0;
     const PERSIST: i32 = UNKNOWN + 1;
     const METADATA: i32 = PERSIST + 1;
     const CACHE: i32 = METADATA + 1;
     const UNENCRYPTED: i32 = CACHE + 1;
     const DATA: i32 = UNENCRYPTED + 1;
-    const OLD: i32 = DATA + 1;
+    const EXISTING: i32 = DATA + 1;
 
     let encrypted = env::var("ISENCRYPTED").map_or(false, |var| var == "true");
 
     let mut matched = UNKNOWN;
-    let mut major = 0u32;
-    let mut minor = 0u32;
+    let mut rules_dev = String::new();
 
     if let Ok(lines) = read_lines("/proc/self/mountinfo") {
         for line in lines {
             if let Ok(line) = line {
                 let new_matched;
                 if line.contains("/.magisk/sepolicy.rules ") {
-                    new_matched = OLD;
+                    new_matched = EXISTING;
                 } else if line.contains(" - ext4 ") && !line.contains("/dm-") {
                     if line.contains(" / /cache ") && matched < CACHE {
                         new_matched = CACHE;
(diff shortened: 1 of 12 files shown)


yujincheng08:
@HuskyDG no, it wont

native/src/init/mount.cpp
```diff
@@ -117,99 +117,36 @@ static void switch_root(const string &path) {
...
-            return;
-        strcpy(p, MODULEROOT);
-        if (xaccess(path, F_OK) != 0) {
-            goto cache;
+    int major = 0;
+    int minor = 0;
+
+    parse_prop_file(".backup/.magisk", [&](auto key, auto value) -> bool {
+        if (key == "RULES_MAJOR") {
+            major = std::stoi(std::string(value));
```
canyie, line 125:
```suggestion
            major = parse_int(value);
```

native/src/init/mount.cpp
```diff
@@ -117,99 +117,36 @@ static void switch_root(const string &path) {
...
+        } else if (key == "RULES_MINOR") {
+            minor = std::stoi(std::string(value));
```
canyie, line 127:
```suggestion
            minor = parse_int(value);
```

vvb2060:
RULES_DEV=makedev(major, minor)

native/src/init/lib.rs
```diff
@@ -1,11 +1,87 @@
...
+pub fn rules_device_config() -> i32 {
+    const UNKNOWN: i32 = 0;
+    const PERSIST: i32 = UNKNOWN + 1;
+    const METADATA: i32 = PERSIST + 1;
+    const CACHE: i32 = METADATA + 1;
+    const UNENCRYPTED: i32 = CACHE + 1;
+    const DATA: i32 = UNENCRYPTED + 1;
+    const OLD: i32 = DATA + 1;
+
+    let encrypted = env::var("ISENCRYPTED").map_or(false, |var| var == "true");
```
shymega, line 27: This PR was brought to my attention, and I just wanted to mention that this line isn't very Rust-like at all. There are better approaches to this code, and I thought I'd leave an amendment below:

```
let encrypted = env::var("ISENCRYPTED").unwrap_or(String::from("false"));
```

Presuming you want the default value to be "false". If not, you can change `String::from("false")` to `String::from("true")`.

You can then use traditional comparison operators on the `String`, or `.eq(T)`. 

I really like Magisk, so this is my way of just helping out with Rust side of things.

shymega: Having looked it again, I see something else. You're comparing it to a boolean. I didn't see that part clearly, it was nested. You could use `bool::from_str`, but I'm not convinced using `..map_or` is very Rusty. You're using `.map_or` to get a default value, right?

vvb2060:
I don't think there's an advantage to parsing from native code, it's not performance sensitive code, and it's not going to be reused in native.
I'm all for rewriting shell code in native, but it should be a separate file, not moving methods to magiskinit and then calling them form shell, that's just weird.

native/src/init/lib.rs
yujincheng08: @shymega 
1. `encrypted` is bool, but `env::var("ISENCRYPTED").unwrap_or(String::from("false"))` is `String`, and thus it's not correct.
2. if use `bool::from_str`, I have to `bool::from_str(&env::var("r").unwrap_or(String::from("false"))[..]).unwrap_or(false)`. I don't think this verbose way is better.

Cyberfast2022:
Indeed, and I believe bootloops and boot failure won't end just there. That's true

native/src/init/lib.rs
shymega: @yujincheng08 Yeah, I got the wrong impression. I think your solution, now I have re-read the PR, works fine. My apologies for the misunderstanding - your way is cleaner.

native/src/init/mount.cpp
```diff
@@ -117,24 +117,12 @@ static void switch_root(const string &path) {
...
-            major = parse_int(value);
-        } else if (key == "RULES_MINOR") {
-            minor = parse_int(value);
-        }
-        return !major || !minor;
-    });
-
-    if (!major || !minor) return;
-    xmknod(BLOCKDIR "/rules", S_IFBLK | 0600, makedev(major, minor));
+    if (config->rules_dev == 0)
```
yujincheng08, line 120: second stage wont load BootConfig and thus this is always false

topjohnwu (approved):

PR #6502, opened 2023-01-03
Create changes.md

Tahiel01:


Tahiel01:
Ho

illgetit87:
Hey, my user name is ***@***.*** I'm a huge fan of magisk. I was wondering if
someone is able to help me with my phone. I'm from Canada and looking to
unlock bootloader and root with magisk. Also wondering if there is a Custom
recovery (TWRP or Orangefox). If you can help me or can point me in the
right direction that would be great too

Device: moto g pure
Codename: Ellis
Model #: XT2163-4
Android number: 12
Build number: S3RHS32.20-42-10-2

illgetit87:
Sorry I thought I would add the screen shot of build number

PR #6530, opened 2023-01-14
Make changelog scrollable in module update dialog

chenxiaolong:
This also removes the 1000 code point limit because the entire text is already loaded into memory and the truncation can sometimes break formatting.

When the changelog is short, the UI behavior is exactly the same as it is now (dialog with short height). When the changelog is long, the dialog expands to fill the screen, adding scrollbars if necessary (eg: [screenshot](https://user-images.githubusercontent.com/646253/212441443-f6c248cb-a98f-4a53-ad82-cac7928339d5.png)).


vvb2060:
We intentionally limit the changelog length

PR #6549, opened 2023-01-20
unshare magiskd

vvb2060:
su processes using the global namespace are not affected and still get the same file view as before and shared with other su processes.
zygisk modules may use files in magisktmp (new path is `/proc/magiskd/root/sbin`), so add selinux rules to allow.


vvb2060:
this is intentional

PR #6576, opened 2023-01-31
Ignore AMLogic "normal" slot suffix in scripts as well

### LINKED ISSUE #6572: Flashing Magisk.zip fails on Amlogic slot a-only devices due to slot 'normal' naming convention
Device: FireTV 2nd gen Cube
Android version: Android 9 (FireOS 7)
Magisk version name: Canary 831a398b
Magisk version code: 25206
Amlogic s922x SOC, slot-A only, SAR, no super partition device

When trying to flash magisk.zip within TWRP, to patch the boot partition I get the following error:
```
**********************************
 Magisk 831a398b(25206) Installer 
**********************************
- Current boot slot: normal
! Cannot mount /system
- Unmounting partitions
Updater process ended with ERROR: 1
Error installing zip file '/data/media/0/magisk-canary.zip'
Done processing script file
```
I think this may be related to some Amlogic devices using the slot label 'normal'.  Previously, I was not able to get this device to boot with a patched Magisk boot, because Magisk didn't recognize what slot 'normal' is.
https://github.com/topjohnwu/Magisk/issues/5806
In that instance @canyie applied a fix to whitelist 'normal'.  I'm wondering if a similar whitelisting is needed in this case?

osm0sis:
Fixes #6572

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -268,6 +268,7 @@ mount_partitions() {
     SLOT=`grep_cmdline androidboot.slot`
     [ -z $SLOT ] || SLOT=_${SLOT}
   fi
+  [ "$SLOT" = "normal" ] && unset SLOT
   [ -z $SLOT ] || ui_print "- Current boot slot: $SLOT"
 
   # Mount ro partitions


finnzz:
> Fixes #6572

Thank you, this fixed the ability to install the Magisk-patch to the boot image from within TWRP, as well as from within Magisk Manager.

PR #6578, opened 2023-02-01
update & fix Hebrew translation

YuvalZilber:
I'm a native Hebrew speaker.
As a community member that helps people hack Android, I am familiar with the tech language.
I tried to check the context of any text I changed/added, but some were trivial to translate, so I didn't put too much effort into it.


topjohnwu:
Why is Magisk translated into a native language? It is a proper noun and should not be translated?

Issue #6585, opened 2023-02-03
Magisk ignores `init_boot.img` in tar file


DrKetan:
Device: Samsung S23 Ultra
Android version: A13
Magisk version name:831a398b
Magisk version code:25206

Patched samsung S23 Ultra (SM-S918B) kernel & vbmeta with Magisk and Flashed without error, Magisk not installed . Log and kernel (patched & stock) included

[Magisk_debug_s23u.zip](https://github.com/topjohnwu/Magisk/files/10582765/Magisk_debug_s23u.zip)

[boot_Patched.zip](https://github.com/topjohnwu/Magisk/files/10582772/boot_Patched.zip)
[boot_stock.zip](https://github.com/topjohnwu/Magisk/files/10582777/boot_stock.zip)

BlackMesa123:
S23 device family shipped with Android 13 meaning it uses the new boot [header version 4](https://source.android.com/docs/core/architecture/bootloader/boot-image-header?hl=en#header-v4), adding the new `init_boot.img` in the [tar patching](https://github.com/topjohnwu/Magisk/blob/57d83635c6512f5e58753c7a1ae2d515c18cb70f/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt#L177) process should fix the issue.

canyie:
Upload installation log？

canyie:
After taking a look at the code I believe the issue exists, but "simply setting `srcBoot` to `init_boot.img` if exists" will cause some kernel patches failed to apply. Not sure if Samsung's GKI kernel still need these kernel patches, need further investigation on it.

BlackMesa123:
> After taking a look at the code I believe the issue exists, but "simply setting `srcBoot` to `init_boot.img` if exists" will cause some kernel patches failed to apply. Not sure if Samsung's GKI kernel still need these kernel patches, need further investigation on it.

Are you talking about the Samsung-specific kernel image patches? Are these still working nowadays? Since they’re pretty old (RKP patch from SuperSU and the Defex one is 5 yo). Would be great if someone could test a patched init_boot.img to check if everything is okay

BlackMesa123:
I also was able to extract the kernel config from the stock image and it looks like some security kernel components aren't anymore a thing, probably due to how GKI works (talking about microHypervisor, RKP etc., of course we can’t be sure until we get our hands on the kernel sources)

canyie:
[...]

PR #6586, opened 2023-02-06
Add `init_boot.img` patching for Samsung tar firmware packages

BlackMesa123:
As discussed in #6585, Samsung devices shipping with Android 13 (Eg. S23 family, A14 5G) follow Google guidelines for [generic boot partitions](https://source.android.com/docs/core/architecture/partitions/generic-boot?hl=en), thus moving the ramdisk from `boot.img` to `init_boot.img`. Since most of the users patch the tar firmware packages provided by Samsung rather than the single image individually, add support for `init_boot.img` patching in the `processTar` function.


yujincheng08 (approved):

RV7PR:
Works on my S23 Plus🎉

PR #6597, opened 2023-02-11
No kernel repack if it isn't patched at all 

### LINKED ISSUE #5124: Repacking kernel causes bootloop on many MT6763 devices

### LINKED ISSUE #6204: Ulefone Power 3 - boot loop
## For humans

I tried to root this old device using the image patching technique but the phone ended up in a boot loop so I had to reflash the unpatched `boot.img`.

The phone is freshly reinstalled to stock (last update) official ROM.

As there is no separate vbmeta partition, I always checked `Patch vbmeta in boot image` option. I tried with both `Preserve force encryption` checked and unchecked. I did not tried to check `Preserve AVB 2.0/dm-verity`.

Note: It seems that the image patching technique was reported working back in the days when magisk was v7.5 (not 100% sure).

- Device: Ulefone Power 3
- Android version: 7.1.1
- Android security patch level: May 5, 2018
- Kernel version: `4.4.22+release@release32 #2 Fri Nov 24 19:25:19 CST 2017`
- Build Number: Power_3_G03
- Magisk version name: 25.2
- Magisk version code: 9b61bdfc (25201) (D)
- `/proc/mounts` logs: https://termbin.com/7t3w
- `/proc/last_kmsg`: https://termbin.com/j5jl
- unpatched `boot.img`: https://www36.zippyshare.com/v/gh1nvCpf/file.html
- boot ramdisk: yes
- separate vbmeta partition: no
- A/B (Seamless) System Updates: no

## For MagiskBot

Trying to make [MagiskBot](https://github.com/MagiskBot) happy (past fails #6202 #6203)

Device: Ulefone Power 3
Android version: 7.1.1
Magisk version name: 25.2
Magisk version code: 25201

### LINKED ISSUE #6566: Oppo A83 CPH1729 - bootloop
**I tried to root this old device using patch boot image technique but ended up bootlooping**

**These options are ticked when patching:**
- Preserve force encryption
- Patch vbmeta in boot image

**Some information about this device:**
- Device: Oppo A83
- Model: CPH1729
- Android version: 7.1.1
- Kernel version: 4.4.22-G202009032342
- Android Security Patch Level: October 5, 2019
- Magisk version name: 25.2 (25206)
- Magisk version code: 831a398b (25206) (D)
- Ramdisk: Yes
- A/B partition: No

**Files**

- Stock boot image: [here](https://drive.google.com/file/d/1NrAsuLtidvfgvBOvI2nN-QgkXxtswI13/view?usp=sharing)

- Patched boot image: [here](https://drive.google.com/file/d/1hZP0v8DIBkr8cDOj1U87kzNDRGesqKWH/view?usp=sharing)

- Log file after patching boot image: [here](https://drive.google.com/file/d/1QR5BVw6LX9RRdQyai1qWEtSQXARaoO7w/view?usp=sharing)

- Bug report file: [here](https://drive.google.com/file/d/1yMJCT3kd_gCxJnN5WbZAGPDBb0jNzMki/view?usp=sharing) (contains last_kmsg and dmesg)

- /proc/mounts: [here](https://drive.google.com/file/d/1Owx74615jPQv5NDIsPR-EZph6CI_QlOJ/view?usp=sharing)
___
Device: Oppo A83
Android version: 7.1.1
Magisk version name: 25.2
Magisk version code: 25206

canyie:
It turns out that decompressing and recompressing the kernel is enough to break booting on many devices that use MT6763.
So don't repack the kernel if it isn't patched.
As the kernel contains "skip_initramfs" but actually the device uses rootfs, we also need to check if the device is SAR before applying the SAR patch

Fix #5124, fix #6204, fix #6566

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -222,7 +222,7 @@ if [ -f kernel ]; then
   736B69705F696E697472616D667300 \
   77616E745F696E697472616D667300 \
   && PATCHEDKERNEL=true
-  
+
   # If the kernel doesn't need to be patched at all,
   # keep raw kernel to avoid bootloops on some weird devices
   $PATCHEDKERNEL || rm -f kernel
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -6,7 +6,7 @@
 # Usage: boot_patch.sh <bootimage>
 #
 # The following flags can be set in environment variables:
-# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE
+# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT
 #
 # This script should be placed in a directory with the following files:
 #
@@ -73,6 +73,7 @@ fi
 [ -z $KEEPFORCEENCRYPT ] && KEEPFORCEENCRYPT=false
 [ -z $PATCHVBMETAFLAG ] && PATCHVBMETAFLAG=false
 [ -z $RECOVERYMODE ] && RECOVERYMODE=false
+[ -z $SYSTEM_ROOT ] && SYSTEM_ROOT=false
 export KEEPVERITY
 export KEEPFORCEENCRYPT
 export PATCHVBMETAFLAG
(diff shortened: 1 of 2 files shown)


yujincheng08 (approved):

scripts/boot_patch.sh
```diff
@@ -203,21 +204,28 @@ for dt in dtb kernel_dtb extra; do
...
+  # Force kernel to load rootfs for legacy SAR devices
   # skip_initramfs -> want_initramfs
-  ./magiskboot hexpatch kernel \
+  $SYSTEM_ROOT && ./magiskboot hexpatch kernel \
   736B69705F696E697472616D667300 \
-  77616E745F696E697472616D667300
+  77616E745F696E697472616D667300 \
+  && PATCHEDKERNEL=true
+  
+  # If the kernel doesn't need to be patched at all,
```
vvb2060, line 226:
```suggestion
  && PATCHEDKERNEL=true

  # If the kernel doesn't need to be patched at all,
```

topjohnwu (approved):

osm0sis:
Just occurred to me that this could leave a weird case if the kernel has a kernel_dtb, so perhaps someone can clarify; would a patched kernel_dtb replace the one attached to the untouched kernel in the .img (good), or would it just be appended to the untouched kernel, growing the kernel and boot.img size unnecessarily (bad)?

osm0sis:
Found one edge case, which is when there's a replacement kernel file which is already compressed and has kernel_dtb attached, the existing kernel_dtb will just get appended to it, increasing the kernel+kernel_dtb size like I worried about.

I don't think this is an issue Magisk itself will run into, but other projects that use magiskboot (TWRP, kernel community, etc.) will; I'll see if I can address it with a patch to the `magiskboot split` applet or something. 👍

PR #6610, opened 2023-02-17
Allow modifying denylist configuration without enabling zygisk in magisk app

HuskyDG:
I knew denylist can still be _configured in CLI_ or temporarily toggle Zygisk off. Why not allow it unconditionally? Please don't  just close PR without proper explanation


vvb2060:
Poka-yoke

PR #6657, opened 2023-03-01
Some minor cleanup

vvb2060:


osm0sis:
Shouldn't it be $NVBASE/modules or something to avoid hardcoding /data/adb?

vvb2060:
https://github.com/topjohnwu/Magisk/blob/cbbbbab4833e403dde53c14ac3f15b2e2ead5e75/docs/guides.md?plain=1#L16-L19

After booting, this is the only standard position. $NVBASE in script is for recovery env.

yujincheng08 (approved):

PR #6659, opened 2023-03-01
Refactor zygisk to use native bridge to inject

yujincheng08:
TODO
- [x] Test on more versions of Android with modules
- [x] Rollback after 3 times zygote crashes
- [x] resetprop after all zygote starts

close #6920

PR #6660, opened 2023-03-01
Add support for getting property context

canyie:
The planned `resetprop -Z` has been given up because it requires root privilege to run while `getprop` doesn't, and Android before 7.0 do not support property context.
But this is still useful for #6659

### DIFF
--- a/native/src/external/systemproperties/contexts_split.cpp
+++ b/native/src/external/systemproperties/contexts_split.cpp
@@ -344,6 +344,16 @@ prop_area* ContextsSplit::GetPropAreaForName(const char* name) {
   return cnode->pa();
 }
 
+const char* ContextsSplit::GetContextForName(const char* name) {
+  auto entry = ListFind(prefixes_, [name](PrefixNode* l) {
+      return l->prefix[0] == '*' || !strncmp(l->prefix, name, l->prefix_len);
+  });
+  if (!entry) {
+    return nullptr;
+  }
+  return entry->context->context();
+}
+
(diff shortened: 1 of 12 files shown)


native/src/resetprop/resetprop.cpp
```diff
@@ -118,6 +118,7 @@ static void read_prop(const prop_info *pi, void *cb) {
 struct sysprop_stub {
     virtual int setprop(const char *name, const char *value, bool trigger) { return 1; }
     virtual string getprop(const char *name, bool persist) { return string(); }
+    virtual const char* getcontext(const char *name) { return nullptr; }
```
yujincheng08, line 121: should return `std::string` like `getprop`

yujincheng08 (changes requested):

native/src/external/systemproperties/contexts_split.cpp
```diff
@@ -344,6 +344,16 @@ prop_area* ContextsSplit::GetPropAreaForName(const char* name) {
   return cnode->pa();
 }
 
+const char* ContextsSplit::GetContextForName(const char* name) {
```
vvb2060, line 347: Sync upstream, then use GetPrefixNodeForName
https://android-review.googlesource.com/c/platform/bionic/+/1417669

vvb2060 (changes requested):

native/src/resetprop/resetprop.cpp
canyie: I think using `std::string` anywhere is a very bad idea. I won't do it. Closing pull request.

PR #6666, opened 2023-03-02
Filter duplicate mount points for reverting

yujincheng08:
This prevents umounting existing overlay mount points.

close #6663


vvb2060:
If all targets are valid, can xumount2 be used now?

topjohnwu (approved):

PR #6672, opened 2023-03-03
Make magic mount read only

### LINKED ISSUE #3171: Systemless hosts may result in system freezes (F2FS driver bug)
## topjohnwu EDIT: 
This is caused by an F2FS driver issue that unfortunately Magisk is unable to fix. The only way around it is to use a custom kernel with the required fix. For more technical details, please check the reply by @freak07: https://github.com/topjohnwu/Magisk/issues/3171#issuecomment-752024237

## Original message:
Using current canary (MM299, Magisk 20425) enabling the systemless hosts feature results in freezing all of the system. Can only hard rebooted. Without hosts it works good so far.

vvb2060:
Only write to the safe `/data/adb/modules`, make the f2fs partition mounted with empty parameters read only, fix #3171

### DIFF
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -113,8 +113,8 @@ EOF
 
 add_hosts_module() {
   # Do not touch existing hosts module
-  [ -d $MAGISKTMP/modules/hosts ] && return
-  cd $MAGISKTMP/modules
+  [ -d $NVBASE/modules/hosts ] && return
+  cd $NVBASE/modules
   mkdir -p hosts/system/etc
   cat << EOF > hosts/module.prop
 id=hosts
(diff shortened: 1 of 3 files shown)


vvb2060:
Every user should assume that /system/etc/hosts is not writable, this will not cause confusion.  
Any modification should be written to /data/adb/modules/hosts/system/etc/hosts, which is what users expect.

vvb2060:
it's a bug

vvb2060:
> Any modification should be written to `/data/adb/modules/hosts/system/etc/hosts`

vvb2060:
In fact, any files overlay by Magisk modules can be written, not just hosts, it will bring accidents. Because normal apps and system tools believe that System, Vendor and other partitions are read only.
Therefore, I think it is not a deliberate behavior to write /system/etc/hosts. Apps should always modify module in /data/adb/modules.

vvb2060:
Anyone can bypass the permission protection of /data/adb and directly modify these files. If there is a module file context is `magisk_file`, this is even a security issue.

vvb2060:
Of course, this is a breaking change, so it will not appear in the upcoming stable release.

PR #6675, opened 2023-03-03
Make build reproducible

vvb2060:
- [x] Set stable random seed
- [ ] enable configuration cache


vvb2060:
Will re-submit the fixes when I have enough time.

PR #6676, opened 2023-03-03
Remove obsolete and long broken addon.d support

vvb2060:
https://github.com/yujincheng08/Magisk-Addon

https://github.com/programminghoch10/Lygisk/blob/ci-management/README.md?plain=1#L9-L11


osm0sis:
I am extremely disappointed/saddened that included functionality is being intentionally reduced.

As I've said previously, the goal should be making users' lives easier, and making them go hunt for a module to have their ROM updates remain seamless is definitely not that.

vvb2060:
> I have previously committed to continue to maintain the included addon.d support as I have for the years since its addition, and I stand by that.

https://github.com/programminghoch10/Lygisk/blob/ci-management/README.md?plain=1#L9-L11

osm0sis:
Sounds like a feature request?

It's not "long broken", that's a dramatic statement. It simply doesn't work in recovery if the recovery doesn't support decryption.

vvb2060:
This is more of an indication that the addon.d should be detached from magisk so it can update faster and ignore requests from topjohnwu.

pndwal:
> ... it doesn't even support FBE encrypted data partitions.

Sorry, isn't it the case that addon.d "fails to reinstall Magisk during OTA if the device does _not_ support FBE decryption"?... Seems to me therefore that it doesn't support old FDE... 🤔... What am I missing?

osm0sis:
FBE works fine on addon.d-v2, @pndwal is correct.

topjohnwu:
@osm0sis I'm not merging this change anytime soon, feel free to express your thoughts and provide more information to allow us to make a sound decision.

Howard20181:
@osm0sis Is the Recovery you are using supporting decrypted data?
Nowadays, most of the Recovery does not support decryption, such as the official LineageOS Recovery.

yujincheng08:
1. Many want to keep it because "it just works; why?". However, it has been long broken. Current implementations have not been working for a long time:
    - It requires the custom recovery to decrypt data partition. But nowadays, many recoveries need this ability. Even TWRP takes a long time to adapt to the new Android version. Magisk also does not recommend custom recoveries anymore (see below).
    - Install via patching boot does not support addon (and addon scripts cannot be updated via direct install on the Magisk app, see https://github.com/topjohnwu/Magisk/pull/3037#issuecomment-667522249). It only works when installed via custom recovery. However, Magisk has deprecated this way of installation for a long time. (Someone addresses [confusion about this]( https://github.com/topjohnwu/Magisk/issues/3820#issuecomment-1366226450)), so let's explain it briefly: The reason is that there are too many recoveries, and it's hard for Magisk to support them individually. They have caused too many troubles for us: we have to have a static build `magiskboot` because of broken `libc`, we have to find a way to get complete device information which can quickly get on a booted system, we can only write dirty scripts (and this is the most challenging parts to maintain on Magisk now, which also make Magisk app so complex) to support flashing on custom recoveries, low kernel version make the busybox unreliable, and so on).
    - Addon scripts are located in the system partition, which is against the "systemless" goal of Magisk, let alone it requires a writable system partition, which is generally becoming less feasible as most OEM uses read-only filesystems. Of course, we can overwrite the partition table and reformat the system as a writable filesystem, but the bar is too high for most people.

1. addon supports are mostly for custom ROMs and are not standardized by AOSP. It does not mean addon is a good design, even if most custom ROMs or the most popular custom ROM uses it. Backward compatibility prevents addon from moving to a better solution, e.g., placing scripts in the metadata partition instead.
1. A third-party module ([PoC](https://github.com/yujincheng08/Magisk-Addon)) is a better option for addon support.
    - It can upgrade freely without upgrading Magisk. Since Magisk has too many functions, it requires a longer test period than regular modules. This helps faster bug fixes and feature addition.
    - It works after upgrading Magisk via direct installation and can provide data-free support (like [this](https://github.com/programminghoch10/Lygisk)) that Magisk won't officially support.
    - Users can easily choose which version of addon (and other potential newer OTA survival implementations) to use instead of sticking to v2.
    - As for "making them go hunt for a module," I think that's the correct usage of Magisk. Otherwise, why Magisk supports modules rather than merging all modules' features?
1. As A/B devices are more prevalent nowadays, OTA survival support becomes easier by clicking install to another slot.
1. The deprecation announcement concern from https://github.com/topjohnwu/Magisk/issues/3820#issuecomment-1366226450. Magisk does not mention any addon support in its documentation (except in the changelog), so there's no place to place the announcement but the changelog. Whatever, it's a good sign of deprecation that this feature has long been broken without fixes (like #3820, nearly two years without addressing). Of course, once removed, we should well notify users about this.
1. Many people may also think, "even if it's not working for some people, why not just keep it for those who have it work?" Well, the same problem applies to the removed Magisk Hide feature. The existing official support may hinder third-party implementations and keeps people asking why it does not work. Unlike Magisk Hide, this feature is not opt-in, and people using ROMs without addon support cannot opt-out of this feature but accept some potential security issues caused by this (like [making update_engine permissive](https://github.com/topjohnwu/Magisk/blob/1aade8f8a8ddf2cdd207e2e1bdd304d3e5f8a4d3/native/src/sepolicy/rules.cpp#L193)).

osm0sis:
> @osm0sis Is the Recovery you are using supporting decrypted data? Nowadays, most of the Recovery does not support decryption, such as the official LineageOS Recovery.

addon.d-v2 is designed for the booted system, and it works fine as currently written (which is why no real changes for the past year).

addon.d(-v1) unencrypted works great in recovery, but FDE is a gap in coverage, as @topjohnwu had always been unhappy with the idea of keeping any Magisk binaries in multiple locations which he worried could get out of sync for various reasons.

Some further hack could be added to the current script to address this if John has come around on that idea. At one point I was discussing this with the Lineage developers and I seem to recall them talking design for addon.d-v3 as able to natively include support binaries with the addon.d scripts, so that could be something to look into as well.

onli:
(Please just delete if it's not okay to chime in. Just trying to help, as a follow developer).

In my eyes the problem here is deleting the existing approach without providing a known-working alternative. It was very annoying to run through the recommended installation and then see Magisk vanishing. It would completely make me drop Magisk if the only remaining alternative were dropped, as this PR seems to suggest - not only because it feels user hostile, but more because there simply would be no way for me to use Magisk anymore.

But this shows a way forward: Have both. Extend the existing installation documentation with the steps to make use of the PoC module to keep Magisk alive after upgrades, so user like me can test it. Undeprecate the recovery sideloading installation, and let @osm0sis keep the addon.d support alive as much as possible given the encryption hurdles.

I have a small testsuite here (work with Android in my day job) and could offer testing support, if it is wanted. I'm sure many others would provide similar support. Just let us know...

vvb2060:
> addon.d-v2 is designed for the booted system, and it works fine as currently written (which is why no real changes for the past 2 years).

No, it never worked properly, LineageOS needing permissive domains has not been fixed yet, they don't even want to modify their sepolicy and ask Magisk to do it.

> addon.d(-v1) unencrypted works great in recovery, but FDE is a gap in coverage, as topjohnwu had always been unhappy with the idea of keeping any Magisk binaries in multiple locations which he worried could get out of sync for various reasons.

You have admitted that non-A/B devices need unencrypted data partition or twrp that can decrypt data, and know that it cannot be improved to support encrypted data partition when it is bundled in Magisk, why still do not agree to let it break away from Magisk?

> At one point I was discussing this with the Lineage developers and I seem to recall them talking design for addon.d-v3 as able to natively include support binaries with the addon.d scripts, so that could be something to look into as well, given the same caveat.

It looks like the addon.d developers are aware of encrypted data partition, which is good.  But there's never been a discussion from them in this issue with Magisk, and I wonder why they haven't fixed the update engine needing a permissive domain.  Is it because Magisk actually misuses addon.d, which is not designed to modify boot partition?  When the design goal is to modify system partition, it makes sense to put the script in /system/addon.d, but Magisk is systemless, it has always conflicted with Magisk.

pndwal:
Again, for the sake of accuracy in this discussion only:

> You... know that it cannot be improved to support encrypted data partition when it is bundled in Magisk, why still do not agree to let it break away from Magisk?

Unless I'm missing something again, it seems addon.d(-v1) _does_ support encrypted data partition when it is bundled in Magisk, just not in devices using FDE... Am I wrong or did you actually mean "it cannot be improved to support _FDE_ encrypted data partition when it is bundled in Magisk"? 🤔

pndwal:
@yujincheng08, @vvb2060,

A number of good points made...

Just a thought:

If addon.d were moved to a module, could it still be kept as an _official_ Magisk component?...

I'm envisioning it being like the Systemless Hosts module, installable from Magisk settings page perhaps with a legend like:

**Magisk Survival for Custom ROM updates**
_Magisk Survival Script support for ROMs supporting addon.d_

This way
1) it remains available for Custom ROM users in Magisk out-of-the-box
2) the function is presented as a user option and use-case can be made clear in legend
3) it is only activated if user knows and intends to use it
4) it can receive both official and 3rd party support (via Magisk GitHub Issue reports and Pull requests)
5) potential for a plethora of additional (confusing/unwanted) forks is checked and proper/official support and PRs are encouraged  (I understand your view that it may be good to encourage 3rd party iterations, but this also has the potential to become very messy.)
6) it is _officially_ available whether Magisk is installed via custom ROM .zip installation or App patch/fastboot flash method.
7) an amicable compromise is reached AFAICS... addon.d Magisk script becomes a module but remains an official component, and this may leave all interested parties, ie. Magisk Owner/Devs, the script author(s), custom ROM users as well as 3rd party contributors, HAPPY! 😃

And let's face it, despite Magisk being very stock-ROM-centric these days (understandably if not well understood...), custom ROMs are the traditional Android modder's staple and it would not be a mistake at all for a custom mod of Magisk's calibre to _officially_ support custom ROM users in particular with a module like this... (Even Google offers special support for custom ROM users!) 😜

pixincreate:
> If addon.d were moved to a module, could it still be kept as an official Magisk component?...

Similar to `systemless hosts` module that can be installed through settings? Great!  
I would suggest keeping these in the module section itself in that case so that the things are categorized properly. 

Magisk, as far as I remember, had been developed while keeping the community in mind and I hope it continues to do the same.  

I read somewhere above @osm0sis said, "letting users hunt for modules". From a naive average users' perspective who doesn't do much tinkering their phones but rather install some phone customizing modules, this project already been went too far from reach. Unless the user digs in to see what's happening and search for modules they want on the internet, he/she might have already gave up. I personally took 5+ days to search for `AOSP Mods` and `xiaomi sdk` initially.  
I still want the native online module section to exist but that doesn't seem to have a revival and have to move on with that. If possible in the future, I'll make a simple static webpage where I'll host all the links to the projects that users wish to see which is open for all.  At least, by that we can try to narrow down hunting reasons.

osm0sis:
> I don't agree with `addon.d` removal too but I think `addon.d` feature should be implemented in another location (such as `/persist`, etc...) rather than `/system`. When I flash any Custom ROM, I always delete `/system/addon.d` anyway

I'll address the other replies when I have a bit more time, but addon.d is implemented by the ROM and is in /system because it supports all Android from before these other partitions existed.

I'll definitely concede that it would work better elsewhere, but currently that's up to the ROM. One pie in the sky idea would be to make a Zygisk module that hooks update_engine to add a new "magisk-addon.d" replacement that adds it to everywhere including stock ROM. 🤯

osm0sis:
> but `addon.d` is not zygote

No, but update_engine is.

Terminator-J:
> > @osm0sis Is the Recovery you are using supporting decrypted data? Nowadays, most of the Recovery does not support decryption, such as the official LineageOS Recovery.
> 
> addon.d-v2 is designed for the booted system, and it works fine as currently written (which is why no real changes for the past 2 years).

Yes, exactly this.
It *works right now* when using the booted Updater in crDroid (mostly lineage-based) on the oneplus 6-series that I maintain, which runs the scripts after updating. In fact, the way recent canary versions after 25.2 were working broke the ability to "apply to inactive slot" after the Updater process had begun, so literally the only way to allow for seamless updating (hard A/B, system-as-root, recovery-as-boot device) was to keep using the addon.d v2 implementation. And it has worked flawlessly.

(for certain values of "flawless" that include 500mb of accumulated magisk-backed-up recovery ramdisks in /data, but most people aren't flashing as often as I am as the ROM maintainer :D )

Nothing works from /system/addon.d when installing updates in Lineage-based recoveries on A/B devices like mine, but that's a known issue *with the recovery environment* and not a failure of Magisk. I don't know if it's a philosophical resistance by the Lineage team to enabling custom recoveries to be more powerful than AOSP intended, or a technical hurdle due to FBE; but that's not really important. Users know they need to manually re-flash their GApps zip and Magisk apk if they update "manually" in recovery on an A/B device like this.

Please don't break the booted Updater functionality just because of a known issue with the recovery updater that can't be solved anyway without the Lineage team changing their implementation.

osm0sis:
> > addon.d-v2 is designed for the booted system, and it works fine as currently written (which is why no real changes for the past 2 years).
> 
> No, it never worked properly, LineageOS needing permissive domains has not been fixed yet, they don't even want to modify their sepolicy and ask Magisk to do it.

Sorry, but you're mistaken here. LineageOS is actually the main ROM which ships update_engine domain permissive out-of-the-box themselves. Presumably they do this because addon.d is their baby and they know it requires a fair amount of access for GApps/add-on backup, restore and any other operations. This also worked to Magisk's advantage in allowing John and I to get it to do what it needed to do for Magisk addon.d-v2, so we agreed that it made sense to bring that permissive domain to all ROMs: https://github.com/topjohnwu/Magisk/commit/cba0d040000d6ff38c6646bb9d7daa12beae2878

Not sure if any of the other custom ROMs have improved things since then, but it would certainly be interesting to look into it further and discuss things with the major ROM project leads.

While I don't believe this to be a major concern or attack vector since only custom ROMs allow addon.d and therefore access to that domain, and a user would need to consciously install any addon.d script just as they do root (an obviously worse potential vector), but one solution could be to only make update_engine permissive when /system/addon.d is present.

> > addon.d(-v1) unencrypted works great in recovery, but FDE is a gap in coverage, as topjohnwu had always been unhappy with the idea of keeping any Magisk binaries in multiple locations which he worried could get out of sync for various reasons.
> 
> You have admitted that non-A/B devices need unencrypted data partition or twrp that can decrypt data, and know that it cannot be improved to support encrypted data partition when it is bundled in Magisk, why still do not agree to let it break away from Magisk?

This was a known limitation since the very beginning when John and I wrote it, so I don't think my "admission" is the smoking gun you believe it to be for unshipping it. Our concensus at the time was that FDE devices were on the way out (and they are, it's deprecated), and most older FDE devices which people still use tend to be unencrypted, so FBE was the important one and addon.d-v2 handles that well.

For those that do run FDE encrypted or are A-only and use a recovery that doesn't support decryption, we went to some lengths to output an error message for the user so that they were aware and could reflash Magisk manually: https://github.com/topjohnwu/Magisk/blob/master/scripts/addon.d.sh#L20-L40

Just to reiterate, if John has come around on it then I've certainly had some ideas over the years to either bundle the Magisk binaries with the existing addon.d-v2 script or update to (at least appear like) addon.d-v3 which I believe may allow for addon.d "support binaries" as I already mentioned...
 
> > At one point I was discussing this with the Lineage developers and I seem to recall them talking design for addon.d-v3 as able to natively include support binaries with the addon.d scripts, so that could be something to look into as well, given the same caveat.
> 
> It looks like the addon.d developers are aware of encrypted data partition, which is good. But there's never been a discussion from them in this issue with Magisk, and I wonder why they haven't fixed the update engine needing a permissive domain. Is it because Magisk actually misuses addon.d, which is not designed to modify boot partition? When the design goal is to modify system partition, it makes sense to put the script in /system/addon.d, but Magisk is systemless, it has always conflicted with Magisk.

I was speaking on behalf of Magisk at that time, since we all used to collaborate in the Slack together, when the Slack was still in use, if you recall. Again, Lineage shipped update_engine permissive first, Magisk just used that.

To be clear here, Magisk is root. Root can do whatever it wants, including write to system or boot if the user desires that. Your strong stance against system ever being written to is purely your opinion. If the ROM has been built to allow rw by not using erofs or ext4-dedup, with extra space in system, product and system_ext for GApps and other complex mods, maintained through OTAs with addon.d which is (to this point) intended permissive, then Magisk is there to help with that, and it's my opinion that Magisk should continue to use that existing infrastructure to make users' lives easier unless something better than addon.d comes along.

osm0sis:
> 1. Many want to keep it because "it just works; why?". However, it has been long broken. Current implementations have not been working for a long time:
>    
>    * It requires the custom recovery to decrypt data partition. But nowadays, many recoveries lack this ability. Even TWRP takes a long time to adapt to the new Android version. Magisk also does not recommend custom recoveries anymore (see below).

Only on FDE devices. And again, a known limitation since it was created, and not a grave concern because FDE itself is deprecated. I would love to see this support for older devices improved, but that's no reason to remove Magisk's addon.d script, since all newer devices are FBE and A/B, and addon.d-v2 support does work for them, as I explain in my previous post.

>    * Install via patching boot does not support addon (and addon scripts cannot be updated via direct install on the Magisk app, see [Copy required files to system for addon.d #3037 (comment)](https://github.com/topjohnwu/Magisk/pull/3037#issuecomment-667522249)). It only works when installed via custom recovery. However, Magisk has deprecated this way of installation for a long time. (Someone raises [confusion about this](https://github.com/topjohnwu/Magisk/issues/3820#issuecomment-1366226450), so let's explain it briefly: The reason is that there are too many recoveries, and it's hard for Magisk to support them individually. They have caused too many troubles for us: we have to have a static build `magiskboot` because of broken `libc`, we have to find a way to get complete device information which, on the other hand, can quickly get on a booted system, we can only write dirty scripts (and this is the most challenging parts to maintain on Magisk now, which also make Magisk app so complex) to support flashing on custom recoveries, low kernel version make the busybox unreliable, and so on).

Updating the addon.d scripts by Direct Install absolutely could be supported by the Magisk app, Magisk is root, you just don't agree that it should, and you're entitled to that opinion, but that doesn't invalidate the request for that support, nor the entire Magisk addon.d feature as it currently exists.

For clarity, there are only actually 2 recoveries right now, TWRP-based and Lineage-based, though I agree the different older Android trees which older TWRP devices are built on have always presented some problems.

While I agree there are some compromises to installing Magisk in recovery, especially recently for the serules/preinit refactor, it is also the only way to root your device before it ever boots, so should continue to exist in some form or another, even if that form is diminished to basically the same as unrooted boot patching. I fully support that.

>    * Addon scripts are located in the system partition, which is against the "systemless" goal of Magisk, let alone it requires a writable system partition, which is generally becoming less feasible as most OEM uses read-only filesystems. Of course, we can overwrite the partition table and reformat the system as a writable filesystem, but the bar is far too high for most people.

Nobody is suggesting Magisk force all system partitions to be writeable. The only topic at hand is custom ROMs which have already clearly, explicitly chosen to be rw by not using erofs or ext4-dedup, and even go so far as to include extra free space in system, product and system_ext for GApps and modifications. They indicate they support addon.d by including /system/addon.d and therefore remounting system rw is acceptable on that ROM. Magisk can be systemless overall but still support the custom ROMs which allow rw and addon.d. You have a difference of opinion on this, clearly, but it's just that, an opinion, and we are here to discuss, so please stop acting as though all rw is against Magisk; Magisk is root and root can do whatever it wants if we want it to.

> 2. addon supports are mostly for custom ROMs and are not standardized by AOSP. Even if most custom ROMs or the most popular custom ROM uses this machanism, it does not mean it is a good design. Backward compatibility prevents addon from moving to a better solution, e.g., placing scripts in the metadata partition instead.

addon.d is only for custom ROMs. CyanogenMod created it, and Lineage currently maintains it, releasing addon.d-v3 with Lineage 19.0 to support product, vendor and system_ext in addition to the original addon.d (and addon.d-v2) system support.

I agree keeping them elsewhere could be useful, but on a ROM where system is already allowed rw, there's no technical issue with the scripts still being in system.

For example if there was some Zygisk-backed replacement for addon.d which could hook update_engine to add something similar to addon.d support to any ROM including OEM, well then it would definitely need to use /metadata or something to support both FDE and FBE, but as you say, OEM ROMs don't allow system rw anymore, so that idea's use case would be pretty limited on OEM ROMs, e.g. to keeping Magisk and/or TWRP installed through an OTA.

> 3. A third-party module ([PoC](https://github.com/yujincheng08/Magisk-Addon)) is a better option for addon support.
>    
>    * It can upgrade freely without upgrading Magisk. Since Magisk has too many functions, it requires a longer test period than regular modules. This helps faster bug fixes and feature addition.
>    * It works after upgrading Magisk via direct installation and can provide data-free support (like [this](https://github.com/programminghoch10/Lygisk)) that Magisk won't officially support.
>    * Users can easily choose which version of addon (and other potential newer OTA survival implementations) to use instead of sticking to v2.
>    * As for "making them go hunt for a module," I think that's the correct usage of Magisk. Otherwise, why Magisk supports modules rather than merging all modules' features? Moreover, Google always helps.

I completely agree /data-free support in Magisk's addon.d would be helpful, and as I've mentioned a few times it could also be added to Magisk's existing script fairly easily, which I'd love to look into and wouldn't require much iteration since FDE is deprecated. It could work as a separate module, sure, but I'm not really seeing any actual reason to remove it from Magisk other than "rw is bad" presented here. Again, all of these limitations were known when John and I wrote it. Let's continue to support the custom ROM community by making it easy for them to stay rooted.

> 4. As A/B devices are more prevalent nowadays, OTA survival support becomes easier by clicking "install to another slot."
> 5. As for the deprecation announcement concern from [Nothing installed in Lineage /system/addon.d/ via Magisk app Direct Install + fix_env #3820 (comment)](https://github.com/topjohnwu/Magisk/issues/3820#issuecomment-1366226450). Magisk does not mention any addon support in its documentation (except in the changelog), so there's no place to place the announcement but in the changelog. Anyway, it's an obvious sign of deprecation that this feature has long been broken without fixes (like [Nothing installed in Lineage /system/addon.d/ via Magisk app Direct Install + fix_env #3820](https://github.com/topjohnwu/Magisk/issues/3820), nearly two years without fixes). Of course, once removed, we will well notify users about this.

Weird way to approach this... It seems like your argument here is "it was never really officially supported in the first place"? John and I wrote it. It is supported and works in every respect it was intended to work (i.e. excluding the known A-only FDE in-recovery limitation).

Then you seem to to be suggesting that my not contributing the few lines in fix_env and/or the Direct Install .kt to mount Magisk's own system mirror rw if the addon.d directory exists and push the script is somehow an indicator the entire feature support is dead?

First of all, that feels like some kind of attack on me, just like some of your previous abrupt comments in that issue thread about removing addon.d for "feature parity" do, but I won't rehash that further, I want to continue our debate here in good faith.

I have my 13 month old daughter sleeping in the next room to me, as evidence of the time constraints of having a home and family in addition to a full time job, but the real reason for no movement was that I couldn't bring myself to spend what little free time for Android I have on something that you and vvb had already so negatively indicated you weren't in favor of. The issue remained open to see if John might review it and change his mind to support improving the feature parity between booted and recovery by adding more support not less, at which point I'd happily do what I could, but, well, here we now are.

> 6. Many people may also think, "even if it's not working for some people, why not just keep it for those who have it work?" Well, the same problem applies to the removed Magisk Hide feature. The existing official support may hinder third-party implementations and keeps people asking why it does not work. Unlike Magisk Hide, this feature is not opt-in, and people using ROMs without addon support cannot opt-out of this feature but accept some potential security issues caused by this (like [making update_engine permissive](https://github.com/topjohnwu/Magisk/blob/1aade8f8a8ddf2cdd207e2e1bdd304d3e5f8a4d3/native/src/sepolicy/rules.cpp#L193)). Also, how about those people that use addon supported ROM but cannot have this feature work? At least in my thought, if we hand this feature out to third-party modules, those people could enough this feature. (AFAIK, some of them have gaven up addon, and some of them use third-party Magisk).

MagiskHide support was an ongoing battle with Google and banking apps, whereas addon.d support has remained very straightforward and unchanging. But I agree about the concerns for opt-in, so we could compromise as I previously suggested: never install the addon.d script by default through any method, but have a button in the Magisk app settings (a la Systemless Hosts) which is only visible on custom ROMs with /system/addon.d and will only install the addon.d script when tapped; an opt in only feature.

See my previous post where I discussed the update_engine permissive domain concern; I believe that could be worked around with some built-in logic as well if you still feel it's serious enough to require addressing, though John certainly agreed to it when we discussed it at the time.

At the end of the day this whole thing remains a difference of opinion between us about what Magisk as a root solution "should" and "shouldn't" do. You think it should never touch system under any circumstances, and I think it can do whatever it wants if the situation calls for it. We're likely not going to convince eachother, but we'll see what John thinks in reviewing our debate, and that will be that.

2shrestha22:
We need `addon.d`. Removing it makes harder and harder to use Magisk. Currently I am using LineageOS because it support `addon.d`.

andheartman:
@osm0sis I didn't know that the legendary developer co-wrote magisk with John. Chris doesn't talk for nothing. We support you. ❤️

osm0sis:
> @osm0sis I didn't know that the legendary developer co-wrote magisk with John. Chris doesn't talk for nothing. We support you. ❤️

Appreciated, but you're saying vast exaggerations here. 😘

I have only helped out in a few places with some scripts and things over the years. 🙂

vvb2060:
> Sorry, but you're mistaken here. LineageOS is actually the main ROM which ships update_engine domain permissive out-of-the-box themselves. Presumably they do this because addon.d is their baby and they know it requires a fair amount of access for GApps/add-on backup, restore and any other operations. This also worked to Magisk's advantage in allowing John and I to get it to do what it needed to do for Magisk addon.d-v2, so we agreed that it made sense to bring that permissive domain to all ROMs

I don't care how many bells and whistles there are in addon.d.sh, it doesn't bother me. I care that update_engine cannot be permissive. Only a few people use custom ROM, even fewer use addon.d, and they're pretty much all LineageOS, but you're bringing the permissive update_engine domain of LineageOS to all devices.

topjohnwu:
@osm0sis I'm totally fine with keeping addon.d scripts as it is. However, having `update_engine` as permissive is not something I'd love to keep in Magisk's code base. MagiskSU specifically allows `update_engine` to call `su`, and the script is also designed to run in a root shell. So for addon.d specifically, making that domain permissive is not necessary.

If custom ROMs require `update_engine` to be permissive to support other use cases, they should patch their own sepolicy on their side. It doesn't make sense to poke such a hole in SELinux on all Magisk users.

osm0sis:
> @osm0sis I'm totally fine with keeping addon.d scripts as it is. However, having `update_engine` as permissive is not something I'd love to keep in Magisk's code base. MagiskSU specifically allows `update_engine` to call `su`, and the script is also designed to re-exec and run in a root shell. So for addon.d specifically, making that domain permissive is not necessary.
> 
> If custom ROMs require `update_engine` to be permissive to support other use cases, they should patch their own sepolicy on their side. It doesn't make sense to poke such a hole in SELinux on all Magisk users.

Yep, that part is fine with me. I'd be happy to debug that change with the various ROM implementations of update_engine/addon.d-v2.

I'd still like to suggest the addon.d script becomes opt-in only as a button in the app settings, similar to a Systemless Hosts.

PR #6677, opened 2023-03-03
Fix `SYSTEM_ROOT` not passed to boot patch

yujincheng08:


canyie (changes requested):
Use `Info.isSAR` instead

app/src/main/java/com/topjohnwu/magisk/ui/install/InstallViewModel.kt
```diff
@@ -104,7 +104,7 @@ class InstallViewModel(
             Config.keepVerity,
             Config.keepEnc,
             Config.patchVbmeta,
-            Config.recovery
+            Config.recovery,
```
topjohnwu, line 107: No need to modify this file

yujincheng08: Done.

PR #6682, opened 2023-03-03
Remove support of partitions w/o symlink in /system

yujincheng08:
Since the current support of mounting more partitions does not satisfy people (see #3338), we remove the support for the next stable release to keep the semantics of `/system` unchanged.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -120,7 +120,7 @@ All files you want to replace/inject should be placed in this folder. This folde
 
 If you place a file named `.replace` in any of the folders, instead of merging its contents, that folder will directly replace the one in the real system. This can be very handy for swapping out an entire folder.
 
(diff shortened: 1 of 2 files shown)


vvb2060:
We should not change the semantics of the `system` folder, it will cause confusion. A new api should be designed to support other partitions, for example using new folder.

vvb2060 (approved):

osm0sis (approved):

Luffitys:
While on many devices `/odm` is only mounted in the root directly, on the OnePlus 12 it is symlinked in `/system/vendor/` as can be seen in following Termux output:

```
:/ # ls -l /system/vendor/ | grep odm
lrw-r--r--  1 root      root       4 2009-01-01 01:00 odm -> /odm
```

Hence it would make sense to support it along with the already supported symlinked partitions.

PR #6699, opened 2023-03-08
Fix shebang

s1204IT:
I've been wondering this for a long time.
No problem installing the module.

But in fact `/sbin/sh` doesn't exist.


yujincheng08:
This isn't a big deal. Magisk does not directly `exec` those shell scripts but runs them in a shell. Some scripts even start with `#!/usr/bin/env bash`.

s1204IT:
> This isn't a big deal. Magisk does not directly `exec` those shell scripts but runs them in a shell. Some scripts even start with `#!/usr/bin/env bash`.

Of course I understand that.
I'm simply curious.

s1204IT:
If going to declare shebang, I think might as well write what exists.

osm0sis:
Yeah, that's just a throwback to pre-Android 10 recoveries where /sbin/sh did exist, but Magisk hasn't supported recovery installation of modules for awhile, so it should probably be changed.

vvb2060:
I recommend removing shebang to avoid it being executed directly

osm0sis:
Agreed, good idea.

PR #6725, opened 2023-03-09
Fix app accessibility

### LINKED ISSUE #2125: Add support for Google Talkback assistance in Magisk Manager
Magisk is a really nice tool - but unfortunately blind users can't use it fully since some of the UI elements don't provide textual hints for Google Talkback assistance.

For example: in Magisk Hide there is a list with checkboxes next to the app names - these checkboxes will only be announced as "unnamed control" and without and indication if the checkbox is checked, unchecked or partially checked.

UI elements can provide a textual description for voice assistance in the android:contentDescription property (see https://github.com/arnowelzel/periodical/blob/master/app/src/main/res/layout/main.xml as an example) or by setting the description with `setContentDescription()` (see https://github.com/arnowelzel/periodical/blob/master/app/src/main/java/de/arnowelzel/android/periodical/CalendarCell.java as an example).

canyie:
This pull request is still in progress. Many many components need to be fixed. 
I'm busy recently so this won't be done any time soon

Close #2125

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/ui/MainActivity.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/ui/MainActivity.kt
@@ -120,20 +120,14 @@ class MainActivity : SplashActivity<ActivityMainMd2Binding>() {
         }
     }
 
-    override fun onOptionsItemSelected(item: MenuItem): Boolean {
-        when (item.itemId) {
-            android.R.id.home -> onBackPressed()
-            else -> return super.onOptionsItemSelected(item)
-        }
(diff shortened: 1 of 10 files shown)


Stuyve:
Can I use my Magisk now

DaUs3103:
[tsconfig.json](https://github.com/topjohnwu/Magisk/files/15302313/tsconfig.json)

DaUs3103:
[README.md](https://github.com/topjohnwu/Magisk/files/15302316/README.md)

canyie:
Will re-submit the fixes when I have enough time.

PR #6736, opened 2023-03-13
Support systemlessly deleting files or folders

yujincheng08:
After we refactor the magic mount and always mount folder as tmpfs, we can easily support deleting files or folders now. We recognize dummy devices with major number 0 and minor number 0 as an indicator for removing files and folders. This indicator is borrowed from overlayfs.


osm0sis:
Hopefully this one makes a return. Looked very cool. 🤞

aviraxp:
Can we get this in?

topjohnwu (approved):

SteveZMTstudios:
Finally!

PR #6745, opened 2023-03-14
Add zip and payload.bin support to Magisk app

yujincheng08:
Depends on #6438


yashikada:
#6438 is merged so also this PR can be merged, rigth?

yujincheng08:
Yes but too much conflict. I need to refine it.

PR #6746, opened 2023-03-14
Add Crowdin translation platform

yujincheng08:
https://magisk.crowdin.com/

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -1,0 +1,28 @@
+name: Crowdin Action
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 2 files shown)


osm0sis:
Good choice! Seems to work well for Lawnchair, plus they use GitHub Actions to commit the updated translations so that could be done here as well: https://github.com/LawnchairLauncher/lawnchair/tree/12.1-dev/.github/workflows

vvb2060:
I don't think it's good, I can trust PR translations to be reliable, but can't trust a translation platform where anyone can participate.  In fact lspsoed strings are often maliciously modified.

yujincheng08:
John agrees with @vvb2060.

PR #6748, opened 2023-03-14
Remove meaningless umount in magiskinit

yujincheng08:
This is no longer required since we redirect to /data/magiskinit


canyie (approved):
Checked the source code of 25.2 and still not sure the original reason for this unmount, but I agreed that it is meaningless now. 👍

PR #6750, opened 2023-03-16
Specify context, user and group for init.rc's exec

yujincheng08:
Note that in Android 5.0 there's no `exec` support thus we don't have to care about the `exec` doc of Android 5.0.

https://cs.android.com/android/platform/superproject/+/android-5.1.1_r38:system/core/init/builtins.c;drc=d26135b34d4ee9d983931d64c8cb18dd2562b938;l=257


yujincheng08:
https://github.com/topjohnwu/Magisk/commit/af054e4e31fe250ccb9d98dbe4b16d4fe9835c7e

PR #6753, opened 2023-03-17
Use app_dir to detect systemui instead

yujincheng08:
uid may be shared by other apps (e.g., in MIUI, systemui's uid is 1000 and shared by many system apps).

Note that we don't use process names because any app can name its process `com.android.systemui`.


aviraxp:
In pixels, it is called com.google.android.systemui

yujincheng08:
The previous logic is to stat `com.android.systemui` so it does not matter.

topjohnwu (approved):

PR #6773, opened 2023-03-22
Update README.MD

Howard20181:
People complain that the documentation doesn't make it clear how to bypass signature verification

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -57,6 +57,11 @@ For Magisk app crashes, record and upload the logcat when the crash occurs.
 
 - The certificate of the key used to sign the final Magisk APK product is also directly embedded into some executables. In release builds, Magisk's root daemon will enforce this certificate check and reject and forcefully uninstall any non-matching Magisk apps to protect users from malicious and unverified Magisk APKs.
 - To do any development on Magisk itself, switch to an **official debug build and reinstall Magisk** to bypass the signature check.
+  For how to do this:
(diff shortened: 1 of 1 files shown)


README.MD
```diff
@@ -57,6 +57,13 @@ For Magisk app crashes, record and upload the logcat when the crash occurs.
 
 - The certificate of the key used to sign the final Magisk APK product is also directly embedded into some executables. In release builds, Magisk's root daemon will enforce this certificate check and reject and forcefully uninstall any non-matching Magisk apps to protect users from malicious and unverified Magisk APKs.
 - To do any development on Magisk itself, switch to an **official debug build and reinstall Magisk** to bypass the signature check.
+
+    For how to do this:
+
+    1. Change the channel to custom `https://github.com/topjohnwu/magisk-files/raw/master/debug.json`
```
yujincheng08, line 63: Don't need to set the channel. Installing the debug version is enough.

Howard20181: Should I point to the debug APK raw link?
If the update channel is not changed then those who randomize the package names need to restore to the default.

pndwal: Not sure if restoring is needed, but I think we should  refer to 'Magisk App' not 'manager' (line 64) since Manager has been 'dead' since 22.0 (last Manager app version being ver 8.0.7) and this may cause confusion. 🤪
https://topjohnwu.github.io/Magisk/releases/22000.html#magisk-manager-is-deadlong-live-the-magisk-app

vvb2060:
Just install the debug version directly…

PR #6782, opened 2023-03-23
New way to unload zygisk

yujincheng08:
Tested platform:

Android 6 x86_64
Android 6 x86
Android 14 DP2 x86_64
Android 13 arm64
Android 13 arm


canyie (changes requested):
And the new way needs to be tested with more architectures and more Android versions

canyie (approved):

PR #6823, opened 2023-04-05
Allows the su to switch to the MNT_NS of the pid

CCint3:
Allows the su command to switch to the mount namespace of the target process.

The usage of the new feature is as follows:

```shell
su --ns-mnt 1 -c /system/bin/sh
```

This command will create a new `sh` process, and switch the namespace of the new process to the `1(init)` process.


yujincheng08:
nsenter -t x

CCint3:
> nsenter -t x

oh, thanks. i got it!!

CCint3:
@yujincheng08  magisk 有讨论组吗? 例如 tg

PR #6866, opened 2023-04-10
Temp workaround for module file context

yujincheng08:


freak07:
This fix works for me. Previously I´d see a selinux denial during bootup reading the powerhal config file when replacing the powerhal config file via magisk module on pixel 6. The selinux denial ultimately causes the phone to freeze and crash. 

That commit solves the crash on my end.

topjohnwu (approved):

Issue #5959, opened 2022-06-10
Bootloop after OTA


capntrips:
Device: Pixel 6 Pro
Android version: 12.1 (SQ3A.220605.009.B1)
Magisk version name: cacf8736
Magisk version code: 25001

The standard OTA process from the [guide](https://topjohnwu.github.io/Magisk/ota) was followed, but it bootlooped. Here's the log:

[console-ramoops-0.log](https://github.com/topjohnwu/Magisk/files/8876167/console-ramoops-0.log)

The relevant lines seem to be:

```
[    1.257611][    T1] init: [libfs_mgr]Created logical partition system_b on device /dev/block/dm-0
...
[    1.263952][    T1] init: [libfs_avb]Returning avb_handle with status: Success
[    1.264042][    T1] init: [libfs_avb]Built verity table: '1 /dev/block/dm-0 /dev/block/dm-0 4096 4096 216517 216517 sha256 b5abedbc7a21e42b338ea3977800ccc43a5cd0e3a682afeee52271a552ca7c20 1a5edec5bd4392e3181b5f908b2a446b31116d8e8010660bd5035b16c539a51f 10 use_fec_from_device /dev/block/dm-0 fec_roots 2 fec_blocks 218224 fec_start 218224 restart_on_corruption ignore_zero_blocks'
[    1.264207][    T1] device-mapper: verity: sha256 using implementation "sha256-ce"
[    1.264717][    T1] init: [libfs_mgr]Set readahead_kb: 128 on /sys/class/block/dm-10/queue/read_ahead_kb
[    1.264754][    T1] init: [libfs_mgr]Set readahead_kb: 128 on /sys/class/block/dm-0/queue/read_ahead_kb
[    1.264803][    T1] init: [libfs_mgr]Set readahead_kb: 128 on /sys/class/block/sda26/../queue/read_ahead_kb
[    1.266138][  T265] device-mapper: verity-fec: 254:0: FEC: recursion too deep
[    1.266142][  T265] device-mapper: verity: 254:0: metadata block 216517 is corrupted
[    1.266158][  T265] debug-reboot: Create reboot monitor timer now
[    1.269090][  T265] Reboot command: 'dm-verity device corrupted'
```

Note this is not from a device I have access to. It has happened every month. This is on the stock ROM with the stock kernel. Booting from the patched image after installing the latest build with the flash tool works without issue.

[...]

PR #6884, opened 2023-04-12
Update ota.md: Add note for Pixel 6/7 users OTA

n0k0m3:
Add note for Pixel 6/7 users OTA to reflect current status of https://github.com/topjohnwu/Magisk/issues/5959#issuecomment-1152868270

### DIFF
--- a/docs/ota.md
+++ b/docs/ota.md
@@ -14,6 +14,9 @@ Magisk does not modify most read-only partitions, which means applying official 
 
 ### Devices with A/B Partitions
 
+**Notes:** For the moment `Install to Inactive Slot` method won't work for all Pixel 6 and 7 models (https://github.com/topjohnwu/Magisk/issues/5959). Either follow ["Non A/B" Devices](#non-ab-devices) guide below or use a 3rd party tool like [PixelFlasher](https://github.com/badabing2005/PixelFlasher)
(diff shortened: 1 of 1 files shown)


docs/ota.md
```diff
@@ -14,6 +14,9 @@ Magisk does not modify most read-only partitions, which means applying official
 
 ### Devices with A/B Partitions
 
+**Notes:** For the moment `Install to Inactive Slot` method won't work for all Pixel 6 and 7 models (https://github.com/topjohnwu/Magisk/issues/5959). Either follow ["Non A/B" Devices](#non-ab-devices) guide below or use a 3rd party tool like [PixelFlasher](https://github.com/badabing2005/PixelFlasher)
```
SammCheese, line 17: Avoid linking 3rd-party-tools. Even if they are popular/safe atm, it can change at any point.

n0k0m3: Does saying "use other 3rd party pixel rooting/flashing tool" without explicitly mentioning PixelFlasher or name of other tool work?

SammCheese: Aye

vvb2060:
We don't know the definitive compatibility list, and why some devices are available.  So let it be as it is.

PR #6906, opened 2023-04-16
Update/refine Spanish translations

naxitoo:
It's not perfect but I tried to:

- Fix misspellings
- Make strings look not too formal
- Make strings sound more natural/friendly
- Well, make everything look a bit better in general

Anyway, please make sure I didn't mess up any variable or escape character before merging lol (although I double-checked before PRing)


Displax:
No need to translate commented pointers.

topjohnwu (approved):

PR #6917, opened 2023-04-20
Retain PREINITDEVICE in recovery flash if previously found during booted patching

osm0sis:
Should make for fewer cases where the extra reflash step is needed 🙂

### DIFF
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -83,8 +83,8 @@ Supported actions:
         Create ramdisk backups from ORIG
       restore
         Restore ramdisk from ramdisk backup stored within incpio
-      sha1
-        Print stock boot SHA1 if previously backed up in ramdisk
+      config VARIABLE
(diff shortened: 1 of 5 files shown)


vvb2060:
`magiskboot cpio file sha1` -> `magiskboot cpio file config SHA1`
so, we can use `magiskboot cpio file config PREINITDEVICE`

osm0sis:
Is that an undocumented feature? I only see `magiskboot cpio <file> sha1` and I thought about adapting that for a `preinit` option, but it seemed silly to add 25 lines of code where a simple one-liner in shell would suffice.

osm0sis:
Anything I'm missing @vvb2060 , @topjohnwu ?

topjohnwu:
@osm0sis thanks for you change, I'm removing the ability to extract vars with `magiskboot` and move the logic to shell scripts instead :)

PR #6920, opened 2023-04-22
Fix executing app_process failed if magiskd restarted

canyie:
This bug causes app_process fails to start after magiskd restarts, which can cause Magisk app fails to open, executing `su` always timeout and throws permission denied error, or bootloops if zygotes are started after magiskd restarts


vvb2060:
The only way to exit magiskd is `magisk --stop`, restart magiskd in other cases is undefined behavior

canyie:
9fbd079

PR #6932, opened 2023-04-26
resetprop: Support get property context

vvb2060:
close #6660


native/src/resetprop/resetprop.cpp
```diff
@@ -211,6 +214,15 @@ struct resetprop : public sysprop {
         return val;
     }
 
+    string getcontext(const char *name) override {
+        string val;
+        if (!check_legal_property_name(name))
+            return val;
+        val = __system_property_get_context(name);
```
canyie, line 221: If `__system_property_get_context` returns nullptr, undefined behavior occurs

vvb2060: done

topjohnwu:
Merged in f36b21bae55d13317b126ddf1489719870739801

PR #6933, opened 2023-04-26
Fix meizu rootfs type

vvb2060:


native/src/init/twostage.cpp
```diff
@@ -40,7 +40,7 @@ bool SecondStageInit::prepare() {
     argv[0] = (char *) INIT_PATH;
 
     // Some weird devices like meizu, uses 2SI but still have legacy rootfs
-    if (struct statfs sfs{}; statfs("/", &sfs) == 0 && sfs.f_type == RAMFS_MAGIC) {
+    if (struct statfs sfs{}; statfs("/", &sfs) == 0 && sfs.f_type == TMPFS_MAGIC) {
```
yujincheng08, line 43: I think we can keep both.

yujincheng08 (changes requested):

yujincheng08 (approved):

PR #6998, opened 2023-05-16
Support settings gids of su

yujincheng08:


vvb2060:
https://man7.org/linux/man-pages/man1/su.1.html
```
       -g, --group=group
           Specify the primary group. This option is available to the
           root user only.

       -G, --supp-group=group
           Specify a supplementary group. This option is available to
           the root user only. The first specified supplementary group
           is also used as a primary group if the option --group is not
           specified.
```

vvb2060:
https://github.com/topjohnwu/Magisk/blob/fe1ca52f6daa9b7ae22981666f5de4406d4ae3ec/native/src/su/su.cpp#L130-L132

This feature should be reimplemented as there is no permission to set context after setuid

vvb2060:
add groups info to app log?

yujincheng08:
> there is no permission to set context after setuid

```
emu64a:/ $ su 1000
emu64a:/ $ id
uid=1000(system) gid=1000(system) groups=1000(system) context=u:r:magisk:s0
emu64a:/ $ runcon u:r:shell:s0 sh
emu64a:/ $ id
uid=1000(system) gid=1000(system) groups=1000(system) context=u:r:shell:s0
```

PR #7019, opened 2023-05-26
Update app deps

pashamcr:


vvb2060 (changes requested):
https://developer.android.com/build/migrate-to-ksp

Note: If a module has any kapt processors remaining, stubs are still generated in that module. This means that the majority of performance improvement will occur only when all usages of kapt are removed from the module.

app/build.gradle.kts
```diff
@@ -108,14 +109,14 @@ dependencies {
     implementation("androidx.navigation:navigation-fragment-ktx:${vNav}")
     implementation("androidx.navigation:navigation-ui-ktx:${vNav}")
 
-    implementation("androidx.biometric:biometric:1.1.0")
+    implementation("androidx.biometric:biometric-ktx:1.2.0-alpha05")
```
vvb2060, line 112: why use alpha version?

topjohnwu:
Handled in c79bc8327534f98b2c6deced39e0a1029be6159e

PR #7022, opened 2023-05-26
Add INTERNALDIR

Zackptg5:
In commit 4e2b88b, '.magisk' was removed from MAGISKTMP and INTERNALDIR was added in the documentation of the replacement for what it was. However, INTERNALDIR isn't defined in magisk manager and so module expecting it will fail.
Either it needs added like in this pull or removed from the documentation

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/utils/ShellInit.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/utils/ShellInit.kt
@@ -42,6 +42,7 @@ class ShellInit : Shell.Initializer() {
 
             if (shell.isRoot) {
                 add("export MAGISKTMP=\$(magisk --path)")
+                add("export INTERNALDIR=\$MAGISKTMP/.magisk")
                 // Test if we can properly execute stuff in /data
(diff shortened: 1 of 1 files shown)


Zackptg5:
> You should not depend on this internal behavior. Why not `MAGISKTMP=$(magisk --path)/.magisk` if you want?

True. MAGISKTMP is a variable used throughout magisk/magisk manager for the path all the time which is why I've used it.
Main issue being addressed with this pull is the discrepancy in the documentation. The variable is listed there as defined when it in fact isn't.

vvb2060:
Public Variables
https://topjohnwu.github.io/Magisk/guides.html#variables

Internal Details
https://topjohnwu.github.io/Magisk/details.html

Zackptg5:
> Public Variables
> https://topjohnwu.github.io/Magisk/guides.html#variables
> 
> Internal Details
> https://topjohnwu.github.io/Magisk/details.html
> 
> "Internal" means, you should not use.

That's fine, details page should be updated then to say that and that variables listed may not actually exist. I'm sure others will ask the same questions down the road

PR #7047, opened 2023-06-05
Clean up unrooted boot patching log errors

osm0sis:
Clean up unrooted boot patching cp errors in log

- remove call since cp_readlink shouldn't be necessary with installDir getting cleaned up regularly

    cp: can't preserve ownership of 'busybox': Operation not permitted
    cp: can't preserve ownership of 'magisk32': Operation not permitted
    cp: can't preserve ownership of 'magisk64': Operation not permitted
    cp: can't preserve ownership of 'magiskboot': Operation not permitted
    cp: can't preserve ownership of 'magiskinit': Operation not permitted
    cp: can't preserve ownership of 'magiskpolicy': Operation not permitted

See: [magisk_install_log_2023-06-04T20.03.33.log](https://github.com/topjohnwu/Magisk/files/11647630/magisk_install_log_2023-06-04T20.03.33.log)

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -456,10 +456,7 @@ abstract class MagiskInstallImpl protected constructor(
             return false
         }
 
-        // Fix up binaries
         srcBoot.delete()
-        "cp_readlink $installDir".sh()
-
         return true
     }
 


scripts/boot_patch.sh
```diff
@@ -173,7 +173,7 @@ if [ -n "$PREINITDEVICE" ]; then
   echo "PREINITDEVICE=$PREINITDEVICE" >> config
 fi
 [ -n "$SHA1" ] && echo "SHA1=$SHA1" >> config
-RANDOMSEED=$(tr -dc 'a-f0-9' < /dev/urandom | head -c 16)
+RANDOMSEED=$(tr -dc 'a-f0-9' < /dev/urandom 2>/dev/null | head -c 16)
```
vvb2060, line 176: RANDOMSEED is no longer used and can be deleted

scripts/util_functions.sh
```diff
@@ -318,10 +318,12 @@ loop_setup() {
 
 mount_apex() {
   $BOOTMODE || [ ! -d /system/apex ] && return
-  local APEX DEST
+  local APEX APEXJARS DEST FWK PATTERN SHCON
   setup_mntpoint /apex
   mount -t tmpfs tmpfs /apex -o mode=755
-  local PATTERN='s/.*"name":[^"]*"\([^"]*\).*/\1/p'
+  SHCON=$(cat /proc/self/attr/current)
+  echo "u:r:su:s0" > /proc/self/attr/current # work around LOS Recovery not allowing loop mounts in recovery context
```
vvb2060, line 325: u:r:su:s0 may not exist

scripts/util_functions.sh
```diff
@@ -469,7 +472,7 @@ install_magisk() {
   cd $MAGISKBIN
 
   if [ ! -c $BOOTIMAGE ]; then
-    eval $BOOTSIGNER -verify < $BOOTIMAGE && BOOTSIGNED=true
+    [ -f /system/bin/dalvikvm ] && eval $BOOTSIGNER -verify < $BOOTIMAGE && BOOTSIGNED=true
```
vvb2060, line 475: AVB 1.0 is only required for very old devices, and they are unlikely to have LOS Recovery. I recommend skipping mount apex and exec dalvikvm for new devices

app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
```diff
@@ -456,10 +456,7 @@ abstract class MagiskInstallImpl protected constructor(
             return false
         }
 
-        // Fix up binaries
         srcBoot.delete()
-        "cp_readlink $installDir".sh()
```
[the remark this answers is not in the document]
osm0sis: Doesn't make sense though, it produces the errors in the commit message and the app is continually deleting installDir to start fresh anyway.

scripts/boot_patch.sh
osm0sis: Ahh, you're right! This is because of debug_ramdisk being used now?

scripts/util_functions.sh
osm0sis: It does. All recoveries use recovery for flashing zips and su for adb shell, which is why I had trouble diagnosing the issue: loop mounts always worked in adb shell but never in the flashing zip.

scripts/util_functions.sh
osm0sis: That would be another option yeah

osm0sis: Many old devices have Lineage Recovery however.

vvb2060: oh, I thought it came after the A/B device. So, is there any other way to skip for avb 2.0 devices?

scripts/boot_patch.sh
vvb2060: Yes, magiskinit doesn't need random now.

scripts/util_functions.sh
vvb2060: https://review.lineageos.org/#/c/257053

aviraxp: I think we make recovery permissive 4 years ago?

osm0sis: All I know is shell with recovery context fails to do loop mounts, but change it to su context and it works. You can see the errors in the commit and log.

scripts/util_functions.sh
osm0sis: Yeah there's definitely A-only Lineage Recovery too for some devices. Could we somehow grab from magiskboot when it's AVB 2.0 and skip based on that?

scripts/util_functions.sh
vvb2060: userdebug_or_eng(`
permissive recovery;
')

recovery context is always permissive. Is there an avc log?

osm0sis: I'll pull one. It's like this on any current Lineage device.

app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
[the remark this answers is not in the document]
aviraxp: Just curious, does toybox's mount work?

scripts/util_functions.sh
osm0sis: I'm thinking I could grab the AVB status using `tee` if you think that could be a good way to go about it.

scripts/util_functions.sh
osm0sis: Here is a fresh recovery.log of 26102 and the dmesg.log from the same session. Some parts say permissive=1 and some say permissive=0, which is interesting and may explain what I've been seeing and the workaround I found.

[recovery.log](https://github.com/topjohnwu/Magisk/files/11749729/recovery.log)
[dmesg.log](https://github.com/topjohnwu/Magisk/files/11749730/dmesg.log)

vvb2060:
The reason for adding avb 1.0 signature is that some devices will read the boot signature info even after unlock bootloader and cause boot fail. I don't think these old devices can be upgraded to Android 10. Given the complexity of recovery env, we should do as little as possible. I think we can skip avb 1.0 detection for devices with apex exists.

aviraxp:
After contacting with lineage staff internally, mounting apex is unsupported usecase. We can handle it here imo.

osm0sis:
> The reason for adding avb 1.0 signature is that some devices will read the boot signature info even after unlock bootloader and cause boot fail. I don't think these old devices can be upgraded to Android 10. Given the complexity of recovery env, we should do as little as possible. I think we can skip avb 1.0 detection for devices with apex exists.

I think the reason apex mount support was added at all was because there are devices relying on avb 1.0 working as you say, that have been updated to Android 10, at least via custom ROM, so apex was needed to run /system/bin/dalvikvm to sign via recovery.

Now, I agree this is a losing battle however, since linkerconfig and namespaces have made it so dalvikvm can't be run from recovery anymore in Android 11, and dalvikvm moving to /apex/com.android.art/bin in Android 13 has made attempting it even more complicated.

scripts/util_functions.sh
osm0sis: I guess this might explain why Open GApps does `setenforce 0` as part of their installer! I'd never noticed it until now:

https://github.com/opengapps/opengapps/commit/a0832927c596153b2549a600061fd8ffeb67dc79

osm0sis: Logs are clean of the mount errors in Permissive of course. So yeah, can either switch context like I suggested, or full SELinux Permissive for recovery installs. Whichever seems more appealing to @vvb2060, I suppose.

[recovery.log](https://github.com/topjohnwu/Magisk/files/11777244/recovery.log)
[dmesg.log](https://github.com/topjohnwu/Magisk/files/11777245/dmesg.log)

osm0sis:
Oh wait, I just had a brainwave. How about we just look for /system/bin/dalvikvm, and then only ever even attempt apex mounts when it exists? Seems simple and would check all the boxes for all of our concerns.

scripts/util_functions.sh
osm0sis: Consider this resolved then since my workaround is viable? Or would you prefer full `setenforce 0` for recovery installs?

scripts/util_functions.sh
osm0sis: Added skipping apex mounts entirely when /system/bin/dalvikvm isn't present nicely excludes all the latest Android versions and devices, so I think we can consider this resolved?

yujincheng08 (changes requested):
No need to mount apex now.

https://github.com/topjohnwu/Magisk/commit/af65d0745648456c80d0682e273a97dfa1fe2f56

osm0sis:
I see the new built-in avb 1.0 signing adds a new error/warning to the logs:

```
[boot/sign.rs:241] unexpected ASN.1 DER tag: expected SEQUENCE, got APPLICATION [1] (primitive)
```

But I don't think I can figure out how to fix that one.

PR #7048, opened 2023-06-05
Don't patch devices that use dynamic partitions

### LINKED ISSUE #6786: Bootloop upon fastboot installation with FP3-8901.4.A.0021.0-gms-0b368514-user
Device: Fairphone 3+ (FP3)
Android version: 11
Magisk version name: 981ccabb (D)
Magisk version code: 25210

FP3-8901.4.A.0021.0-gms-0b368514-user has introduced retrofitted Dynamic Partitions to the fairphone 3(+) and when preforming the update via fastboot boot.img patching results in a bootloop of the device until a counter is hit and the device is booted into fastboot.
Checking Fairphone's forurms preforming the update via ota from FP3-8901.4.A.0019.2-gms-53c17a08-user works as expected but not patching the boot.img as per the linked thread: https://forum.fairphone.com/t/magisk-fairphone-os/93689/6     
I have attached the boot.img files and the output of cat /proc/mounts made on FP3-8901.4.A.0021.0-gms-0b368514-user.

[Output of `adb shell cat /proc/mounts`]
[proc_mounts.txt](https://github.com/topjohnwu/Magisk/files/11068769/proc_mounts.txt)
[Clean original boot.img] 
[FP3-8901.4.A.0021.0-gms-0b368514-user_orignal_boot.img.zip](https://github.com/topjohnwu/Magisk/files/11068695/FP3-8901.4.A.0021.0-gms-0b368514-user_orignal_boot.img.zip)
[Patched with magisk manager 981ccabb (25210) (D)
[FP3-8901.4.A.0021.0-gms-0b368514-user_magisk_patched-25210_boot.img.zip](https://github.com/topjohnwu/Magisk/files/11068692/FP3-8901.4.A.0021.0-gms-0b368514-user_magisk_patched-25210_boot.img.zip)

chaosmaster:
Devices that use dynamic partitions always have a ramdisk, while they should not be using skip_initramfs at all and use androidboot.force_normal_boot=1 instead, some vendors decided to patch their kernel instead of updating their bootloader when retrofitting dynamic partitions.

See also:
https://source.android.com/docs/core/ota/dynamic_partitions/implement#system-as-root-changes https://gerrit-public.fairphone.software/plugins/gitiles/kernel/msm-4.9/+/refs/heads/rel/11/fp3/8901.4.A.0022/fs/proc/cmdline.c

Fixes #6786

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -6,7 +6,8 @@
 # Usage: boot_patch.sh <bootimage>
 #
 # The following flags can be set in environment variables:
-# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT
+# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT,
+# DYNAMIC_PARTITIONS
 #
 # This script should be placed in a directory with the following files:
 #
@@ -74,6 +75,8 @@ fi
 [ -z $PATCHVBMETAFLAG ] && PATCHVBMETAFLAG=false
 [ -z $RECOVERYMODE ] && RECOVERYMODE=false
 [ -z $SYSTEM_ROOT ] && SYSTEM_ROOT=false
+[ -z $DYNAMIC_PARTITIONS ] && DYNAMIC_PARTITIONS=false
+
 export KEEPVERITY
 export KEEPFORCEENCRYPT
 export PATCHVBMETAFLAG
@@ -231,12 +234,17 @@ if [ -f kernel ]; then
   # After:  [mov w2, #-32768]
   ./magiskboot hexpatch kernel 821B8012 E2FF8F12 && PATCHEDKERNEL=true
 
-  # Force kernel to load rootfs for legacy SAR devices
-  # skip_initramfs -> want_initramfs
-  $SYSTEM_ROOT && ./magiskboot hexpatch kernel \
-  736B69705F696E697472616D667300 \
-  77616E745F696E697472616D667300 \
-  && PATCHEDKERNEL=true
+  # Don't patch on devices that use dynamic partitions, as these devices always
+  # use a ramdisk and patching their kernel may break booting on devices that
+  # did not correctly implement androidboot.force_normal_boot.
+  if [ "$DYNAMIC_PARTITIONS" = "false" ]; then
+    # Force kernel to load rootfs for legacy SAR devices
+    # skip_initramfs -> want_initramfs
+    $SYSTEM_ROOT && ./magiskboot hexpatch kernel \
+    736B69705F696E697472616D667300 \
+    77616E745F696E697472616D667300 \
+    && PATCHEDKERNEL=true
+  fi
 
   # If the kernel doesn't need to be patched at all,
   # keep raw kernel to avoid bootloops on some weird devices
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -380,7 +380,7 @@ umount_apex() {
 
 # After calling this method, the following variables will be set:
 # KEEPVERITY, KEEPFORCEENCRYPT, RECOVERYMODE, PATCHVBMETAFLAG,
-# ISENCRYPTED, VBMETAEXIST
+# ISENCRYPTED, VBMETAEXIST, DYNAMIC_PARTITIONS
 get_flags() {
   getvar KEEPVERITY
   getvar KEEPFORCEENCRYPT
@@ -418,6 +418,14 @@ get_flags() {
     fi
   fi
   [ -z $RECOVERYMODE ] && RECOVERYMODE=false
+  if [ -z $DYNAMIC_PARTITIONS ]; then
+    DYNAMIC_PARTITIONS=$(getprop ro.boot.dynamic_partitions)
+    if [ "$DYNAMIC_PARTITIONS" = "true" ]; then
+      ui_print "- Dynamic partitions, don't force ramdisk"
+    else
+      DYNAMIC_PARTITIONS=false
+    fi
+  fi
 }
 
 find_boot_image() {
--- a/app/src/main/java/com/topjohnwu/magisk/core/Info.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/Info.kt
@@ -33,6 +33,7 @@ object Info {
     @JvmStatic val isFDE get() = crypto == "block"
     @JvmField var ramdisk = false
     @JvmField var vbmeta = false
+    @JvmField var dynamicPartitions = false
     var crypto = ""
     var noDataExec = false
     var isRooted = false
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -493,6 +493,7 @@ abstract class MagiskInstallImpl protected constructor(
             "KEEPVERITY=${Config.keepVerity} " +
             "PATCHVBMETAFLAG=${Config.patchVbmeta} " +
             "RECOVERYMODE=${Config.recovery} " +
+            "DYNAMIC_PARTITIONS=${Info.dynamicPartitions} " +
             "SYSTEM_ROOT=${Info.isSAR} " +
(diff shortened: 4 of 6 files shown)


scripts/boot_patch.sh
```diff
@@ -231,12 +231,18 @@ if [ -f kernel ]; then
...
-  # Force kernel to load rootfs for legacy SAR devices
-  # skip_initramfs -> want_initramfs
-  $SYSTEM_ROOT && ./magiskboot hexpatch kernel \
-  736B69705F696E697472616D667300 \
-  77616E745F696E697472616D667300 \
-  && PATCHEDKERNEL=true
+  # Don't patch on devices that use dynamic partitions, as these devices always
+  # use a ramdisk and patching their kernel may break booting on devices that
+  # did not correctly implement androidboot.force_normal_boot.
+  DYNAMIC_PARTITIONS=$(grep_get_prop ro.boot.dynamic_partitions /default.prop /vendor/build.prop)
```
yujincheng08, line 237: This should be set as an environment variable as `KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT`.

yujincheng08 (changes requested):

scripts/boot_patch.sh
```diff
@@ -74,6 +75,10 @@ fi
 [ -z $PATCHVBMETAFLAG ] && PATCHVBMETAFLAG=false
 [ -z $RECOVERYMODE ] && RECOVERYMODE=false
 [ -z $SYSTEM_ROOT ] && SYSTEM_ROOT=false
+
+[ -z $DYNAMIC_PARTITIONS ] && DYNAMIC_PARTITIONS=$(grep_get_prop ro.boot.dynamic_partitions /default.prop /vendor/build.prop)
```
yujincheng08, line 79: pass this in magisk app like others.

yujincheng08 (changes requested):

scripts/boot_patch.sh
chaosmaster: You mean via `export $DYNAMIC_PARTITIOINS` ? since that isn't done here for `$SYSTEM_ROOT` either?
Or can you point me to what you mean?

yujincheng08: https://github.com/topjohnwu/Magisk/blob/5e2ef1b7f41aa695d425db152add83629cc93363/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt#L492-L496

chaosmaster: Wouldn't boot_patch.sh then only work when called by Magisk app or there'd need to be two places to be maintained for checking dynamic partitions?

yujincheng08: keep consistency with other variables

yujincheng08:
/default.prop and /vendor/build.prop are not accessible without root.

chaosmaster:
> /default.prop and /vendor/build.prop are not accessible without root.

They were meant for root-case such as in recovery `grep_get_prop` falls back to `getprop` if the grep fails.

yujincheng08:
so getprop works for both cases. why not just use getprop?

chaosmaster:
@yujincheng08 is there anything else needed from my side?

scripts/boot_patch.sh
```diff
@@ -6,7 +6,8 @@
 # Usage: boot_patch.sh <bootimage>
 #
 # The following flags can be set in environment variables:
-# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT
+# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT,
+# DYNAMIC_PARTITIONS
```
norpol, line 10: There are other lines also long, might want to make this comment in a single line for easier grepping dunno.

```suggestion
# KEEPVERITY, KEEPFORCEENCRYPT, PATCHVBMETAFLAG, RECOVERYMODE, SYSTEM_ROOT, DYNAMIC_PARTITIONS
```

scripts/boot_patch.sh
```diff
@@ -231,12 +234,17 @@ if [ -f kernel ]; then
...
 
-  # Force kernel to load rootfs for legacy SAR devices
-  # skip_initramfs -> want_initramfs
-  $SYSTEM_ROOT && ./magiskboot hexpatch kernel \
-  736B69705F696E697472616D667300 \
-  77616E745F696E697472616D667300 \
-  && PATCHEDKERNEL=true
+  # Don't patch on devices that use dynamic partitions, as these devices always
+  # use a ramdisk and patching their kernel may break booting on devices that
+  # did not correctly implement androidboot.force_normal_boot.
```
norpol, line 239: I've noticed that your `boot.img` still contains the `force_normal_boot` as a `/proc/cmdline` parameter. Should this be also removed if `getprop ro.boot.dynamic_partitions`  is true?

screenshot of diff showing vendor vs patched containing `force_normal_boot`

norpol (commented):
Looks good to me, I think what's missing is updating the `docs/changes.md` and adding documentation similar to `PATCHVBMETAFLAG` in `docs/tools.md`. (Disclaimer not a contributor to this repo)

scripts/boot_patch.sh
chaosmaster: It's customary to try keeping lines below 80 characters, I don't think this is an issue

scripts/boot_patch.sh
chaosmaster: I believe you're looking at the wrong file, this PR neither adds nor removes anything from the cmdline and it shouldn't.

(You may be looking at https://github.com/topjohnwu/Magisk/files/11636973/FP3-8901.4.A.0022.0-boot_magisk_patched_modified.img.zip, which was my first (manual) workaround for this issue, that's not what is created by this PR. If you are using this file, you should be using https://github.com/topjohnwu/Magisk/files/11646684/FP3-8901.4.A.0022.0-boot_magisk_patched_skip_initramfs.img.zip instead. As I've explained in https://github.com/topjohnwu/Magisk/issues/6786#issuecomment-1573963698 the former breaks booting into recovery.)

norpol: Oh, sorry for the confusion then! Thanks for clarifying so quickly.

scripts/util_functions.sh
```diff
@@ -418,6 +418,14 @@ get_flags() {
     fi
   fi
   [ -z $RECOVERYMODE ] && RECOVERYMODE=false
+  if [ -z $DYNAMIC_PARTITIONS ]; then
+    DYNAMIC_PARTITIONS=$(getprop ro.boot.dynamic_partitions)
```
vvb2060, line 422: Don't getprop in twrp, because twrp may be made based on an old system(before retrofitted dynamic partitions).

vvb2060 (changes requested):

scripts/util_functions.sh
chaosmaster: I previously used `$(grep_get_prop ro.boot.dynamic_partitions /default.prop /vendor/build.prop)` however this wil probably not work either if TWRP was built without dynamic partitions, as it won't be able to mount the partitions.

What do you suggest to do instead?

yujincheng08: #7147

vvb2060: Unfortunately, we don't have any means. The retrofitted dynamic partition device does not have a super block, and the twrp built on old system cannot mount system partition and has no ro.boot.dynamic_partitions prop. Fairphone's bootloader does not pass the correct parameter, it depends on the kernel, but twrp may use an old kernel.

chaosmaster: I have been working on updating FP3 TWRP to support ROMS with and without dynamic partitions.
There I am checking for "androidboot.super_partition" in the boot.img that is installed in the current slot.
Maybe that would be an option?

Something along these lines:

```
BOOTDEVICE=/dev/block/bootdevice/by-name/boot$(getprop ro.boot.slot_suffix)
dd if="$BOOTDEVICE" bs=2048 count=1 | grep -q androidboot.super_partition
if [ $? -ne 0 ]; then
    echo "ROM does not use dynamic partitions"
else
    echo "ROM uses dynamic partitions"
fi
```

chaosmaster:
#7215 achieves the same result without relying on getprop

PR #6274, opened 2022-09-19
Update resources load

vvb2060:
addAssetPath supports apk and directory

PR #7121, opened 2023-06-26
Fix stub resource loading on Android 9, 10

yujincheng08:
They can only load resources from zip files

### DIFF
--- a/stub/src/main/java/com/topjohnwu/magisk/DownloadActivity.java
+++ b/stub/src/main/java/com/topjohnwu/magisk/DownloadActivity.java
@@ -35,7 +35,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.zip.InflaterInputStream;
+import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
 
 import javax.crypto.Cipher;
 import javax.crypto.CipherInputStream;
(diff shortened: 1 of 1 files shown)


canyie:
#6274

vvb2060:
Directory support was broken in Android 9 and 10.

https://cs.android.com/android/_/android/platform/frameworks/base/+/c07aa702703388747bd6e9b1091127e2736ffcd8

Add ResourcesProvider.loadFromDirectory

This API allows a directory to be loaded as if it was a zipped APK.
This is a substitute for the DirectoryAssetProvider API that currently does not work in the native layer.

stub/src/main/java/com/topjohnwu/magisk/DownloadActivity.java
```diff
@@ -183,19 +185,17 @@ private void loadResources() throws Exception {
...
-
-            // addAssetPath requires a directory containing AndroidManifest.xml on Android 6.0
-            try (var stubApk = new ZipFile(getPackageCodePath());
-                 var manifest = new FileOutputStream(new File(dir, "AndroidManifest.xml"))) {
-                var stubManifest = stubApk.getInputStream(
-                        stubApk.getEntry("AndroidManifest.xml"));
-                APKInstall.transfer(stubManifest, manifest);
+            File res = new File(getCodeCacheDir(), "res.apk");
+            try (var out = new ZipOutputStream(new FileOutputStream(res))) {
+                // addAssetPath requires a directory containing AndroidManifest.xml on Android 6.0
```
canyie, line 190:
```suggestion
                // addAssetPath requires a zip containing AndroidManifest.xml on older Android versions
```

vvb2060, line 190:
```suggestion
                // Android 6 requires AndroidManifest.xml, Android 9-10 directory support is broken
```

canyie (approved):

PR #7122, opened 2023-06-26
Add samloader kotlin to the list

MrDisguised:
Samloader and samfirm are no longer being maintained. Samloader kotlin (also known as bifrost) is a cross platform app to download samsung firmwares which works on mac,windows,linux and android. It is a great alternative to something like frija and is actively being maintained.

Issue #6230, opened 2022-08-30
Bootloop after ODIN flash (update)


RV7PR:
Device: Samsung Galaxy A53 (SM-A536B)
Android version: 12
Magisk version name: 38325e7
Magisk version code: 25203

[logs.zip](https://github.com/topjohnwu/Magisk/files/9451630/logs.zip)

![error](https://user-images.githubusercontent.com/17615589/187401961-72521d9c-ccde-4c6f-ae48-e7caa00ec2fd.jpg)

Exynos-nibba:
make sure you patch the boot image in magisk manager through A53 itself. i.e install magisk manager on A53, patch the boot image and flash it. also what region is your phone from?

RV7PR:
> make sure you patch the boot image in magisk manager through A53 itself. i.e install magisk manager on A53, patch the boot image and flash it. also what region is your phone from?

I have for the past 3 years with 2 different phones, never a problem. I use a usb flash drive to transfer de patched tar file and then flash with odin, i always check if file is corrupted. I'm using EUX, its the european union version

nemesis03:
I probably have the same problem with my Samsung Galaxy S22. With the current firmware (S901BXXU2AVH9 EUX, security patch level 01.09.2022) I get a bootloop after flashing the patched firmware. I patched the firmware like I always do (transfer AP file via ADB patch on my S22, pull the patched file via ADB and then flash the complete firmware with the patched file instead of AP). I tried with magisk 25200 and 25203.
Flashing the old patched firmware (S901BXXU2AVG6, security patch level 01.08.2022) works and the phone boots normally

Igetin:
In case it helps, I'm also using a Galaxy S22 from the same region, and I did a full **initial install** of Magisk with the same latest firmware (S901BXXU2AVH9 EUX), and did **not** experience any boot loops. So, if what you're exeriencing _is_ a bug or incompatibility issue with Magisk (and not an incorrect flashing procedure from the user), the problem is likely only related to upgrading the OS for an existing Magisk install.

One thing that came to my mind: the Odin release version might affect the results, so could you tell which version did you use, @nemesis03? I did the initial install with Odin v3.14.1 (original/unpatched).

Exynos-nibba:
> > make sure you patch the boot image in magisk manager through A53 itself. i.e install magisk manager on A53, patch the boot image and flash it. also what region is your phone from?
> 
[...]

PR #7130, opened 2023-06-29
Refactor Samsung devices install guide

BlackMesa123:
close #7122

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -117,6 +117,9 @@ Having KnoxGuard active will prevent you from installing/running Magisk regardle
 - Allow bootloader unlocking in **Developer options → OEM unlocking**
 - Reboot to download mode: power off your device and press the download mode key combo for your device
 - Long press volume up to unlock the bootloader. **This will wipe your data and automatically reboot.**
+- Go through the initial setup. Skip through all the steps since data will be wiped again in later steps. **Connect the device to Internet during the setup.**
+- Enable developer options, and **confirm that the OEM unlocking option exists and is grayed out.** This means KnoxGuard hasn't locked your device.
+- Your bootloader now accepts unofficial images in download mode
 
 ### Instructions
 
@@ -144,9 +147,8 @@ Once you have rooted your Samsung device, you can no longer upgrade your Android
 
 ### Important Notes
 
-- **Never, ever** try to restore either `boot`, `recovery`, or `vbmeta` partitions back to stock! You can brick your device by doing so, and the only way to recover from this is to do a full Odin restore with data wipe.
+- **Never, ever** try to restore either `boot`, `init_boot`, `recovery`, or `vbmeta` partitions back to stock! You can brick your device by doing so, and the only way to recover from this is to do a full Odin restore with data wipe.
 - To upgrade your device with a new firmware, **NEVER** directly use the stock `AP` tar file with reasons mentioned above. **Always** patch `AP` in the Magisk app and use that instead.
-- Never just flash only `AP`, or else Odin may shrink your `/data` filesystem size. Flash `AP` + `BL` + `CP` + `HOME_CSC` when upgrading.
 
 ## Custom Recovery
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -20,7 +20,7 @@ The result of **Ramdisk** determines whether your device has ramdisk in the boot
 
 > _(Unfortunately, there are exceptions as some devices' bootloader accepts ramdisk even if it shouldn't. In this case, you will have to follow the instructions as if your device's boot partition **does** include ramdisk. There is no way to detect this, so the only way to know for sure is to actually try. Fortunately, as far as we know, only some Xiaomi devices are known to have this property, so most people can simply ignore this piece of information.)_
 
-If you are using a Samsung device that is launched with Android 9.0 or higher, you can now jump to [its own section](#samsung-system-as-root).
+If you are using a Samsung device, you can now jump to [its own section](#samsung-devices).
 
 If your device has boot ramdisk, get a copy of the `boot.img` (or `init_boot.img` if exists).<br>
 If your device does **NOT** have boot ramdisk, get a copy of the `recovery.img`.<br>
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -80,30 +80,51 @@ As a summary, after installing Magisk in recovery **(starting from power off)**:
...
-- Your bootloader now accepts unofficial images in download mode
 
 ### Instructions
 
-- Use either [samfirm.js](https://github.com/jesec/samfirm.js), [Frija](https://forum.xda-developers.com/s10-plus/how-to/tool-frija-samsung-firmware-downloader-t3910594), or [Samloader](https://forum.xda-developers.com/s10-plus/how-to/tool-samloader-samfirm-frija-replacement-t4105929) to download the latest firmware zip of your device directly from Samsung servers.
+- Download the latest firmware package for your device, you can use one of the tools below to download it directly from Samsung servers:
+  - [SamFirm.NET](https://github.com/jesec/SamFirm.NET), [samfirm.js](https://github.com/jesec/samfirm.js)
+  - [Frija](https://forum.xda-developers.com/s10-plus/how-to/tool-frija-samsung-firmware-downloader-t3910594)
+  - [Samloader](https://forum.xda-developers.com/s10-plus/how-to/tool-samloader-samfirm-frija-replacement-t4105929)
+  - [Bifrost](https://forum.xda-developers.com/t/tool-samsung-samsung-firmware-downloader.4240719/)
```
yujincheng08, line 127:
```suggestion
  - [Bifrost](https://forum.xda-developers.com/t/tool-samsung-samsung-firmware-downloader.4240719/)
  - [Samloader kotlin](https://github.com/zacharee/SamloaderKotlin)
```

BlackMesa123: Samloader Kotlin is already there, it's actually called "Bifrost":

https://github.com/zacharee/SamloaderKotlin/blob/64fc960a3c7d0eb491bd465db564a7eeebed5b2a/README.md?plain=1#L6

But I can replace the XDA thread link with the GitHub repo page if you want.

pndwal:
Not sure if everything for Samsung SAR+ should now be under a general "Samsung" heading...

For example, are the extra wipes prescribed required for devices with launch versions less than Android 9? - I believe Vaultkeeper was only introduced with S10 series and A9...

If devices with launch versions less than A9 are upgrading to A10+, they are converted from "legacy SAR" to modern "2SI ramdosk SAR" however...

BlackMesa123:
> Not sure if everything for Samsung SAR+ should now be under a general "Samsung" heading...

There's nothing specific to SAR+ devices in the Samsung guide, it can be used on legacy devices as well since the steps are exactly the same, tested myself on my old Galaxy S8+ running Android 9.

> For example, are the extra wipes prescribed required for devices with launch versions less than Android 9?

The two-steps bootloader unlock (RPMB) has been introduced since the S10 series with AVB 2.0 (iirc), on older devices you only have to enable "OEM unlocking" in Developer Options. However formatting the data partition is required the first time Magisk is installed if the device does have data encryption in place (both FDE or FBE) as I explained [here](https://github.com/topjohnwu/Magisk/issues/6230#issuecomment-1429639687).

> I believe Vaultkeeper was only introduced with S10 series and A9...

KnoxGuard/RMM has been a thing in Samsung devices [since 2018](https://forum.xda-developers.com/t/how-to-root-after-january-security-patch-sm-a520f-a5-2017.3739225/page-2#post-75360965). The only thing you need to account is that the device is either not completely locked or temporarely locked, since the bootloader will refuse to load a custom kernel regardless of the bootloader lock state (you'll get the `Custom binary blocked due to remaining installment balance` error). There are various guides around the web (eg. [XDA](https://forum.xda-developers.com/t/how-to-bypass-kg-state-and-or-rmm-state-prenormal.3891193/page-4#post-85552805)) that claim you can actually bypass the 168h timer in the device and restore KG State from `Prenormal` to `Completed`.

pndwal:
Does line 121 mean:
**confirm that the OEM unlocking option exists, _is enabled_ and is grayed out.** ... Just a guess. 🙃

BlackMesa123:
> Does line 121 mean: **confirm that the OEM unlocking option exists, _is enabled_ and is grayed out.** ... Just a guess. 🙃

I did copy back the original line I deleted in the first commit, so it's the same as before:
https://github.com/topjohnwu/Magisk/blob/4856da1584c16f5c7387efcefe255c21a1978233/docs/install.md?plain=1#L101

Didn't apply any modification to it since it is correct. The OEM Unlock toggle will not be visible if these conditions aren't respected:
```java
@Override
public boolean isAvailable() {
    return showEnableOemUnlockPreference(this.mContext) && // Device supports OEM unlocking
    	!KnoxGuardManager.getInstance().shouldBlockCustomRom() && // KG is not Prenormal, Active or Locked
    	mOemLockManager != null; // Device is not LDU or has eng firmware
}
```

and will be grayed out if the bootloader has been unlocked in Download Mode:
```java
@Override
public void updateState(Preference preference) {
    // ...
    // KG is Prenormal, Active or Locked
    // The toggle won't be visible anyway because this check is also in isAvailable()
    if (KnoxGuardManager.getInstance().shouldBlockCustomRom()) {
        mPreference.setEnabled(false);
    } else {
        mPreference.setEnabled(enableOemUnlockPreference()); // false if bootloader is unlocked
    }
    // ...
}
```

pndwal:
Ah, that's clear, thanks... I don't have Sammy device to check but assume toggle does show enabled though greyed out...

Still concerned about the wording if instructions are not just for 9+ Launch Version (SAR+) Samsung however... 

"Enable developer options, and **confirm that the OEM unlocking option exists and is grayed out.** This means the VaultKeeper service has unleashed the bootloader" would have applied to all SAR devices, and you do now refer to KnoxGuard which covers Android 7+ LV devices if I understand you comments linked above correctly... 👍

Guess 9+ devices with VaultKeeper retain KnoxGuard too, and toggle behaviour is the same then...

I'm wondering if older devices get VaultKeeper or KnoxGuard with updates... Suspect it's LV dependant.

Seems a note to cover old devices running A6 (some devices will only have stock ROMs up to A6 available) will be needed as KnoxGuard reference won't apply to these and wipes may be superfluous as well as caveats about unrooting etc... You'll know if older devices updated to A7+ are affected by KnoxGuard etc, or prone to bricking if images are restored to stock.

Also, I know older devices had AP binary, but not all 4 binaries mentioned for SAR devices... 🙃

BlackMesa123:
> Ah, that's clear, thanks... I don't have Sammy device to check but assume toggle does show enabled though greyed out...

The fact the toggle is enabled while being grayed out is taken for granted since as I said in https://github.com/topjohnwu/Magisk/pull/7130#issuecomment-1626897036 the toggle is grayed out **when** the bootloader is fully unlocked (it can't be disabled back if you first do not lock your bootloader again via Download mode).

> Still concerned about the wording if instructions are not just for 9+ Launch Version (SAR+) Samsung however...
> 
> "Enable developer options, and **confirm that the OEM unlocking option exists and is grayed out.** This means the VaultKeeper service has unleashed the bootloader" would have applied to all SAR devices, and you do now refer to KnoxGuard which covers Android 7+ LV devices if I understand you comments linked above correctly... +1
> 
> Guess 9+ devices with VaultKeeper retain KnoxGuard too, and toggle behaviour is the same then...
> 
> I'm wondering if older devices get VaultKeeper or KnoxGuard with updates... Suspect it's LV dependant.

The wording in the current guide are wrong, which is what I've fixed with my PR (other than adding how to identify if your device is locked). **VaultKeeper** is actually a security system service used to read/write data in the device's RPMB, which is then used by other system services such as KnoxGuard. **KnoxGuard** is the actual security feature that prevents flashing unsigned firmware images, and it also does more than that, you can check [Samsung Knox documentation](https://docs.samsungknox.com/admin/knox-guard/welcome.htm) for more info. Older versions of this feature are named **RMM (Remote Mobile Manager)**, the name is different but it's the exact same. Again, this has nothing to do with the device having SAR, but with the current running Android and Knox version the device is running on.

> Seems a note to cover old devices running A6 (some devices will only have stock ROMs up to A6 available) will be needed as KnoxGuard reference won't apply to these and wipes may be superfluous as well as caveats about unrooting etc... You'll know if older devices updated to A7+ are affected by KnoxGuard etc, or prone to bricking if images are restored to stock.

The main difference is that legacy devices might not have the "two-steps bootloader unlock" (devices shipped with Android 8 and below) or KG/RMM.

> Also, I know older devices had AP binary, but not all 4 binaries mentioned for SAR devices... upside_down_face

Not really an issue since legacy devices with one tar file can be patched without any issue.

AByakugans:
I am a little bit lost here. What is the correct way to root for example Samsung A51 which was already rooted before and I need to keep the data? (No wipe data) when Install/update official Samsung Firmware with Odin.

1) Do I just need to patch AP with Magisk or do I need additional files like to put vbmeta_disabled file in USERDATA slot so AVB remains disabled and my data preserved (from https://forum.xda-developers.com/t/how-to-exynos-snapdragon-root-s20-series-and-upgrade-firmware.4079353/ )
2) Do I need to use Canary or standard Magisk?

topjohnwu (approved):

PR #7133, opened 2023-07-01
Add `boot.img` in Android 13-shipped Samsung tar packages

### LINKED ISSUE #7132: Patching AP file deletes stock boot.mg from tar on Samsung S23
Device:Samsung S23 Ultra
Android version:13
Magisk version name:26.1
Magisk version code:26103

Latest Magisk while Patching AP files, successfully patching init_boot.img and root works fine but it deletes stock boot.img file from AP tar
This cause when someone update firmware by just flashing Magisk patched AP (without updating stock firmware), kernel doesn't get update or may cause bootloop if old kernel doesn't work on updated firmware.
Attached screenshot of Stock AP file & Magisk Patched AP file

no_boot_img
Patched AP

stock_AP
Stock AP

BlackMesa123:
#6586 causes the `processTar` function to provide `init_boot.img` instead of `boot.img` to support Samsung devices shipping with Android 13 and above. Add the stock `boot.img` in the patched tar generated by the app to avoid issues caused by mismatched kernel images when updating to a newer firmware

Fixes #7132

### DIFF
--- a/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -230,32 +230,37 @@ abstract class MagiskInstallImpl protected constructor(
         val boot = installDir.getChildFile("boot.img")
         val initBoot = installDir.getChildFile("init_boot.img")
         val recovery = installDir.getChildFile("recovery.img")
-        if (Config.recovery && recovery.exists() && boot.exists()) {
+        if (Config.recovery && recovery.exists()) {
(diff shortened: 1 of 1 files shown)


BlackMesa123:
Same as https://github.com/topjohnwu/Magisk/pull/6586#issuecomment-1419263895, the kernel image in `boot.img` still doesn't gets the Samsung specific hex patches in [boot_patch.sh](https://github.com/topjohnwu/Magisk/blob/43b9a09c9b5915924ca297e75df0773b39ad610a/scripts/boot_patch.sh#L221), tho nobody have reported any issues caused by one of the various Samsung's LSM.

yujincheng08 (approved):

PR #7147, opened 2023-07-09
Force kernel to load rootfs only for legacy SAR devices

vvb2060:
close #7048


scripts/util_functions.sh
```diff
@@ -418,6 +418,14 @@ get_flags() {
     fi
   fi
   [ -z $RECOVERYMODE ] && RECOVERYMODE=false
+  local CMDLINE_SAR_FLAG=false
+  grep -q 'skip_initramfs' /proc/cmdline && CMDLINE_SAR_FLAG=true
```
chaosmaster, line 422: In recovery boot `skip_initramfs` would not be present on cmdline

PR #7172, opened 2023-07-20
Fix $BOOTMODE set by manager.sh being incorrectly overridden

canyie:
`manager.sh` sets `$BOOTMODE` to true, but `util_functions.sh` unconditionally sets it to empty. 16d728f
This regression causes the following bugs:
- `ui_print` prints message to `/proc/self/fd` instead of the manager console if not rooted
- `boot_patch.sh` thinks we are in recovery mode and preinit device resolving being incorrectly skipped
- maybe more?

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -9,7 +9,7 @@
 ###################
 
 # True if the script is running on booted Android, not something like recovery
-BOOTMODE=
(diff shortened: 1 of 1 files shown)


osm0sis:
Maybe just comment all the "documentation" lines so it's back to how it was before?

topjohnwu:
I ended up taking @osm0sis's suggestion :). Thanks all!

Issue #7176, opened 2023-07-25
magiskpolicy allowxperms overwrites old xperm bits


chenxiaolong:
Device: Pixel 7 Pro (`cheetah`)
Android version: 13
Magisk version name: Debug
Magisk version code: 26104

---

When using `magiskpolicy` with an `allowxperm` rule, it seems that old xperm bits are being overwritten. For example, when running:

```bash
magiskpolicy \
    --load /sdcard/policy.old \
    --save /sdcard/policy.new.1_xperm \
    'type test_app' \
    'allowxperm test_app test_app { rawip_socket tcp_socket udp_socket } ioctl 0x5401'

magiskpolicy \
    --load /sdcard/policy.old \
    --save /sdcard/policy.new.2_xperm \
    'type test_app' \
    'allowxperm test_app test_app { rawip_socket tcp_socket udp_socket } ioctl 0x5401' \
    'allowxperm test_app test_app { rawip_socket tcp_socket udp_socket } ioctl 0x5414'
```

The resulting files looks like this:

```bash
❯ sesearch --allowxperm -s test_app -t test_app -ds -dt policy.new.1_xperm
allowxperm test_app test_app:rawip_socket ioctl 0x5401;
allowxperm test_app test_app:tcp_socket ioctl 0x5401;
allowxperm test_app test_app:udp_socket ioctl 0x5401;

❯ sesearch --allowxperm -s test_app -t test_app -ds -dt policy.new.2_xperm
allowxperm test_app test_app:rawip_socket ioctl 0x5414;
allowxperm test_app test_app:tcp_socket ioctl 0x5414;
allowxperm test_app test_app:udp_socket ioctl 0x5414;
```

The second statement's 0x5414 xperm bits overwrote the 0x5401 bits.

I'm not super familiar with magiskpolicy's xperm logic yet, but I suspect the issue might be here:

https://github.com/topjohnwu/Magisk/blob/6141bb5bb322c0d801abafccb3fb9426990b0c31/native/src/sepolicy/sepolicy.cpp#L296

`sepol_impl::add_xperm_rule` memsets `datum->xperms`, but the function never reads the existing value from it for merging.

Should this be reading the original value from the avtab entry if it exists?

https://github.com/topjohnwu/Magisk/blob/6141bb5bb322c0d801abafccb3fb9426990b0c31/native/src/sepolicy/sepolicy.cpp#L266

[...]

PR #7177, opened 2023-07-25
magiskpolicy: Fix old xperms being cleared when adding new xperms

chenxiaolong:
This commit updates `sepol_impl::add_xperm_rule()` so that it loads the current xperm bits from the existing avtab entry before setting or clearing xperm bits. This fixes new `allowxperm` rules causing old xperm rules within the same xperm specified/driver to be removed.

Fixes: #7176


topjohnwu (approved):

chenxiaolong:
Thanks for the quick review and merge!

PR #7215, opened 2023-08-07
Force kernel to load rootfs only for legacy SAR devices

vvb2060:
https://cs.android.com/android/platform/superproject/main/+/main:build/make/core/config.mk;l=1098;drc=6196841f4d47433f9ecc353c4cf75d6871f96082


vvb2060:
#7048 cc @chaosmaster

chaosmaster:
@vvb2060 I can confirm this works both from the App and from TWRP (built with dynamic partition support) on Android 12.
Have not tested on Android 13 yet, but I assume it will work the same.

topjohnwu (approved):

PR #7220, opened 2023-08-08
manager.sh: add ro.boot.vbmeta.size + ro.product.ab_ota_partitions to vbmeta check

osm0sis:
- ro.boot.vbmeta.device doesn't seem to be in use on all devices, so add some alternative props which may be instead

### DIFF
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -203,7 +203,11 @@ get_flags() {
   [ "$(getprop ro.crypto.state)" = "encrypted" ] && ISENCRYPTED=true
   KEEPFORCEENCRYPT=$ISENCRYPTED
   VBMETAEXIST=false
-  [ -n "$(getprop ro.boot.vbmeta.device)" -o -n "$(getprop ro.boot.vbmeta.size)" ] || getprop ro.product.ab_ota_partitions | grep -wq vbmeta && VBMETAEXIST=true
+  if [ -n "$(getprop ro.boot.vbmeta.device)" -o -n "$(getprop ro.boot.vbmeta.size)" ]; then
+    VBMETAEXIST=true
+  elif getprop ro.product.ab_ota_partitions | grep -wq vbmeta; then
+    VBMETAEXIST=true
+  fi
   # Preset PATCHVBMETAFLAG to false in the non-root case
   PATCHVBMETAFLAG=false
   # Make sure RECOVERYMODE has value
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -203,7 +203,7 @@ get_flags() {
   [ "$(getprop ro.crypto.state)" = "encrypted" ] && ISENCRYPTED=true
   KEEPFORCEENCRYPT=$ISENCRYPTED
   VBMETAEXIST=false
-  [ -n "$(getprop ro.boot.vbmeta.device)" ] || getprop ro.product.ab_ota_partitions | grep -wq vbmeta && VBMETAEXIST=true
+  [ -n "$(getprop ro.boot.vbmeta.device)" -o -n "$(getprop ro.boot.vbmeta.size)" ] || getprop ro.product.ab_ota_partitions | grep -wq vbmeta && VBMETAEXIST=true
   # Preset PATCHVBMETAFLAG to false in the non-root case
   PATCHVBMETAFLAG=false
   # Make sure RECOVERYMODE has value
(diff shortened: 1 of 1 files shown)


vvb2060:
https://github.com/topjohnwu/Magisk/blob/0f5963f23120fabdfaa57ba0f144bbfac4803993/app/src/main/java/com/topjohnwu/magisk/ui/install/InstallViewModel.kt#L40

I think we may have forgotten something

vvb2060:
https://android-review.googlesource.com/c/platform/external/avb/+/320888
It looks like ro.boot.vbmeta.device exists on some a-only devices

osm0sis:
Hmm yeah, the original path check also had the non-A/B partition name:
```
[ -e /dev/block/by-name/vbmeta_a ] || [ -e /dev/block/by-name/vbmeta ]
```

But if there's no A-only equivalent to ro.product.ab_ota_partitions that will tell us vbmeta exists then I'm not sure what else we can do for unrooted checks.

osm0sis:
My A-Only Poco X2 on stock MIUI 12 (Android 11) has a vbmeta partition. Here's what I got there:

```
:/ $ getprop | grep vbmeta
[ro.boot.vbmeta.avb_version]: [1.0]
[ro.boot.vbmeta.device_state]: [locked]
[ro.boot.vbmeta.digest]: [1e3b40d142d6ab0b644a448d8bbc4e4b0a13d47f0bd592814c5431440bf79bd2]
[ro.boot.vbmeta.hash_alg]: [sha256]
[ro.boot.vbmeta.invalidate_on_error]: [yes]
[ro.boot.vbmeta.size]: [6720]
:/ $
```

osm0sis:
Per my above output and https://android.googlesource.com/platform/external/avb/+/88b13e12a0ebe3c5195dbb5f48ba00ec896d1517 size having a value could be another good indicator.

vvb2060:
ro.product.ab_ota_partitions must be an a/b device, we have detected a/b device, so it can be deleted.

osm0sis:
> ro.product.ab_ota_partitions must be an a/b device, we have detected a/b device, so it can be deleted.

But do A/B devices always have a vbmeta partition?

vvb2060:
https://github.com/topjohnwu/Magisk/blob/0f5963f23120fabdfaa57ba0f144bbfac4803993/app/src/main/java/com/topjohnwu/magisk/ui/install/InstallViewModel.kt#L40

BlackMesa123:
I still don't understand if implementing AVB without a vbmeta image is possible and there are devices that do not have it. Also regarding the "Patch vbmeta in boot image" option in the install page: from my understanding, changing the flags in a non-vbmeta image shouldn't be possible ([avb_slot_verify.c#801](https://android.googlesource.com/platform/external/avb/+/f6fbb68838cabad2e007b146e6bf3ca6ea30f310/libavb/avb_slot_verify.c#801)), is it?

osm0sis:
At least the first generation of A/B devices (e.g. Pixel XL, Essential PH-1) didn't have a vbmeta partition.

As for the boot.img vbmeta flags, yeah, Magisk started doing that awhile back.

ipdev99:
> At least the first generation of A/B devices (e.g. Pixel XL, Essential PH-1, and apparently some ChromeOS devices before that) didn't have a vbmeta partition. Not sure what their prop situation looks like (maybe someone can post output from last stock?)

Dug out my PH-1 (mata) and flashed the last factory build.
(QQ1A.200105.032)

No `ro.product.ab_ota_partitions` prop.

No separate vbemta partition.
No vbmeta at all?

```
mata:/ $ getprop | grep vb
1|mata:/ $ getprop | grep ro.product.ab_ota_partitions
1|mata:/ $ 
```

---

Patched with Magisk f86c66c99-ip (26105).
- Magisk did <b>not</b> give the option to patch vbmeta in boot image.

```
mata:/ $ su
mata:/ # getprop | grep vb
1|mata:/ # getprop | grep ro.product.ab_ota_partitions
1|mata:/ # 
```

[PH1 dev/block/by-name](https://github.com/topjohnwu/Magisk/files/12327913/PH1_by-name.txt)
[PH1 getprop](https://github.com/topjohnwu/Magisk/files/12327914/PH1_getprop.txt)
[PH1 getprop as root](https://github.com/topjohnwu/Magisk/files/12327915/PH1_su_getprop.txt)
[Magisk install log](https://github.com/topjohnwu/Magisk/files/12327921/magisk_install_log_2023-08-12T11.56.42.log)
[Magisk initial boot log](https://github.com/topjohnwu/Magisk/files/12327922/magisk_log_2023-08-12T11.59.36.log)

osm0sis:
> Edit: It looks like vbmeta was introduced in AVB 2.0 ? PH-1 appears to be using AVB 1.0.

Sort of? Check out my Poco X2 output above, avb 1.0 (with prop as such), but has A-only vbmeta partition.

osm0sis:
@vvb2060 I wonder if we could just check if ro.boot.vbmeta.avb_version has a value at all, it does seem to be the common prop on all the vbmeta devices

PR #7267, opened 2023-08-29
Fix syntax error in util_functions.sh

canyie:


osm0sis:
In addition to the missing then, it's true/false so could just be `elif ! $DATA; then` 👍

osm0sis (approved):

topjohnwu:
@osm0sis `$DATA` may be an empty string, so we cannot directly do that

topjohnwu (approved):

osm0sis:
> @osm0sis `$DATA` may be an empty string, so we cannot directly do that

@topjohnwu, oh.. how would that happen? It's only set to true/false in check_data and is treated as such there, and I don't see any other references to $DATA or DATA= anywhere..

https://github.com/topjohnwu/Magisk/blob/master/scripts/util_functions.sh#L510-L520

topjohnwu:
The `check_data` function is only called in recovery and will not be called in the Magisk app.

osm0sis:
Ahh, thanks, that makes more sense now. 🙂

PR #7284, opened 2023-09-01
Retain PREINITDEVICE during A-only addon.d

osm0sis:
Just realized this was missing from our recovery dirty flash PREINITDEVICE support!

### DIFF
--- a/scripts/addon.d.sh
+++ b/scripts/addon.d.sh
@@ -72,6 +72,12 @@ initialize() {
 
 main() {
   if ! $backuptool_ab; then
+    # Restore PREINITDEVICE from previous A-only partition
+    if [ -f config.orig ]; then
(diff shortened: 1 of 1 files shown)


osm0sis:
Adjusted the restore placement to be before the sleep 5 to ensure there aren't any potential timing issues that could result in config.orig getting removed.

Should be good to go now! 👍

osm0sis:
This should have been merged before stable to avoid loss in function for A-only users... 🫤

topjohnwu (approved):

osm0sis:
> I JUST installed 26.3 on such a device. What do I need to be careful about?

Install latest canary in a recovery and you're fine.

osm0sis:
> Alright. Is it okay to go back to 26.1 though? I would much prefer staying on stable.

Not a good idea.

PR #7333, opened 2023-09-12
No trailing zeros if the signed boot img is larger

yujincheng08:


osm0sis (approved):
Confirmed fixed in the CI build, thanks! 😁

topjohnwu (approved):

PR #7334, opened 2023-09-12
Update Gradle and Correction of artifact name, Modify issue template

s1204IT:

PR #7336, opened 2023-09-13
Update Gradle `8.3-rc-4`

s1204IT:
#7334


yujincheng08:
No reason to use RC.

s1204IT:
@vvb2060

s1204IT:
> No reason to use RC.

So why is the RC version currently being used?

yujincheng08:
@s1204IT Because we didn't upgrade it.

PR #7381, opened 2023-09-24
Fix $RECOVERYMODE from config being incorrectly overridden

### LINKED ISSUE #7346: App loses Recovery mode user setting for A-only legacy SAR devices
[magisk_log_2023-09-13T07.25.50.log](https://github.com/topjohnwu/Magisk/files/12614486/magisk_log_2023-09-13T07.25.50.log)
Device: Xiaomi Redmi Note 8T
Android version: 11
Magisk version code: 26301

From 26105, Magisk App loses Recovery mode user setting for A-only legacy SAR devices... 

On my Xiaomi RN8T I do not need Recovery mode despite this being an A-only legacy SAR (Ramdisk - No) device (bootloader supports ramdisk in /boot as added by Magisk patching) so I uncheck the option. However whenever the App is fully closed and reopened or some App functions are performed, the option is automatically checked again in the Install menu...

This means that unless users are aware their selection is lost/overridden and reset this every time they open/use Magisk App, Xiaomi devices etc behave strangely:

if user takes reboot system they end up in recovery, and if taking Direct Install /recovery is patched... On Samsung devices /recovery will be patched when taking Patch image also...

I had hoped that this issue would be solved by these fixes in 26202:
[Fix device detection scripts and logic](https://github.com/topjohnwu/Magisk/commit/75d905a56d6d7ceee62dc83be9c637870bc35a65) and [Always check mounts to detect legacy SAR on bootmode](https://github.com/topjohnwu/Magisk/commit/de00f1d5a94b83ad7bf49b0bdffab310e58db471)
but it continues to affect L-SAR (A-only) device users...

Including a log here taken from Alpha Magisk, but I did confirm that the issue occurs on TJW Canary and Debug Magisk also... I can supply new (Debug build) log if needed.

canyie:
Move legacy SAR checking logic into mount_partitions, and avoid calling get_flags before check_boot_ramdisk 
Fix #7346

### DIFF
--- a/app/src/main/res/raw/manager.sh
+++ b/app/src/main/res/raw/manager.sh
@@ -195,6 +195,9 @@ mount_partitions() {
   # Check whether non rootfs root dir exists
   SYSTEM_AS_ROOT=false
   grep ' / ' /proc/mounts | grep -qv 'rootfs' && SYSTEM_AS_ROOT=true
+
+  LEGACYSAR=false
+  grep ' / ' /proc/mounts | grep -q '/dev/root' && LEGACYSAR=true
 }
 
 get_flags() {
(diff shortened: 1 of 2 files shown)


osm0sis (approved):

pndwal:
... This still needs work per my issue comment, but thanks Devs! 👍

Issue #7408, opened 2023-10-08
Release build(s) broken after "Move more I/O operations into Rust"


ipdev99:
Release build(s) are broken starting with "Move more I/O operations into Rust" a55d570.

The release build patched images boot but, Magisk is not active.
Save for my Nexus 5, magisk is active but somewhat broken.

Nexus 5 Lineage 14.1 (Android v7.x).
Using the release build(s), magisk is active on boot but, gives the "Requires Additional Setup" dialog box.
_Even after selecting "OK" and the device reboots, the dialog box shows every time the app is opened._

---

The debug build(s) work as expected.

---

Device: Multiple
Android version: Multiple
Magisk version name: a55d57021 -> 6076b52c4
Magisk version code: 26301

Logs and stuff:
PocoF1 | MIUI Global 12.0.3 (Android 10).
[f1-boot_patched-a55d57021_R.zip](https://github.com/topjohnwu/Magisk/files/12841567/f1-boot_patched-a55d57021_R.zip)
[f1_magisk_first_boot.log](https://github.com/topjohnwu/Magisk/files/12841568/f1_magisk_first_boot.log)
[f1_magisk_install.log](https://github.com/topjohnwu/Magisk/files/12841569/f1_magisk_install.log)

Nexus 5 | Lineage 14.1 (Android 7.1.2).
[n5-boot_patched-a55d57021_R.zip](https://github.com/topjohnwu/Magisk/files/12841570/n5-boot_patched-a55d57021_R.zip)
[n5_magisk_first_boot.log](https://github.com/topjohnwu/Magisk/files/12841571/n5_magisk_first_boot.log)
[n5_magisk_install.log](https://github.com/topjohnwu/Magisk/files/12841572/n5_magisk_install.log)
[bugreport-NJH47F-2023-10-08-13-21-30.zip](https://github.com/topjohnwu/Magisk/files/12841582/bugreport-NJH47F-2023-10-08-13-21-30.zip)

Pixel 7 | Stock 20231005 (Android 14).
[p7-init_boot_patched-a55d57021_R.zip](https://github.com/topjohnwu/Magisk/files/12841573/p7-init_boot_patched-a55d57021_R.zip)
[p7_magisk_first_boot.log](https://github.com/topjohnwu/Magisk/files/12841574/p7_magisk_first_boot.log)
[p7_magisk_install.log](https://github.com/topjohnwu/Magisk/files/12841575/p7_magisk_install.log)
[...]

Issue #7356, opened 2023-09-17
Magisk not active if patched with release build(s) after "Add more to the Utf8CStr family".


ipdev99:
Hi.

I run personal build(s) of Magisk from the official repo (not a fork) so, the newest commits are always included.
I run new builds when I get home from work if there is a new commit.
_[ I understand that they are not supported because the commit(s) are (sometimes) still a work in progress. ]_

Long story short.

Last week, I updated using the "Direct Install" option from "4eaf701cb" to "4745e86c1" and lost root on reboot.
I went back to "4eaf701cb" for the night.

The next night, I installed the debug version on a different device to grab logs but, it worked?
So I tried the release build and once again, Magisk was not active after boot.

Tried on a few more devices.
Same story, the debug build(s) worked but the release build(s) did not.

This morning, I went backwards though the commits (using git bisect) to find when the release build(s) started to fail.
"Add more to the Utf8CStr family" 89aee6ffa was it.

Patched with release build.

```
oriole:/ $ magisk -V
/system/bin/sh: magisk: inaccessible or not found
```

Patched with debug build.
```
oriole:/ $ magisk -V
26301
oriole:/ $ magisk -v
89aee6ffa-ip:MAGISK:D
```

---

Device: Multiple
Android version: Multiple
Magisk version name: 89aee6ffa -> 877aeb66c
Magisk version code: 26301

Install log(s) look normal. (?)
_From Pixel 6 - Stock Android 13 (20230705)._
[P6 Debug install log](https://github.com/topjohnwu/Magisk/files/12643533/magisk_install_log_89aee6ffa-d.log)
[P6 Release install log](https://github.com/topjohnwu/Magisk/files/12643534/magisk_install_log_89aee6ffa-r.log)
[P6 Debug patched image](https://github.com/topjohnwu/Magisk/files/12643538/magisk_patched-89aee6ffa_debug.zip)
[P6 Release patched image](https://github.com/topjohnwu/Magisk/files/12643540/magisk_patched-89aee6ffa_release.zip)

yujincheng08:
need bugreport

canyie:
The signal that causes daemon to exit is SIGTRAP
![image](https://github.com/topjohnwu/Magisk/assets/31466456/bce59519-ff86-4ea6-8970-7fbc2b94c32b)

Issue #7068, opened 2023-06-11
Patching an already patched boot image produces an image with no `.backup/init`


canyie:
Device: vince
Android version: 12
Magisk version name: c556dd0
Magisk version code: 26102

canyie:
https://github.com/topjohnwu/Magisk/blob/feb44f875e59caf6cc52122d05bbdb88825bf503/scripts/boot_patch.sh#L128-L130

The current cpio command parser will exit when it executes "extract", thus the following "restore" command never gets executed.

PR #7409, opened 2023-10-09
AVD test release builds as well

canyie:
This helps us catch bugs that can only be reproduced in release builds (like #7356 and #7408) or bugs that can only be reproduced with an already patched ramdisk (like #7068)


canyie:
@osm0sis #7408 is the reason why tests fails, it is expected so you don't need to rerun them 😆

topjohnwu (approved):

PR #7441, opened 2023-10-20
Update Portuguese Translation

VD171:

PR #7442, opened 2023-10-20
Update portuguese translation

igormiguell:
I improved the grammar and corrected some translations that were in Portuguese from Portugal.


osm0sis:
There's already an open pt-rBR pull request at #7441, so please collaborate on that one instead of opening your own alternate, conflicting one.

PR #7543, opened 2023-11-19
Change Safe Mode trigger

rushiranpise:
Press Volume Down Key more than 3 times to trigger Magisk's safemode ( Similar to KernelSU) instead of long pressing volume down key for 3 secs or more

Long pressing the volume down key also enable android's safe mode on some devices. Hence, change safe mode trigger for Magisk  to avoid such conditions.


vvb2060:
this is expected

yujincheng08:
this is incorrectly implemented.

rushiranpise:
> this is incorrectly implemented.

thanks for the code review, if possible can you help with proper implementation?

yujincheng08:
no

PR #7546, opened 2023-11-21
Change Safe Mode trigger (copy of #7543 but fixed)

PHATWalrus:
Press Volume Down Key more than 3 times to trigger Magisk's safemode ( Similar to KernelSU) instead of long pressing volume down key for 3 secs or more

Long pressing the volume down key also enable android's safe mode on some devices. Hence, change safe mode trigger for Magisk to avoid such conditions (copy of https://github.com/topjohnwu/Magisk/pull/7543 but fixed hopefully)


yujincheng08:
not correct.

Issue #7593, opened 2023-12-04
Magisk 26.4 Zygisk fails on Pixel 7a GrapheneOS Android 14 231105.003


rockaport:
Zygisk fails to start/load. There's a couple of errors that seem related below. I understand this is a GrapheneOS device, so it might not be anything Magisk can do, but I noticed #7589 had the same issue.

Device: Pixel 7a
Android version: 14
Magisk version name: 26.4
Magisk version code: 26402 D

```
12-04 16:11:04.048   836   838 E : ^@mount /debug_ramdisk/magisk32 -> /system/lib/libzygisk.so failed with 2: No such file or directory
12-04 16:11:04.048   836   838 E : ^@mount (null) -> /system/lib/libzygisk.so failed with 22: Invalid argument
...
12-04 16:11:05.187   903   903 E : ^@zygisk64: unknown signature of nativeForkAndSpecialize(II[II[[IILjava/lang/String;Ljava/lang/String;[I[IZLjava/lang/String;Ljava/lang/String;Z[Ljava/lang/String;[Ljava/lang/String;ZZ[J)I
12-04 16:11:05.187   903   903 E : ^@zygisk64: unknown signature of nativeSpecializeAppProcess(II[II[[IILjava/lang/String;Ljava/lang/String;ZLjava/lang/String;Ljava/lang/String;Z[Ljava/lang/String;[Ljava/lang/String;ZZ[J)V
```

Robotition:
Any reason as to why is this not planned?

yujincheng08:
GrapheneOS officially does not recommend root, let alone zygisk, and thus we wont spend any effort to support it. if it works, all good; if it does not, let it go then.

heckarman:
Magisk lately seems to closing doors for many stuff and getting more invasive at the same time.

GrapheneOS openly admits that they hate the term root and discourage anyone from using that. So does every other OEM but they never openly admitted like this but you can guess that by their actions.

ccoager:
> GrapheneOS officially does not recommend root, let alone zygisk, and thus we wont spend any effort to support it. if it works, all good; if it does not, let it go then.

Which ROM's are supported by Magisk and also their vendor?  I looked through the Magisk documentation and couldn't find this anywhere, perhaps I missed it.

Issue #7589, opened 2023-12-03
Magisk won't work and Zygisk won't enable on latest GrapheneOs for Pixel 7


nodeam:
---Magisk Logs---
aa6c7c15 (26401)

12-03 14:35:32.987   846   846 I : Magisk aa6c7c15(26401) daemon started
12-03 14:35:32.988   846   846 I : * Device API level: 34
12-03 14:35:32.997   846   848 I : ** post-fs-data mode running
12-03 14:35:32.998   846   848 I : * Mounting mirrors
12-03 14:35:33.014   846   848 I : * Initializing Magisk environment
12-03 14:35:33.025   846   848 I : * Running post-fs-data.d scripts
12-03 14:35:33.028   846   848 I : * Running module post-fs-data scripts
12-03 14:35:33.033   846   848 I : * Loading modules
12-03 14:35:33.033   846   848 I : hosts: loading mount files
12-03 14:35:33.042   846   848 E : open /debug_ramdisk/magisk32 failed with 2: No such file or directory
12-03 14:35:33.042   846   848 E : sendfile failed with 9: Bad file descriptor
12-03 14:35:33.875   872   872 E : zygisk64: unknown signature of com/android/internal/os/Zygote#nativeForkAndSpecialize: (II[II[[IILjava/lang/String;Ljava/lang/String;[I[IZLjava/lang/String;Ljava/lang/String;Z[Ljava/lang/String;[Ljava/lang/String;ZZ[J)I
12-03 14:35:33.875   872   872 I : zygisk64: replaced com/android/internal/os/Zygote#nativeForkSystemServer
12-03 14:35:33.875   872   872 E : zygisk64: unknown signature of com/android/internal/os/Zygote#nativeSpecializeAppProcess: (II[II[[IILjava/lang/String;Ljava/lang/String;ZLjava/lang/String;Ljava/lang/String;Z[Ljava/lang/String;[Ljava/lang/String;ZZ[J)V
12-03 14:35:34.607   846   848 I : ** late_start service mode running
12-03 14:35:34.607   846   848 I : * Running service.d scripts
12-03 14:35:34.607   846   848 I : * Running module service scripts
12-03 14:35:34.865   846   848 I : denylist: initializing internal data structures
12-03 14:35:39.076   846   848 I : ** boot-complete triggered

panther-factory-2023112900.zip
https://releases.grapheneos.org/panther-factory-2023112900.zip

Device:  Pixel 7
[...]

PR #7606, opened 2023-12-09
fix(zygisk): Fix Zygisk on GrapheneOS Android 14

pixincreate:
The discussions have been done in these places:
- https://github.com/chenxiaolong/avbroot/issues/213
- https://github.com/topjohnwu/Magisk/issues/7593
- https://github.com/topjohnwu/Magisk/issues/7589
- https://github.com/HuskyDG/magisk-files/issues/20

And specially [this](https://github.com/topjohnwu/Magisk/issues/7593#issuecomment-1842929738)

GrapheneOS code changes can be seen [here](https://github.com/GrapheneOS/platform_frameworks_base/blob/14/core/java/com/android/internal/os/Zygote.java)

P.S: I don't have any context about how Magisk works at core level. I cherry-picked HuskyDG's [commit](https://github.com/HuskyDG/Magisk/commit/20a72f781cd635e766c27d10887b9a6aa5f27107) and modified it to make it work with Magisk.

Tests are run [here and `test on 34` failed](https://github.com/pixincreate/Magisk/actions/runs/7151323140).

### DIFF
--- a/native/src/core/zygisk/gen_jni_hooks.py
+++ b/native/src/core/zygisk/gen_jni_hooks.py
@@ -181,6 +181,10 @@ fas_samsung_p = ForkAndSpec('samsung_p', [uid, gid, gids, runtime_flags, rlimits
     se_info, Anon(jint), Anon(jint), nice_name, fds_to_close, fds_to_ignore, is_child_zygote,
     instruction_set, app_data_dir])
 
+fas_grapheneos_u = ForkAndSpec('grapheneos_u', [uid, gid, gids, runtime_flags, rlimits, mount_external,
+    se_info, nice_name, fds_to_close, fds_to_ignore, is_child_zygote, instruction_set, app_data_dir, 
+    is_top_app, pkg_data_info_list, whitelisted_data_info_list, mount_data_dirs, mount_storage_dirs, Anon(jlongArray)])
+
 spec_q = SpecApp('q', [uid, gid, gids, runtime_flags, rlimits, mount_external, se_info,
     nice_name, is_child_zygote, instruction_set, app_data_dir])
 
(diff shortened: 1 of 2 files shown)


pixincreate:
Ok, kindly update your Magisk's documentation explaining what operating systems / devices are supported (since the vision of Magisk today seems to cover basically the topline Pixel models running OS and Samsung devices and nothing else) so that we don't waste our time trying to fix it in here or ask for support.

To what I understand, @topjohnwu never had a vision of making this project to target only specific OSes but rather the vision of a project that serves an entire open source android community.

Robotition:
literally why was this closed? this PR properly fixes GrapheneOS support

Robotition:
perhaps someone could fork?

canyie:
GrapheneOS hates root and any modification, No reason to be compatible with operating systems that hate us.

Robotition:
Maybe because GrapheneOS is a properly hardened Operating system? People are free to root their device if they please while still having a more secure rooted experience than on AOSP.

yujincheng08:
Magisk wont make extra effort to support any specific customized ROM. Otherwise anyone can create a customized ROM that breaks Magisk and Magisk has to constantly update to support it.

Basically, Magisk tries its best to support most OEM OSes (unless those intentionally anti-root, e.g. vivo's OSes). As for customized ROMs, well, since Magisk will try to make sure it works on their upstreams, they can easily avoid breaking Magisk. If they still break Magisk, it means they dont care about, or even they are aganist Magisk. In that case, there's no reason for Magisk to still add special support.

pixincreate:
> GrapheneOS hates root and any modification, No reason to be compatible with operating systems that hate us.

There's nothing termed as "hating root" or "hating you". It is just that the project's threat model asks for it to be avoided. However but that does not mean that users of GrapheneOS should be deprived of root just because upstream does some changes with the way a certain feature work.

> Magisk wont make extra effort to support any specific customized ROM. Otherwise anyone can create a customized ROM that breaks Magisk and Magisk has to constantly update to support it.

I agree with this to an extent. But Magisk has also down significant efforts to make it compatible with Samsung devices that contradicts with what is mentioned above. Samsung did do a lot of stuffs that goes against Magisk but still Magisk was made to support it unlike in other cases like Vivo (FuntouchOS).

yujincheng08:
Samsun's OSes are OEM OSes, so Magisk would support. For VIVO, we have tried to support but with no luck, otherwise VIVO devices are supported.

pixincreate:
Why not GrapheneOS? Of course it is not an OEM, neither it is an after market Operating System but rather an alternate Operating System that supports verified boot. I'm saying this after using using it for more than 10 months and it feels like an OS from an OEM given that it is always up-to-date (they did support Android 14 within 4 -5 days of it being released into public way too early than any other OEM out there after Google, in case you're unaware) unlike other custom ROMs.

canyie:
> Maybe because GrapheneOS is a properly hardened Operating system? People are free to root their device if they please while still having a more secure rooted experience than on AOSP.

https://github.com/GrapheneOS/platform_frameworks_base/commit/9a9282b82be409576dbe74f4908b89e792192923
They introduce changes that would even break normal apps for their so-called "security". Why we should take effect to support a custom system that even doesn't pass CTS? Only systems that comply with the Android Compatibility Definition Document are Android.

pixincreate:
> They introduce changes that would even break normal apps for their so-called "security". 

I agree with this, and they're doing their stuff of trying to make it work with normal apps and I don't talk about that.  

> Only systems that comply with the Android Compatibility Definition Document are Android.

Any OS that is built on top of AOSP is Android. Android CDD is a requirement documentation that has to be met to make a certain version of Android say 14 compatible and it is what the first itself states:
> This document enumerates the requirements that must be met in order for devices to be compatible with Android 14.

And there is no where mentioned that systems that comply with Android CDD only has to be considered Android. 

CTS is exected to not pass as GrapheneOS is not a partner of Google by any means to get approval to get Play Services running.

Sources:
https://source.android.com/docs/compatibility/14/android-14-cdd
https://developer.android.com/google/play/integrity/verdicts

yujincheng08:
> Why not GrapheneOS? Of course it is not an OEM, neither it is an after market Operating System but rather an alternate Operating System that supports verified boot.

not a reason to support. every customized rom can highlight their feature.

and, as i hv said, if its built upon AOSP, they can avoid breaking magisk (because magisk supports AOSO). if they really break magisk, that means they dont care about magisk. you should ask them to support magisk (or submit their so-called security changes to AOSP), rather than begging here.

PR #7616, opened 2023-12-11
Update faq.md

RenateUSB:
Adding one more way to circumvent a bad module.


yujincheng08:
As said, incorrect. You dont have permission to set this prop.

RenateUSB:
I just recovered my device doing this. Without root on a stock boot image.

yujincheng08:
AVD 14: `type=1107 audit(1702305444.484:70): uid=0 auid=4294967295 ses=4294967295 subj=u:r:init:s0 msg='avc:  denied  { set } for property=persist.sys.safemode pid=6178 uid=2000 gid=2000 scontext=u:r:shell:s0 tcontext=u:object_r:safemode_prop:s0 tclass=property_service permissive=0'`

yujincheng08:
okay. but still it does not work on AVD 14.

PR #7620, opened 2023-12-12
Fixes for Android 14 QPR2 B2

KieronQuinn:
Added new method signatures and arguments to handle new `bindMountSyspropOverrides` argument.

I **think** both ForkAndSpec and Spec are required, though quite honestly I'm not 100% sure. 

Tested locally and working on a Pixel 8 Pro.


KieronQuinn:
@vvb2060 do you want this on your fork too or is here ok?

diosvany:
when will this be pushed out cant use any banking apps because i cant install usnf since i can't use zygisk i don't mean to rush anyone i just dont understand whats going on sometimes

topjohnwu (approved):

PR #7629, opened 2023-12-16
Skip svc for ro properties

yujincheng08:
ro properties' triggers should only be triggered once, otherwise it
may undefined behaviour.
This patch avoids triggering ro properties' actions again when using
resetprop to modify them.


canyie (dismissed):

topjohnwu:
Merged manually

PR #7630, opened 2023-12-16
Grant root to shell by default when using debug version

yujincheng08:
Since we now have explicit debug builds, and they are not recommended for daily usages, we can grant root to shell by default for better debugging experience.


vvb2060:
```sh
run-as com.topjohnwu.magisk
su
```

PR #7638, opened 2023-12-19
Revert "Don't support alternative binary paths"

vvb2060:


yujincheng08 (approved):
Tested on arm64 AVD 26.

PR #7664, opened 2023-12-30
WindowCompat

thonsi:
https://github.com/topjohnwu/Magisk/blob/master/app/src/main/java/com/topjohnwu/magisk/arch/UIActivity.kt#L53

or we can set in AndroidManifest.xml

android:windowSoftInputMode="adjustPan" or "adjustResize"


vvb2060:
I can't understand

PR #7665, opened 2023-12-30
Avoid hexpatch /init for 2SI when possible

yujincheng08:
Previous we hexpatch /init from /system/bin/init to /data/magiskinit to redirect the second stage init. However, some devices like sony has /init that does not directly invoke /system/bin/init, and thus the hexpatch fails.

In this patch, we further make use of AOSP `SwitchRoot` to help us bind mount /data/magisk to /system/bin/init after `SwitchRoot`.

Two important assumption about 2SI are i) that the second stage init is always /system/bin/init and ii) that the /sdcard (path after `SwitchRoot`) is always a symlink to `/storage/self/primary`. When these assumptions hold, during first stage init (before `SwitchRoot`) we can bind mount magiskinit to /sdcard, and create a symlink /storage/self/primary to /system/system/bin/init. By these steps, during `SwitchRoot`, AOSP init will try to mount move /sdcard to /system/sdcard. And /system/sdcard is symlink to /storage/self/primary, it will try to mount move /sdcard to /storage/self/primary. And /storage/self/primary in ramfs is now a symlink that points to /system/system/bin/init, thus AOSP will try to mount move /sdcard (which is a bind mount to magiskinit) to /system/system/bin/init. After chroot done by AOSP init, we then have a magiskinit bind mount on /system/bin/init, which is the second stage init.

An edge case is that some devices (like meizu) use 2SI but does not switch root. In this case, they must already have a /sdcard in the ramfs, thus we can check if /sdcard exists and fallback to hexpatch.

### DIFF
--- a/native/src/init/twostage.cpp
+++ b/native/src/init/twostage.cpp
@@ -10,11 +10,37 @@ using namespace std;
 
 void FirstStageInit::prepare() {
     prepare_data();
-    restore_ramdisk_init();
-    auto init = mmap_data("/init", true);
-    // Redirect original init to magiskinit
-    for (size_t off : init.patch(INIT_PATH, REDIR_PATH)) {
-        LOGD("Patch @ %08zX [" INIT_PATH "] -> [" REDIR_PATH "]\n", off);
+
+    if (struct stat st{}; fstatat(-1, "/sdcard", &st, AT_SYMLINK_NOFOLLOW) != 0 &&
+        fstatat(-1, "/first_stage_ramdisk/sdcard", &st, AT_SYMLINK_NOFOLLOW) != 0) {
+        if (config->force_normal_boot) {
+            xmkdirs("/first_stage_ramdisk/storage/self", 0755);
+            xsymlink("/system/system/bin/init", "/first_stage_ramdisk/storage/self/primary");
+            LOGD("Symlink /first_stage_ramdisk/storage/self/primary -> /system/system/bin/init\n");
+            close(xopen("/first_stage_ramdisk/sdcard", O_RDONLY | O_CREAT | O_CLOEXEC, 0));
+        } else {
+            xmkdirs("/storage/self", 0755);
+            xsymlink("/system/system/bin/init", "/storage/self/primary");
+            LOGD("Symlink /storage/self/primary -> /system/system/bin/init\n");
+        }
+        xrename("/init", "/sdcard");
+        // Try to keep magiskinit in rootfs for samsung RKP
+        if (mount("/sdcard", "/sdcard", nullptr, MS_BIND, nullptr) == 0) {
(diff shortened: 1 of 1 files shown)


BlackMesa123:
Regarding [Samsung RKP](https://github.com/topjohnwu/Magisk/pull/7665/files#diff-ad257acd61b18a5486ec237bd66ae00466b4973e4c40cef6883747f5ee4e748fR27-R35), I've tested this patch on a Galaxy S8+ with latest stock firmware by also removing the specific [RKP hex patches](https://github.com/topjohnwu/Magisk/blob/b04e1394c0dadbfd1314cc999f367a9a9781cfd8/scripts/boot_patch.sh#L222-L226), but it fails to boot ([last_kmsg.zip](https://github.com/topjohnwu/Magisk/files/14229104/last_kmsg.zip)).

yujincheng08:
@BlackMesa123 

At least `magiskinit` now passes, but `/sbin/magisk` itself is still causing kernel panic, which is wired:

```
<4>[    3.821657]  [7:         magisk: 3280] 
<4>[    3.821657]  [7:         magisk: 3280]  Superblock Mismatch #/sbin/magisk# vfsmnt #ffffffc0304b7000#sb #ffffffc870358000:ffffffc879011800:ffffffc87298f800#
<0>[    3.821670]  [7:         magisk: 3280] Kernel panic - not syncing: 
<0>[    3.821670]  [7:         magisk: 3280]  Illegal Execution file_name #/sbin/magisk#
```

yujincheng08:
@BlackMesa123 Can you send the kernel to me? I have an idea.

BlackMesa123:
> @BlackMesa123 Can you send the kernel to me? I have an idea.

Here: [boot.zip](https://github.com/topjohnwu/Magisk/files/14230345/boot.zip)
Kernel source code for reference:
- https://github.com/exynos-linux-stable/dreamlte/tree/tw90-android-stock
- https://github.com/Linux4/samsung_kernel/tree/SM-G955F

topjohnwu (approved):

madodream89:
hello i am trying installing apatch in note n9600 but is semicompatible and magisk modules not work i try uninstalling magisk give bootloops and reboots. the problem is the multidiseable and kribit kernel, the source code has magisk code

i need a version of multidiseable and kribit kernel without magisk boot patch and code

https://github.com/corsicanu/multidisabler-samsung
[update-binary.zip](https://github.com/user-attachments/files/17350728/update-binary.zip)

https://github.com/klabit87/android_kernel_samsung_sdm845/tree/q

[update-binary.zip](https://github.com/user-attachments/files/17350732/update-binary.zip)

the open source file is here before samsung opensource deleted the file and i trying port postmarketos in n9600 https://gitlab.com/madodream89/samsung-galaxy-note-9-n9600-sdm845-postmarketos-port

PR #7680, opened 2024-01-04
Mount only necessary mirror directories that will be used magic mount

HuskyDG:
Mount only necessary mirror directories that will be used magic mount. I don't think we should recursively bind mount all partitions to mirror as some are unnecessary?

### DIFF
--- a/native/src/core/bootstages.cpp
+++ b/native/src/core/bootstages.cpp
@@ -81,27 +81,9 @@ static void mount_mirrors() {
     ssprintf(path, sizeof(path), "%s/" WORKERDIR, get_magisk_tmp());
     xmount("worker", path, "tmpfs", 0, "mode=755");
     xmount(nullptr, path, nullptr, MS_PRIVATE, nullptr);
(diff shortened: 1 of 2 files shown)


vvb2060:
af01a3629652857392f0ecae15d4c2f3d66b0e1d
This is to support overlayfs.  It is mounted atomically on modern devices with only one system call.

yujincheng08:
I think it good. But the implementation seems incorrect.

nihal697:
> I think it good. But the implementation seems incorrect.

what exactly is incorrect ? could you please explain?

yujincheng08:
> what exactly is incorrect ? could you please explain?
1. There are duplicated make private for mirror
2. Most mirror_path is no longer needed, so we dont actually need `mirror_node`
3. Since mirror is not complete, we dont need it anymore and we should umount it.

yujincheng08:
After careful thinking, we no longer need mirror anymore. Will sumbit a new PR instead.

PR #7692, opened 2024-01-05
Update Polish translation

kam821:
- Added missing strings, fixed translation, escaped quotas.
- Some context dependent values (like target_pid) may require better translation in the future.
- Also, 'DenyList' translation, although correct, could be replaced in the future by better fitting polish translation of 'block list' / 'rejection list'.


Wojtaz0w:
As a Pole I approve this pull request. It makes not-so-clear text much simpler to understand.

topjohnwu (approved):

PR #7704, opened 2024-01-10
call pthread_atfork only once

5ec1cff:
prevent from memory leaking


yujincheng08:
`exec_task` will be called by multiple threads, so I think your midification still have a race that the handler may be registered more than once.

PR #7741, opened 2024-01-30
Update faq.md

pndwal:
Since system Safe Mode may activate without Magisk 'Safe Mode', this clarification will prevent users from erroneously concluding either that Safe Mode doesn't work or that modules are not the issue since 'Safe Mode' was apparently triggered... Resolves this:
https://github.com/topjohnwu/Magisk/issues/4624#issuecomment-1891244704

FAQ edit quick link:
https://github.com/pndwal/Magisk/blob/patch-1/docs/faq.md#q-i-installed-a-module-and-it-bootlooped-my-device-help


osm0sis (changes requested):
https://github.com/topjohnwu/Magisk/pull/7741/files/9dcd853bf1264e0e74655fed25779947808a1cf5#diff-57cdcc6b6701a7ce3b3a8f8c0366e0e611e6fb1a1b8dd7cd29e3263b3e064c8bR13

This part seems to suddenly become way too verbose. Can you distill it down to the bare minimum?

docs/faq.md
```diff
@@ -4,7 +4,16 @@
...
+1) Many online guides for entering Safe Mode say 'When the animated logo appears, press and hold the volume down button until the system boots' or similar. This may actually be too late for Magisk detection however and result in activating system Safe Mode but modules aren't disabled.
+2) By pressing the volume down button some seconds before the animation and releasing it when the boot animation appears, Magisk Safe Mode should be activated without activating system Safe Mode and the device should then simply boot to normal system with modules disabled.
+3) By pressing the volume down button some seconds before the animation and holding it until the system boots, both Magisk Safe Mode and system Safe Mode should be activated. Next, after rebooting back to normal system modules will be disabled.
+
+Varying the key combo timing that online guides give for entering system Safe Mode is often needed to activate Magisk 'Safe Mode' as described above, but this will add an empty directory named 'disable' to the modules directory to disable modules when next booted with Magisk thereby bypassing module related bootloop conditions.
```
osm0sis, line 14: Slight tweaks I'd recommended for accuracy and clarity:

*but in any case this

*empty file

docs/faq.md
```diff
@@ -4,7 +4,16 @@
...
+
+Further, achieving timing for scenario 2) above may be preferable to 3) since system Safe Mode will also reset many other user preferences. Disabled modules can be managed through the Magisk app.
```
osm0sis, line 16: I'd drop this entire paragraph, people can discern all this from the method list and surrounding instructions.

osm0sis (changes requested):

docs/faq.md
pndwal: Thanks... Maybe not 'in any case' since point is it'll only happen in second case (ie. when timing is varied) which is the point of this clarification. 🙃... Will try to improve it though.

... Ah yes, empty file...

Haven't discovered how to add bolding, italics etc in these edits strangely... Please say what I'm blind to! 😬

pndwal: Hope that's better... Thanks for input/reviewing! 👍

docs/faq.md
pndwal: Yeah, briefer the better... Done. 👍

pndwal:
Hope that's a bit less bookish... Thanks for your help with it... 👍

docs/faq.md
```diff
@@ -4,7 +4,14 @@
...
+1) Many online guides for entering Safe Mode say 'When the animated logo appears, press and hold the volume down button until the system boots' or similar. This may actually be _too late_ for Magisk detection however and result in activating system Safe Mode but modules are not disabled.
+2) By pressing the volume down button some seconds before the animation and releasing it when the boot animation appears, Magisk Safe Mode should be activated without activating system Safe Mode and the device should then simply boot to normal system with modules disabled.
+3) By pressing the volume down button some seconds before the animation and holding it until the system boots, both Magisk Safe Mode and system Safe Mode should be activated. Next, after rebooting back to normal system modules will be disabled.
+
+Varying the key combo timing that online guides give for entering system Safe Mode is often needed to activate Magisk 'Safe Mode' as described above, but achieving this will cause an empty file named 'disable' to be created in the modules directory to disable modules when next booted with Magisk thereby bypassing module related bootloop conditions.
```
osm0sis, line 14: Last line is still a bit redundant with the previous paragraph "Next, after rebooting back to normal system modules will be disabled", and there's still something needlessly wordy about "named 'disable' to be created in the modules directory to disable modules when next booted with Magisk".

I'd suggest:

*in each module directory, thereby bypassing module related bootloop conditions.

Since it disabling all modules is covered in the first paragraph, and it happens early enough that it occurs in the current boot, and then remaining disabled is covered in the previous paragraph.

pndwal: Hopefully this flows better?...

osm0sis:
Yeah seems concise enough now. 👍

osm0sis (approved):

PR #7742, opened 2024-01-30
Fix zygisk unload

yujincheng08:
Magisk is not hooking itself, so it's safe to directly call `dlclose`. Otherwise, using `old_dlclose` will cause the unload to fail.


vvb2060:
regression 930bb8687f0

vvb2060 (approved):

topjohnwu (approved):

PR #7748, opened 2024-02-02
CI also builds release on win and mac

yujincheng08:
Because release version has some codes that are not activated on debug version. We should also check them.


topjohnwu:
The release code is already checked by the job "Build Magisk artifacts". The "Test building on <os>" jobs are only to ensure the project is buildable on those platforms, it doesn't need to build in release mode.

PR #7754, opened 2024-02-05
Add no decompression flag to magiskboot split

osm0sis:
Addresses https://github.com/topjohnwu/Magisk/pull/6597#issuecomment-1518709863

### DIFF
--- a/native/src/boot/bootimg.cpp
+++ b/native/src/boot/bootimg.cpp
@@ -520,12 +520,12 @@ bool boot_img::verify(const char *cert) const {
     return rust::verify_boot_image(*this, cert);
 }
 
-int split_image_dtb(const char *filename) {
+int split_image_dtb(const char *filename, bool skip_decomp) {
     mmap_data img(filename);
 
     if (int off = find_dtb_offset(img.buf(), img.sz()); off > 0) {
         format_t fmt = check_fmt_lg(img.buf(), img.sz());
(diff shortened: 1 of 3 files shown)


yujincheng08 (commented):
It causes segmentation fault:

> [1]    1782 segmentation fault

yujincheng08 (changes requested):
It causes segmentation fault:

> Program received signal SIGSEGV, Segmentation fault.
0x000000000024c271 in ?? ()

osm0sis:
Thanks, I'll look into it. 🤦‍♂️

osm0sis:
> It causes segmentation fault:
> 
> > Program received signal SIGSEGV, Segmentation fault.
> > 0x000000000024c271 in ?? ()

Worked correctly here, what was your input file?

yujincheng08 (dismissed):
`./magiskboot split -n` will cause SIGSEGV

osm0sis:
Thanks for more info! Reproduced and I'll see about it tonight. 🙂

osm0sis:
Should be good now. Thanks for catching that! I'd looked at the way switches like that are handled in other magiskboot features when I was doing the original changes, but forgot about it after my testing. 😅🤦‍♂️

yujincheng08 (approved):

topjohnwu (approved):

PR #7867, opened 2024-02-28
Always allow modifying denylist

aviraxp:


aviraxp:
Guess I shouldn't test this on ci. Directly leads me to bootloader.

PR #7889, opened 2024-03-08
Upgrade AGP

yujincheng08:


canyie:
https://github.com/topjohnwu/Magisk/pull/7888#issuecomment-1984912167
~~ci failed. you should at least test building it.~~

PR #7899, opened 2024-03-09
Update strings.xml

maingocvinh:
Better message when grant root access


maingocvinh:
this toast message make me remember about old root era, when first superuser app release

PR #7904, opened 2024-03-11
Do not delete the READE.md file during module installation

HChenX:
I think the behavior of automatically deleting README.md files is unreasonable.
Because README.md usually contains basic information about the module, usage methods, copyright statements, etc.
If the module developer needs to delete it, they can delete it themselves in the module installation file, instead of deleting it directly by default.
Additionally, due to the existence of multiple languages, there may be files such as README-en.md, but according to the original code logic, these files will not be deleted by default. Therefore, overall, I believe that deleting README.md by default is redundant.
Thanks.

### DIFF
--- a/scripts/util_functions.sh
+++ b/scripts/util_functions.sh
@@ -718,7 +718,7 @@ install_module() {
   # Remove stuff that doesn't belong to modules and clean up any empty directories
   rm -rf \
   $MODPATH/system/placeholder $MODPATH/customize.sh \
-  $MODPATH/README.md $MODPATH/.git*
+  $MODPATH/.git*
   rmdir -p $MODPATH 2>/dev/null
 
   cd /


vvb2060:
https://github.com/topjohnwu/magisk-module-installer/tree/73dbc1c90dfff16c2d36f453acdeb4b750ca460b
This is for compatibility with module template.

HChenX:
> https://github.com/topjohnwu/magisk-module-installer/tree/73dbc1c90dfff16c2d36f453acdeb4b750ca460b This is for compatibility with module template.

If it is necessary to delete the README.md by default for compatibility with the default template, it doesn't feel quite right.

HChenX:
> https://github.com/topjohnwu/magisk-module-installer/tree/73dbc1c90dfff16c2d36f453acdeb4b750ca460b This is for compatibility with module template.

Because many developers do not use the default template, or rather they should know how to modify or delete the README.md of the default template themselves

vvb2060:
For modules that want to keep the readme file, just use `readme.md` as the file name. (or `README.MD` etc

HChenX:
> For modules that want to keep the readme file, just use `readme.md` as the file name. (or `README.MD` etc

It is indeed feasible, but it doesn't quite align with GitHub's naming conventions. Having to manually change README.md to readme.md for every package distribution module seems a bit cumbersome. Alternatively, using readme.md directly on GitHub is possible, but as mentioned earlier, it doesn't quite adhere to naming conventions.

HChenX:
> For modules that want to keep the readme file, just use `readme.md` as the file name. (or `README.MD` etc

The most important aspect is that if there are files like README-en.md included, they won't be automatically deleted by default. This doesn't achieve the intended goal of 'Remove stuff that doesn't belong to modules and clean up any empty directories,' as described in the original code comments. That's why I consider it unnecessary.

vvb2060:
```sh
mv README.md rEaDme.Md
```
Just add a line of script

HChenX:
> ```shell
> mv README.md rEaDme.Md
> ```
> 
> Just add a line of script只需添加一行脚本

Of course, if you think about it that way, I can't do much about it. I just hope it aligns with the logical expectations of most people.

PR #7893, opened 2024-03-08
Upgrade AGP

yujincheng08:

PR #7924, opened 2024-03-20
Revert "Upgrade AGP"

yujincheng08:
Reverts topjohnwu/Magisk#7893


osm0sis:
8.3.1 is out now too, not sure if it would fix anything though.

PR #7938, opened 2024-03-23
Inform developers live patch mode is discouraged

aviraxp:
Details: https://android-review.googlesource.com/c/kernel/common/+/3009995


canyie (commented):
Shall we add warnings about other commands such as `magiskpolicy --dump` as well, since they are also affected?

PR #7975, opened 2024-04-05
Add `action.sh` so that user can run modules' scripts manually from app

yujincheng08:


app/src/main/java/com/topjohnwu/magisk/core/model/module/LocalModule.kt
```diff
@@ -63,6 +65,11 @@ data class LocalModule(
             }
         }
 
+    fun executeAction() {
+        // su to clear exported environment variables
+        Runtime.getRuntime().exec(arrayOf("su", "-c", "cd $path; ASH_STANDALONE=1 /data/adb/magisk/busybox sh action.sh"))
```
yujincheng08, line 70: Not very familiar with Magisk app so I hardcoded `/data/adb/magisk/busybox` here. Maybe there can be better way to obtain this `NVBASE`.

PR #8012, opened 2024-04-26
add missing `xz_dec_end`

yujincheng08:


canyie (commented):
Wouldn't the allocated memory buffer be released when executing the original init?

yujincheng08:
ofc. but that's not a reason to be lazy.

topjohnwu (approved):

PR #8037, opened 2024-05-10
Remove incorrect instruction of resetprop

aviraxp:
This instruction is wrong because the prop doesn't exist when boot is not completed.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -109,7 +109,7 @@ Update JSON format:
 
 #### Shell scripts (`*.sh`)
 
-Please read the [Boot Scripts](#boot-scripts) section to understand the difference between `post-fs-data.sh` and `service.sh`. For most module developers, `service.sh` should be good enough if you just need to run a boot script. If you need to wait for boot completed, you can use `resetprop -w sys.boot_completed 0`.
+Please read the [Boot Scripts](#boot-scripts) section to understand the difference between `post-fs-data.sh` and `service.sh`. For most module developers, `service.sh` should be good enough if you just need to run a boot script.
 
 In all scripts of your module, please use `MODDIR=${0%/*}` to get your module's base directory path; do **NOT** hardcode your module path in scripts.
 If Zygisk is enabled, the environment variable `ZYGISK_ENABLED` will be set to `1`.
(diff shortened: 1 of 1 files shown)


aviraxp:
cc @yujincheng08

yujincheng08:
I intentionally didnt write it in a while-loop for non-existent props, because not all the people like the `sleep 1`.

vvb2060:
Why not use this?
```sh
while [ "$(getprop sys.boot_completed)" != "1" ]
  do
    sleep 1
done
```
Does `resetprop -w` have any advantages over it?

dxkf177:
> This instruction is wrong because the prop doesn't exist when boot is not completed.

docs/guides.md
```diff
@@ -109,7 +109,7 @@ Update JSON format:
 
 #### Shell scripts (`*.sh`)
 
-Please read the [Boot Scripts](#boot-scripts) section to understand the difference between `post-fs-data.sh` and `service.sh`. For most module developers, `service.sh` should be good enough if you just need to run a boot script. If you need to wait for boot completed, you can use `resetprop -w sys.boot_completed 0`.
+Please read the [Boot Scripts](#boot-scripts) section to understand the difference between `post-fs-data.sh` and `service.sh`. For most module developers, `service.sh` should be good enough if you just need to run a boot script. If you need to wait for boot completed, you can use `while [ ! $(resetprop -w sys.boot_completed 0) ]; do sleep 1; done`.
```
dxkf177, line 112:
```suggestion
Please read the [Boot Scripts](#boot-scripts) section to understand the difference between `post-fs-data.sh` and `service.sh`. For most module developers, `service.sh` should be good enough if you just need to run a boot script. If you need to wait for boot completed, you can use `while [ ! $(resetprop -w sys.boot_completed 0) ]; do sleep 1; done`.
```

docs/guides.md
```diff
@@ -109,7 +109,7 @@ Update JSON format:
...
 
 In all scripts of your module, please use `MODDIR=${0%/*}` to get your module's base directory path; do **NOT** hardcode your module path in scripts.
 If Zygisk is enabled, the environment variable `ZYGISK_ENABLED` will be set to `1`.
```
dxkf177, line 115:
```suggestion
 If Zygisk is enabled, the environment variable `ZYGISK_ENABLED` will be set to `1`.
```

PR #8095, opened 2024-05-31
Change build.py env ANDROID_SDK_ROOT to ANDROID_HOME

False-Mask:
ANDROID_SDK_ROOT, which also points to the SDK installation directory, is deprecated. ref: https://developer.android.com/tools/variables#envar


topjohnwu:
Thank you, I've handled this in b26db8cee6fed78c7625c2cafb17652cbf38f07e

PR #8164, opened 2024-06-29
Handle APPLICATION_PREFERENCES for stub

aviraxp:


buildSrc/src/main/java/Codegen.kt
arizmendileo, line 1:
```java

```

vvb2060:
Intentionally unsupported to be the same as most apps

PR #8194, opened 2024-07-06
Update logo.png

rivercory:
Hi! I improved the design of this repository logo!
Please review my PR and let me know if you have any improvement! Thank you.

PR #8196, opened 2024-07-07
Update logo.png

rivercory:
https://github.com/topjohnwu/Magisk/pull/8194
Changed text color to make it look smooth in dark mode!


rivercory:
@vvb2060 Why?

PR #8224, opened 2024-07-16
Remove NVBASE

vvb2060:
We only move /cache/data_adb/magisk and /data/magisk to /data/adb/magisk (#7638), so NVBASE is redundant and we can just use MAGISKBIN.

### DIFF
--- a/scripts/app_functions.sh
+++ b/scripts/app_functions.sh
@@ -43,7 +43,7 @@ fix_env() {
   # Cleanup and make dirs
   rm -rf $MAGISKBIN/*
   mkdir -p $MAGISKBIN 2>/dev/null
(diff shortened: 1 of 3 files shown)


topjohnwu (approved):

reiryuki:
Hi, I just want to give my opinion that, not caring about backwards compatibility is not good. Hopefully Magisk will care more about backwards compatibility in the future versions.  Thank you.

vvb2060:
@reiryuki We have cared about backward compatibility for 4 years. https://github.com/topjohnwu/Magisk/blob/master/scripts/module_installer.sh

reiryuki:
@vvb2060 sorry I mean about this NVBASE variable removal or any other future variable removal

PR #8257, opened 2024-07-27
Make Zygote injection optional for Zygisk modules

aviraxp:
This is useful when a module has Zygisk features, but also has other features which can work without Zygisk, such as TrickyStore.


vvb2060:
why not rename /data/adb/modules/abc/zygisk

aviraxp:
This won't work for third party zygisk solutions, because it cannot be guaranteed that all modules' scripts have been executed before zygisknext. Thus zygisknext won't load modules becasue it doesn't know if modules is a zygisk module.

PR #8293, opened 2024-08-07
Use `androidboot.partition_map` as a fallback for matching partition names in the preinit finding.

chsbuffer:
Fixes **Play Games Developer Emulator** `magiskinit: Cannot find preinit metadata, abort!`.

only tested on the **Play Games Developer Emulator**, need to do some test on the production one, which runs on older version of Android iirc.

### DIFF
--- a/native/src/init/getinfo.cpp
+++ b/native/src/init/getinfo.cpp
@@ -229,14 +229,6 @@ kv_pairs load_partition_map() {
     return {};
 }
 
-std::string get_partition_name_for_device(const kv_pairs &partition_map, const std::string &query_device){
-    for (const auto &[device, partition] : partition_map) {
-        if (query_device == device)
-            return partition;
-    }
-    return {};
-}
-
 bool check_two_stage() {
     if (access("/apex", F_OK) == 0)
         return true;
--- a/native/src/init/getinfo.cpp
+++ b/native/src/init/getinfo.cpp
@@ -41,16 +41,16 @@ static string extract_quoted_str_until(chars<escapes...>, chars<breaks...>,
 }
 
 // Parse string into key value pairs.
-// The string format: [delim][key][padding]=[padding][value][delim]
-template<char delim, char... padding>
+// The string format: [delim][key][padding][eq][padding][value][delim]
+template<char delim, char eq, char... padding>
 static kv_pairs parse_impl(chars<padding...>, string_view str) {
     kv_pairs kv;
-    char skip_array[] = {'=', padding...};
+    char skip_array[] = {eq, padding...};
     string_view skip(skip_array, std::size(skip_array));
     bool quoted = false;
     for (size_t pos = 0u; pos < str.size(); pos = str.find_first_not_of(delim, pos)) {
         auto key = extract_quoted_str_until(
-                chars<padding..., delim>{}, chars<'='>{}, str, pos, quoted);
+                chars<padding..., delim>{}, chars<eq>{}, str, pos, quoted);
         pos = str.find_first_not_of(skip, pos);
         if (pos == string_view::npos || str[pos] == delim) {
             kv.emplace_back(key, "");
@@ -63,10 +63,13 @@ static kv_pairs parse_impl(chars<padding...>, string_view str) {
 }
 
 static kv_pairs parse_cmdline(string_view str) {
-    return parse_impl<' '>(chars<>{}, str);
+    return parse_impl<' ', '='>(chars<>{}, str);
 }
 static kv_pairs parse_bootconfig(string_view str) {
-    return parse_impl<'\n'>(chars<' '>{}, str);
+    return parse_impl<'\n', '='>(chars<' '>{}, str);
+}
+static kv_pairs parse_partition_map(std::string_view str) {
+    return parse_impl<';', ','>(chars<>{}, str);
 }
 
 #define test_bit(bit, array) (array[bit / 8] & (1 << (bit % 8)))
@@ -214,25 +217,16 @@ void load_kernel_info(BootConfig *config) {
 // https://android.googlesource.com/platform/system/core/+/refs/heads/android13-release/init/devices.cpp#191
 
 kv_pairs load_partition_map() {
-    kv_pairs partition_map;
-    auto fn = [&partition_map](const kv_pairs &kv) {
-        for (const auto &[key, value] : kv) {
-            if (key != "androidboot.partition_map")
-                continue; 
-            for (const auto &map : split(value, ";")) {
-                auto map_pieces = split(map, ",");
-                if (map_pieces.size() != 2) {
-                    LOGE("Expected a comma separated device,partition mapping, but found '%s'", map.c_str());
-                    continue;
-                }
-                partition_map.emplace_back(map_pieces[0], map_pieces[1]);
-            }
-        }
-    };
-
-    fn(parse_cmdline(full_read("/proc/cmdline")));
-    fn(parse_bootconfig(full_read("/proc/bootconfig")));
-    return partition_map;
+    const string_view kPartitionMapKey = "androidboot.partition_map";
+    for (const auto &[key, value] : parse_cmdline(full_read("/proc/cmdline"))) {
+        if (key == kPartitionMapKey)
+            return parse_partition_map(value);
+    }
+    for (const auto &[key, value] : parse_bootconfig(full_read("/proc/bootconfig"))) {
+        if (key == kPartitionMapKey)
+            return parse_partition_map(value);
+    }
+    return {};
 }
 
 std::string get_partition_name_for_device(const kv_pairs &partition_map, const std::string &query_device){
--- a/native/src/init/getinfo.cpp
+++ b/native/src/init/getinfo.cpp
@@ -207,6 +207,42 @@ void load_kernel_info(BootConfig *config) {
     config->print();
 }
 
+// `androidboot.partition_map` allows associating a partition name for a raw block device
+// through a comma separated and semicolon deliminated list. For example,
+// `androidboot.partition_map=vdb,metadata;vdc,userdata` maps `vdb` to `metadata` and `vdc` to
+// `userdata`.
+// https://android.googlesource.com/platform/system/core/+/refs/heads/android13-release/init/devices.cpp#191
+
+kv_pairs load_partition_map() {
+    kv_pairs partition_map;
+    auto fn = [&partition_map](const kv_pairs &kv) {
+        for (const auto &[key, value] : kv) {
+            if (key != "androidboot.partition_map")
+                continue; 
+            for (const auto &map : split(value, ";")) {
+                auto map_pieces = split(map, ",");
+                if (map_pieces.size() != 2) {
+                    LOGE("Expected a comma separated device,partition mapping, but found '%s'", map.c_str());
+                    continue;
+                }
+                partition_map.emplace_back(map_pieces[0], map_pieces[1]);
+            }
+        }
+    };
+
+    fn(parse_cmdline(full_read("/proc/cmdline")));
+    fn(parse_bootconfig(full_read("/proc/bootconfig")));
+    return partition_map;
+}
+
+std::string get_partition_name_for_device(const kv_pairs &partition_map, const std::string &query_device){
+    for (const auto &[device, partition] : partition_map) {
+        if (query_device == device)
+            return partition;
+    }
+    return {};
+}
+
 bool check_two_stage() {
     if (access("/apex", F_OK) == 0)
         return true;
--- a/native/src/init/mount.cpp
+++ b/native/src/init/mount.cpp
@@ -78,7 +78,7 @@ static struct {
 } blk_info;
 
 static dev_t setup_block() {
-    const auto partition_map = load_partition_map();
+    static const auto partition_map = load_partition_map();
     if (dev_list.empty())
         collect_devices(partition_map);
 
--- a/native/src/init/mount.cpp
+++ b/native/src/init/mount.cpp
@@ -56,6 +56,8 @@ static void collect_devices() {
             if (access(path, F_OK) == 0) {
                 auto name = rtrim(full_read(path));
                 strcpy(dev.dmname, name.data());
+            } else {
+                dev.dmname[0] = '\0';
             }
             sprintf(path, "/sys/dev/block/%s", entry->d_name);
             xrealpath(path, dev.devpath, sizeof(dev.devpath));
@@ -70,11 +72,17 @@ static struct {
 } blk_info;
 
 static dev_t setup_block() {
+    const auto partition_map = load_partition_map();
     if (dev_list.empty())
         collect_devices();
 
     for (int tries = 0; tries < 3; ++tries) {
         for (auto &dev : dev_list) {
+            // use androidboot.partition_map as partname fallback.
+            if (dev.partname[0] == '\0') {
+                auto partname = get_partition_name_for_device(partition_map, dev.devname);
+                strcpy(dev.partname, partname.c_str());
+            }
             if (strcasecmp(dev.partname, blk_info.partname) == 0)
                 LOGD("Setup %s: [%s] (%d, %d)\n", dev.partname, dev.devname, dev.major, dev.minor);
             else if (strcasecmp(dev.dmname, blk_info.partname) == 0)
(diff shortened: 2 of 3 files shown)


yujincheng08:
Doesn't `magisk --preinit-device` print the device name before mapping? Like the following

QQ_1723051068867

chsbuffer:
`magisk --preinit-device`:
```
resetprop: get prop [ro.crypto.state]: [encrypted]
resetprop: get prop [ro.crypto.type]: [file]
resetprop: get prop [ro.crypto.metadata.enabled]: [true]
metadata
```

script:
```shell
parse_device() {
  local device_name="$1"
  local devpath="/sys/dev/block/$device_name"
  local uevent_file="/sys/dev/block/$device_name/uevent"
  local dm_name_file="/sys/dev/block/$device_name/dm/name"
        while IFS='=' read -r key value; do
          case "$key" in
            MAJOR)
              local major="$value"
              ;;
            MINOR)
              local minor="$value"
              ;;
            DEVNAME)
              local devname="$value"
              ;;
            PARTNAME)
              local partname="$value"
              ;;
          esac
        done < "$uevent_file"
  local dmname=$(cat "$dm_name_file" 2>/dev/null) || dm_name=""

  printf "%-18s\t%-18s\t%-18s\t%s\n" "$partname" "$dmname" "$devname" "$devpath"
}

printf "%-18s\t%-18s\t%-18s\t%s\n" "partname" "dmname" "devname" "devpath"
for device in /sys/dev/block/*; do
   parse_device "$(basename "$device")"
done
```
output:
```
partname                dmname                  devname                 devpath
                                                ram0                    /sys/dev/block/1:0
                                                ram1                    /sys/dev/block/1:1
                                                ram10                   /sys/dev/block/1:10
                                                ram11                   /sys/dev/block/1:11
                                                ram12                   /sys/dev/block/1:12
                                                ram13                   /sys/dev/block/1:13
                                                ram14                   /sys/dev/block/1:14
                                                ram15                   /sys/dev/block/1:15
                                                ram2                    /sys/dev/block/1:2
                                                ram3                    /sys/dev/block/1:3
                                                ram4                    /sys/dev/block/1:4
                                                ram5                    /sys/dev/block/1:5
                                                ram6                    /sys/dev/block/1:6
                                                ram7                    /sys/dev/block/1:7
                                                ram8                    /sys/dev/block/1:8
                                                ram9                    /sys/dev/block/1:9
                                                zram0                   /sys/dev/block/251:0
                                                vda                     /sys/dev/block/253:0
super                                           vda1                    /sys/dev/block/253:1
                                                vdb                     /sys/dev/block/253:16
boot_a                                          vda2                    /sys/dev/block/253:2
vendor_boot_a                                   vda3                    /sys/dev/block/253:3
                                                vdc                     /sys/dev/block/253:32
vbmeta_a                                        vda4                    /sys/dev/block/253:4
                                                vdd                     /sys/dev/block/253:48
vbmeta_system_a                                 vda5                    /sys/dev/block/253:5
                        system_a                dm-0                    /sys/dev/block/254:0
                        vendor_a                dm-1                    /sys/dev/block/254:1
                        product_a               dm-2                    /sys/dev/block/254:2
                        system-verity           dm-3                    /sys/dev/block/254:3
                        vendor-verity           dm-4                    /sys/dev/block/254:4
                        product-verity          dm-5                    /sys/dev/block/254:5
                        userdata                dm-6                    /sys/dev/block/254:6
                                                loop0                   /sys/dev/block/7:0
                                                loop13                  /sys/dev/block/7:104
                                                loop14                  /sys/dev/block/7:112
                                                loop15                  /sys/dev/block/7:120
                                                loop2                   /sys/dev/block/7:16
                                                loop3                   /sys/dev/block/7:24
                                                loop4                   /sys/dev/block/7:32
                                                loop5                   /sys/dev/block/7:40
                                                loop6                   /sys/dev/block/7:48
                                                loop7                   /sys/dev/block/7:56
                                                loop8                   /sys/dev/block/7:64
                                                loop9                   /sys/dev/block/7:72
                                                loop1                   /sys/dev/block/7:8
                                                loop10                  /sys/dev/block/7:80
                                                loop11                  /sys/dev/block/7:88
                                                loop12                  /sys/dev/block/7:96
```

yujincheng08:
send mountinfo?

chsbuffer:
[mountinfo.txt](https://github.com/user-attachments/files/16536738/mountinfo.txt)

yujincheng08:
also send /proc/self/mounts?

chsbuffer:
[mounts.txt](https://github.com/user-attachments/files/16536776/mounts.txt)

chsbuffer:
Do you think maybe the fix should be a magisk `find_preinit_device()` change instead of a magiskinit change?

yujincheng08:
hmm, both of them show `/dev/block/by-name/metadata` as the mount device and does not follow the link.

chsbuffer:
`/metadata` may be mounted by `init` which follows `/ramdisk/fstab.cut_cvm`.
I guess adding a `xrealpath` call to `find_preinit_device()` can fix this in another way?

yujincheng08:
No, we have no permission to call `realpath` on `/dev/block/by-name`

chsbuffer:
Back to the pr state. I tested the change on the production one. turns out it runs the same version of android as the dev one. The patched system boots and the log looks good, but I can't get adb to work to check if magisk is working.
The developer emulator has an adbproxy service running /vendor/bin/adbproxy connect 127.0.0.1:5555 10.0.2.2:38068 but the production one doesn't. After messing with the Root Directory Overlay System, I still can't get it to work.

native/src/init/getinfo.cpp
```diff
@@ -171,6 +171,45 @@ if (access(file_name, R_OK) == 0) {                                 \
...
+// `androidboot.partition_map` allows associating a partition name for a raw block device
+// through a comma separated and semicolon deliminated list. For example,
+// `androidboot.partition_map=vdb,metadata;vdc,userdata` maps `vdb` to `metadata` and `vdc` to
+// `userdata`.
+// https://android.googlesource.com/platform/system/core/+/refs/heads/android13-release/init/devices.cpp#191
+
+kv_pairs partition_map;
+
+void load_partition_map() {
+    auto fn = [](const kv_pairs &kv) {
```
chsbuffer, line 185:
```suggestion
    auto fn = [](const kv_pairs& kv) {
```

native/src/init/getinfo.cpp
```diff
@@ -171,6 +171,45 @@ if (access(file_name, R_OK) == 0) {                                 \
...
+        for (const auto &[key, value] : kv) {
```
chsbuffer, line 186:
```suggestion
        for (const auto& [key, value] : kv) {
```

native/src/init/getinfo.cpp
```diff
@@ -171,6 +171,45 @@ if (access(file_name, R_OK) == 0) {                                 \
...
+            if (key == "androidboot.partition_map") {
```
chsbuffer, line 187: flip the flow

chsbuffer (commented):
some code style issue

native/src/init/getinfo.cpp
```diff
@@ -171,6 +171,45 @@ if (access(file_name, R_OK) == 0) {                                 \
...
+                        continue;
+                    }
+                    partition_map.emplace_back(map_pieces[0], map_pieces[1]);
+                }
+            }
+        }
+    };
+ 
+    fn(parse_cmdline(full_read("/proc/cmdline")));
+    fn(parse_bootconfig(full_read("/proc/bootconfig")));
```
yujincheng08, line 201: why not add `androidboot.partition_map` to `BootConfig::set` and use `Config` here?

chsbuffer: `BootConfig::set` is only being called from `load_kernel_info`, which doesn't be called in the SecondStageInit

yujincheng08: Then why not make partition_map a `SecondStageInit`'s member

chsbuffer: `LegacySARInit`, `RootFSInit` and `SecondStageInit` call the `setup_block()`. 

It might be worth extending the scope of `load_partition_map`, or adding some comments to make it clear that it was added as a hack for the Google Play Games Developer Emulator.

 I think we should go for the extended scope approach, which will reduce the code paths.

yujincheng08: Then this is not good. Just make `partition_map` a local variable of `setup_block`, since you only use `get_partition_name_for_device` there. I think it's safe for other `Init`s to support this fallback.

chsbuffer: the `partition_map` will never change, I don't see the point of make it a local variable.

yujincheng08: It's not about mutable or not.

native/src/init/mount.cpp
```diff
@@ -70,11 +72,17 @@ static struct {
 } blk_info;
 
 static dev_t setup_block() {
+    const auto partition_map = load_partition_map();
```
yujincheng08, line 75: can be static

native/src/init/mount.cpp
```diff
@@ -70,11 +72,17 @@ static struct {
...
     if (dev_list.empty())
         collect_devices();
 
     for (int tries = 0; tries < 3; ++tries) {
         for (auto &dev : dev_list) {
+            // use androidboot.partition_map as partname fallback.
+            if (dev.partname[0] == '\0') {
+                auto partname = get_partition_name_for_device(partition_map, dev.devname);
```
yujincheng08, line 83: can also inline `get_partition_name_for_device` with `find_if`

yujincheng08:
partition_map was added since Android 13; bootconfig was added since Android 12, so I think we only need to parse bootconfig.

native/src/init/mount.cpp
```diff
@@ -70,11 +72,17 @@ static struct {
...
+                strcpy(dev.partname, partname.c_str());
```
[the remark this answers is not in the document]
yujincheng08: https://github.com/topjohnwu/Magisk/blob/07835a3e0e44ad398a82294eccfa93994ccfd9ff/native/src/init/getinfo.cpp#L145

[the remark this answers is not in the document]
chsbuffer: Do I need to change `c_str()` to `data()` for consistency?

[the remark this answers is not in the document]
yujincheng08: yes. as they are the same, we prefer `data()`

chsbuffer:
> partition_map was added since Android 13; bootconfig was added since Android 12, so I think we only need to parse bootconfig.

Google Play Games passes `androidboot.partition_map` from kernel cmdline, and the AOSP code also parse both bootconfig and kernel cmdline

chsbuffer:
I will upload the kernel cmdline and bootconfig later

chsbuffer:
```
vsoc_kiwi_x86_64:/ # cat /proc/cmdline | tr ' ' '\n'
stack_depot_disable=on
cgroup_disable=pressure
cgroup.memory=nokmem

bootconfig
init=/init
loop.max_part=7
printk.devkmsg=on
bootconfig
panic=-1
console=hvc0
earlycon=uart8250,io,0x3f8
androidboot.verifiedbootstate=orange
androidboot.boot_devices=pci0000:00/0000:00:01.0,pci0000:00/0000:00:02.0,pci0000:00/0000:00:03.0,pci0000:00/0000:00:04.0
androidboot.console=hvc0
androidboot.hardware.gltransport=virtio-gpu-asg
androidboot.hardware.gralloc=minigbm
androidboot.hardware.hwcomposer=ranchu
androidboot.kiwi.adbproxy.enabled=1
androidboot.kiwi.adbproxy.port=38068
androidboot.kiwi.client_version=24.6.902.0
androidboot.kiwi.logcatserial=1
androidboot.kiwi.logcatserial.dev=/dev/hvc0
androidboot.kiwi.logcatserial.filter=System.err:E,.android.common:E,*:I
androidboot.kiwi.metrics_group=devdogfood
androidboot.partition_map=vdb,metadata;vdc,userdata;vdd,misc
androidboot.serialno=2e9e3643154548bf83bd1d5e5bff0afe
androidboot.pcs.recvport=50051
androidboot.pcs.sendport=50050
androidboot.timezone=Asia/Shanghai
androidboot.hw_timeout_multiplier=1
noxsaves
no_timer_check
tsc=reliable
androidboot.pcs.use_vsock=true
androidboot.pcs.enable_host_ime=true
androidboot.ime.ime_port=60006
androidboot.pcs.enable_volume_control=true
androidboot.pcs.enable_pre_game_launch_memory_compaction=false
androidboot.opengles.version=196609
androidboot.dirty.expire.centisecs=500
androidboot.kiwi.prefetch.io_depth=20
androidboot.kiwi.prefetch.max_fds=2000
androidboot.kiwi.prefetch=replay
androidboot.async.persist.writes=true
androidboot.enable.min.free.kbytes=1
androidboot.standard.min.free.kbytes=61440
androidboot.optimized.min.free.kbytes=30720
androidboot.kiwi_enable_android_notifications=true
androidboot.kiwi_enable_multi_android_user=false
androidboot.kiwi_enable_multi_window=false
androidboot.kiwi_disable_force_stop=true
androidboot.kiwi_adv_controls.enable_control_hints_overlay=false
androidboot.kiwi_adv_controls.enable_context_detection=false
androidboot.kiwi_adv_controls.controls_editor_mode=disabled
snd-hda-intel.enable=0
snd_intel8x0.inside_vm=1
snd_intel8x0.ac97_clock=48000
androidboot.hardware.hwcomposer.display_finder_mode=drm
androidboot.hardware.egl=angle
androidboot.hardware.vulkan=ranchu
nopat
androidboot.kiwi.enable_houdini=1
androidboot.locale=zh-CN
androidboot.vsock_gatekeeper_port=31540
androidboot.vsock_keymaster_port=31541
androidboot.kiwi_out_period_count=4
virtio_snd.msg_timeout_ms=5000
androidboot.pcs.events_service=true
androidboot.angle_log_texture_decompression_events=true
ramoops.mem_address=0x1b8000000
ramoops.mem_size=0x100000

vsoc_kiwi_x86_64:/ # cat /proc/bootconfig
androidboot.hardware = "cutf_cvm"
androidboot.slot_suffix = "_a"
vsoc_kiwi_x86_64:/ #
```

chsbuffer:
Maybe the bootloader is bugging out, I notice the difference in bootconfig size between before and after patching the `bios.rom`.

native/src/init/mount.cpp
chsbuffer: I can't write it in a clean way with find_if.

native/src/init/mount.cpp
```diff
@@ -55,7 +57,12 @@ static void collect_devices() {
             sprintf(path, "/sys/dev/block/%s/dm/name", entry->d_name);
             if (access(path, F_OK) == 0) {
                 auto name = rtrim(full_read(path));
-                strcpy(dev.dmname, name.data());
+                strscpy(dev.dmname, name.data(), sizeof(dev.dmname));
+            } else if (auto it = std::ranges::find_if(partition_map, [&](const auto &i) {
```
chsbuffer, line 61: Wrong condition

topjohnwu:
@yujincheng08 can you review and approve once the change is in an acceptable state?

yujincheng08 (approved):

yujincheng08 (approved):

native/src/init/mount.cpp
```diff
@@ -29,21 +29,23 @@ bool avd_hack = false;
...
         if (key == "MAJOR")
             dev->major = parse_int(value.data());
         else if (key == "MINOR")
             dev->minor = parse_int(value.data());
         else if (key == "DEVNAME")
-            strcpy(dev->devname, value.data());
+            strscpy(dev->devname, value.data(), sizeof(dev->devname));
         else if (key == "PARTNAME")
-            strcpy(dev->partname, value.data());
+            strscpy(dev->partname, value.data(), sizeof(dev->devname));
```
DorCoMaNdO, line 42: Is it intentional that this is using `sizeof(dev->devname)` instead of `sizeof(dev->partname)`? Might be irrelevant, I'm not exactly sure how this code is called, I just happened to glance at it the other day.

chsbuffer: It doesn't really matter, partname and devname are both 32.
@yujincheng08 Can you answer whether this is intentional or unquestionable?

I'm also worried that strscpy won't add zero terminators, maybe we should use `strncat(dst, src, buffer_size - 1)`.

Issue #8279, opened 2024-08-01
Verify gms is system app


vvb2060:
https://developer.android.com/privacy-and-security/risks/create-package-context

Diapolo:
@topjohnwu Does this impose any implications for users which are relying on microG?

PR #8321, opened 2024-08-20
Require GMS to be system app

aviraxp:
Fixes https://github.com/topjohnwu/Magisk/issues/8279.


canyie (approved):

pixincreate:
Does this mean that if GMS is not installed or nor a part of system, Magisk would not work or marks the device as insecure? I believe latter is what Magisk is aiming for?  

Care to elaborate?

Issue #8335, opened 2024-08-24
VaultKeeper Crashing


diabl0w:
Device: SM-F926B
Android version: 14
Magisk version name: Canary
Magisk version code: 27007

Phone is running fine, visually, but in the background Vaultkeeper is just constantly crashing and never stops (will continue to try and increase trycount)

according to [this](https://github.com/topjohnwu/Magisk/blob/8e82113bce1486c0dce7d8d1fdc08a1372ab26a6/native/src/init/rootdir.cpp#L67) VaultKeeper should be getting disabled, but I am not sure if that means entirety of VaultKeeper or what

```
08-24 15:06:57.220  5171  5171 I vaultkeeper: There is no property(ro.factory.factory_binary)
08-24 15:06:57.220  5171  5171 I VaultKeeper_COMMON: VK 4.4.1 [Client] /vendor/bin/vaultkeeperd. => [Vault] VK. [CMD NO.] 2040
08-24 15:06:57.220  5171  5171 I VaultKeeper_COMMON: [CMD:INITIALIZE]
08-24 15:06:57.291  5171  5171 E VaultKeeper_COMMON: Occured Error into TA side(-10032//vendor/bin/vaultkeeperd/VK)
08-24 15:06:57.292  5171  5171 E VaultKeeper_COMMON: Failed cmdProcessing(-10032)
08-24 15:06:57.292  5171  5171 E vaultkeeper: runCommand error.(-10032)
08-24 15:06:57.292  5171  5171 E vaultkeeper: There is no VK ID
08-24 15:06:57.292  5171  5171 E vaultkeeper: VaultKeeper service preparation is failed
08-24 15:06:57.672  5216  5216 E CASS_DEBUG: VaultKeeper is not ready. try again Retry cnt(10).
```

diabl0w:
from ps -A, the parent PID is indeed init that is trying to launch it (PPID = 1 below):
`5811     1  1000  5.9   0 20 vaultkeeperd                /vendor/bin/vaultkeeperd`

vvb2060:
https://dumps.tadiphone.dev/dumps/samsung/dm3q/-/blob/dm3qxxx-user-14-UP1A.231005.007-S918BXXS6CXGA-release-keys/vendor/etc/init/vaultkeeper_common.rc
VaultKeeper has been moved to the vendor partition for a long time. The code in magisk is out of date and should be removed.

vvb2060:
[...]

PR #8336, opened 2024-08-25
Remove vaultkeeper handling

aviraxp:
These were added in 2019, but Android 10 move them to vendor
which Magisk cannot handle. Remove them as they never work.

### DIFF
--- a/native/src/init/rootdir.cpp
+++ b/native/src/init/rootdir.cpp
@@ -64,12 +64,7 @@ static void patch_rc_scripts(const char *src_path, const char *tmp_path, bool wr
         if (!dest) return;
         LOGD("Patching " INIT_RC " in %s\n", src_path);
         file_readline(false, src.get(), [&dest](string_view line) -> bool {
-            // Do not start vaultkeeper
-            if (str_contains(line, "start vaultkeeper")) {
-                LOGD("Remove vaultkeeper\n");
-                return true;
-            }
-            // Do not run flash_recovery
+            // Do not run flash_recovery on Android 10 and below
             if (line.starts_with("service flash_recovery")) {
                 LOGD("Remove flash_recovery\n");
                 fprintf(dest.get(), "service flash_recovery /system/bin/true\n");


osm0sis:
There are still Android 9 devices which would need this. Magisk supports Android 6+ currently.

yujincheng08 (approved):

vvb2060:
At least we need to add a note that it only works on Android 9 and earlier.

vvb2060:
For vaultkeeper, I think we should remove the code in magisk and let the module  to stop vaultkeeper/cass/proca services.

osm0sis:
Sounds good!

osm0sis (approved):

salvogiangri:
> For vaultkeeper, I think we should remove the code in magisk and let a module to stop vaultkeeper/cass/proca services.

There's no need to stop these services as they don't do any harm. What would be nice as an addition instead is a kernel hex patch to kill PROCA, since that causes VaultKeeper to always crash in the background due to failed device integrity checks.

salvogiangri:
What I did try myself is to patch the kernel image header since it contains the physical address to the PROCA config stored in the image itself (https://github.com/Mesa-Labs-Archive/android_kernel_samsung_sm7325/blob/e76fd1fdabdca7554e5d7df0dbd10dcbabdfb3fd/arch/arm64/kernel/head.S#L83-L87), there's a specific magic that if matched should disable PROCA entirely in TZ side, unfortunately tho this wasn't enough but I didn't dig enough to find why this alone won't work

diabl0w:
> For vaultkeeper, I think we should remove the code in magisk and let a module to stop vaultkeeper/cass/proca services.

just my 2 cents that I do believe that magisk does have some responsibility here in addressing this, as the crashing is induced by installing magisk itself (vbmeta/boot modification)

salvogiangri:
> just my 2 cents that I do believe that magisk does have some responsibility here in addressing this, as the crashing is induced by installing magisk itself (vbmeta/boot modification)

As I said in https://github.com/topjohnwu/Magisk/issues/8335#issuecomment-2319313175 this isn’t directly caused by Magisk as this happens as soon as the ICD status in ICCC/bootloader changes, which means a custom image has been flashed on the phone (kernel, AVB, OS…). However it would be great for Magisk to properly address this as this is what the majority uses.

> EDIT: also does this mean magisk is officially dropping Android < 10 support? that seems like a big deal just to save a few lines of otherwise harmless code

I believe the VaultKeeper code was firstly added due to the big amount of people being blocked from flashing custom images due to KnoxGuard (ex [RMM](https://xdaforums.com/t/how-to-root-after-january-security-patch-sm-a520f-a5-2017.3739225/post-75360965)) since it was introduced in early 2018, this Knox feature relies on VK to read/store its status in the device’s RPMB. However there are a lot of system components which rely on VK, so imho it’s not a good practice to completely get rid of it.

diabl0w:
> As I said in [#8335 (comment)](https://github.com/topjohnwu/Magisk/issues/8335#issuecomment-2319313175) this isn’t directly caused by Magisk as this happens as soon as the ICD status in ICCC/bootloader changes, which means a custom image has been flashed on the phone (kernel, AVB, OS…). However it would be great for Magisk to properly address this as this is what the majority uses.

The argument for a module because any custom partition will trigger it is sort of a catch 22 since you'd have to use magisk to install the module anyway, so doesn't make much sense to me to not just include the fix.

As you mnetioned, if i only unlock my bootloader, I dont have any issues. It is the moment I use magisk that the services will start to crash, therefore Magisk should be involved in handling this. And this should be considered different than Google device integrity check failure type issues, which magisk shouldnt be respnsible for. The difference being that proca/vaultkeeper is an issue where services are uncontrollably crashing at the kernel level and causing unneeded burden in the background of the device.

> I believe the VaultKeeper code was firstly added due to the big amount of people being blocked from flashing custom images due to KnoxGuard (ex [RMM](https://xdaforums.com/t/how-to-root-after-january-security-patch-sm-a520f-a5-2017.3739225/post-75360965)) since it was introduced in early 2018, this Knox feature relies on VK to read/store its status in the device’s RPMB. However there are a lot of system components which rely on VK, so imho it’s not a good practice to completely get rid of it.
Again, i appreciate your deep level of knowledge on these subjects but i have to disagree with your logic. If you are blocked from flashing custom images (such as magisk) without disabling it, then it sounds like it would be required to stay in code. And I don't see any open issues since 2018 regarding side effects of disabling VK, so I would say the users who this is relevant to aren't having significant problems

salvogiangri:
> Again, i appreciate your deep level of knowledge on these subjects but i have to disagree with your logic. If you are blocked from flashing custom images (such as magisk) without disabling it, then it sounds like it would be required to stay in code. And I don't see any open issues since 2018 regarding side effects of disabling VK, so I would say the users who this is relevant to aren't having significant problems
> 
> Of course, if there is a better way to allow the custom flashing on these older devices then we should do it that way, but until someone has the desire to reinvestigate these old devices, IMO the code should be left as it has been for 5+ years without issue

I haven't dug up on KnoxGuard/RMM since when I firstly discovered it back in 2018 (how time flies!), but I'm pretty sure there's another better way to avoid such issues on legacy devices (Android 8.x I guess). Nowadays KnoxGuard isn't really an issue as the only case where the device would be locked is due to a bad provisioning on first boot (Prenormal state, this is why it's mentioned to stay connected to the internet during the setup wizard) or due to the device being under MDM/bought via a carrier plan (Locked or Active state, in which case it can only be unlocked remotely by Samsung). Samsung addressed the issue where the device could be locked with custom binaries installed already (Broken status), so Magisk or any other root solution do not need to address this.

yujincheng08 (approved):

vvb2060:
@salvogiangri After #8351 is merged, should this PR be merged? Will it break old devices?

salvogiangri:
> @salvogiangri After #8351 is merged, should this PR be merged? Will it break old devices?

I think it should be good for devices running Android 9 onwards (One UI), Idk about legacy devices running Oreo but I had reports of RMM being completely broken on stock too due to Samsung server-side changes

Issue #7254, opened 2023-08-24
Bootloop - Samsung S9


alexcheveau:
Device: Samsung S9 (G9600)
Android version: 10
Magisk version name: v26.1 (debug)
Magisk version code: ee34f775 (26105)

Followed exactly all the steps in https://topjohnwu.github.io/Magisk/install.html#samsung-devices
After flashing the patched AP I got bootloop.

[magisk_install_log_2023-08-24T15.06.10.log](https://github.com/topjohnwu/Magisk/files/12432416/magisk_install_log_2023-08-24T15.06.10.log)
[mounts.txt](https://github.com/topjohnwu/Magisk/files/12432417/mounts.txt)
[last_kmsg.txt](https://github.com/topjohnwu/Magisk/files/12432418/last_kmsg.txt)

BlackMesa123:
By looking at the kernel logs seems like the kernel panic is caused by RKP blocking magiskinit (https://github.com/droidian-starqlte/android_kernel_samsung_sdm845/blob/8fde28b4cdfb0c1ed2501c559cb7e593879a978d/fs/exec.c#L1355), I also noticed in the install logs I can't see the RKP disable patch being applied in the kernel image:

https://github.com/topjohnwu/Magisk/blob/0c28a916be99e0cad31af0500c4e88ad4456df99/scripts/boot_patch.sh#L223-L227

alexcheveau:
Retried all steps with the new version (26.2 - debug) same result: bootloop
[last_kmsg.txt](https://github.com/topjohnwu/Magisk/files/12455445/last_kmsg.txt)
[magisk_install_log_2023-08-28T11.17.06.log](https://github.com/topjohnwu/Magisk/files/12455446/magisk_install_log_2023-08-28T11.17.06.log)

osm0sis:
Is there a newer version of RKP that needs to be patched out then? Is there source available for this device? Clearly the old patch doesn't work for it.

alexcheveau:
> Fact is the existing RKP patch didn't find a match in your kernel. Is there source for this device kernel?

I think it's this: https://github.com/droidian-devices/linux-android-samsung-starqlte

osm0sis:
Alright well if you're lucky someone might be able to figure out how to patch it.

BlackMesa123:
> Is there a newer version of RKP that needs to be patched out then? Is there source available for this device? Clearly the old patch doesn't work for it.

[...]

PR #8351, opened 2024-08-30
Disable Samsung PROCA

### LINKED ISSUE #8335: VaultKeeper Crashing

salvogiangri:
Fixes #8335

### DIFF
--- a/scripts/boot_patch.sh
+++ b/scripts/boot_patch.sh
@@ -220,6 +220,13 @@ if [ -f kernel ]; then
   # After:  [mov w2, #-32768]
   ./magiskboot hexpatch kernel 821B8012 E2FF8F12 && PATCHEDKERNEL=true
 
+  # Disable Samsung PROCA
+  # proca_config -> proca_magisk
+  ./magiskboot hexpatch kernel \
+  70726F63615F636F6E66696700 \
+  70726F63615F6D616769736B00 \
+  && PATCHEDKERNEL=true
+
   # Force kernel to load rootfs for legacy SAR devices
   # skip_initramfs -> want_initramfs
   $LEGACYSAR && ./magiskboot hexpatch kernel \


salvogiangri:
A few notes here:
- Legacy Samsung devices might require a different patch since the PROCA policy is fetched via its physical address in memory, which is stored in the [kernel image header](https://github.com/Mesa-Labs-Archive/android_kernel_samsung_sm7325/blob/e76fd1fdabdca7554e5d7df0dbd10dcbabdfb3fd/arch/arm64/kernel/head.S#L83-L87) as follows:
```c
#ifdef CONFIG_PROCA
#define PROCA_CONF_OFFSET_IMAGE_LE64 \
	DEFINE_IMAGE_LE64(_proca_conf_offset, g_proca_config - _text);
#else
#define PROCA_CONF_OFFSET_IMAGE_LE64
#endif

/*
 * These will output as part of the Image header, which should be little-endian
 * regardless of the endianness of the kernel. While constant values could be
 * endian swapped in head.S, all are done here for consistency.
 */
#define HEAD_SYMBOLS						\
	DEFINE_IMAGE_LE64(_kernel_size_le, _end - _text);	\
	DEFINE_IMAGE_LE64(_kernel_offset_le, TEXT_OFFSET);	\
	DEFINE_IMAGE_LE64(_kernel_flags_le, __HEAD_FLAGS);	\
	PROCA_CONF_OFFSET_IMAGE_LE64
```
An additional patch to change the offset to the special magic Samsung uses to disable PROCA would be a nice addition. I thought about something like this but this might not work on non-64 bit kernel images:
```sh
printf "\xef\xec\xef\xec" | dd of="kernel" bs=1 seek=40 count=4 conv=notrunc
```
- The patch won't apply on Samsung devices shipped with Android 13 onwards due to the new boot header v4 specs (init_boot.img), perhaps a way to patch both boot.img and init_boot.img in [processTar](https://github.com/topjohnwu/Magisk/blob/8e82113bce1486c0dce7d8d1fdc08a1372ab26a6/app/core/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt#L225) should be added?

yujincheng08:
Please also help check https://github.com/topjohnwu/Magisk/issues/7254#issuecomment-2273807100

topjohnwu (approved):

PR #8355, opened 2024-08-31
Check GMS app signature

salvogiangri:


vvb2060:
stub.apk is embedded in the boot.img and it must be as small as possible.

vvb2060:
> Multiple valid signatures exist for this app. If you're having trouble with installation due to a mismatched signature, try a different one. See this FAQ entry for more info.
Known valid signatures: bd32 58e1 3891 2169.

https://www.apkmirror.com/apk/google-inc/google-play-services/google-play-services-24-34-33-release/

PR #8359, opened 2024-09-01
Print Info.legacySAR to log

canyie:


yujincheng08 (approved):

vvb2060 (changes requested):
https://github.com/topjohnwu/Magisk/blob/8e82113bce1486c0dce7d8d1fdc08a1372ab26a6/app/apk/src/main/java/com/topjohnwu/magisk/ui/log/LogViewModel.kt#L88-L89

grep /dev/root

PR #8362, opened 2024-09-03
Improved APK download and installation compatibility with Android 14 (API level 34) and earlier versions. Fixed download spinning bug.

ArKT-7:
- **Bug Fix:** Resolved an issue where the app would get stuck with a spinning progress indicator during downloading on Android 14 and above, as well as on some specific ROMs.
- **Dynamic Loading Adjustment:** Updated the code to handle APK downloads and installations more efficiently, particularly for Android 14 and later versions. The `dlAPK` method was modified to differentiate the APK download and installation process based on the Android version.
- **MediaStore Integration:** For Android 14 and above, integrated with the `MediaStore` API to handle APK downloads and ensure proper installation through the system's package installer.
- **Backward Compatibility:** Maintained backward compatibility by ensuring older Android versions (below 14) still follow the previous method using file download and dynamic loading, while also preserving the original behavior for hidden APKs.
- **Error Handling:** Enhanced error handling during resource injection and downloading processes.
- **Code Refactoring:** General code improvements, including the introduction of content values for file management and clearer separation of logic for different Android versions.

This update not only improves compatibility across different Android versions but also addresses a critical bug affecting the downloading process, ensuring a smoother user experience.

### DIFF
--- a/app/stub/src/main/java/com/topjohnwu/magisk/DownloadActivity.java
+++ b/app/stub/src/main/java/com/topjohnwu/magisk/DownloadActivity.java
@@ -22,6 +22,10 @@ import android.system.Os;
 import android.system.OsConstants;
 import android.util.Log;
 import android.view.ContextThemeWrapper;
+import android.content.ContentValues;
+import android.net.Uri;
+import android.os.Environment;
+import android.provider.MediaStore;
 
 import com.topjohnwu.magisk.net.Networking;
 import com.topjohnwu.magisk.net.Request;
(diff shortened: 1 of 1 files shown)


yujincheng08 (changes requested):
code not formatted

ArKT-7:
where it is not formatted because I tested, and it was working fine

vvb2060:
Intent.ACTION_VIEW? I don't understand what you are doing

PR #8368, opened 2024-09-06
Don't repack kernels into images that never had a kernel

kerneltoast:
If the image to be repacked doesn't have a kernel in it, then it should not have a kernel shoehorned into the new image. This issue would cause magiskboot to generate broken vendor images upon repacking when a kernel file is present in the same directory as the components to be repacked.

Fix it by never repacking a kernel into an image that didn't have one.

### DIFF
--- a/native/src/boot/bootimg.cpp
+++ b/native/src/boot/bootimg.cpp
@@ -703,35 +703,38 @@ void repack(const char *src_img, const char *out_img, bool skip_comp) {
         // Copy zImage headers
         xwrite(fd, boot.z_hdr, boot.z_info.hdr_sz);
     }
-    if (access(KERNEL_FILE, R_OK) == 0) {
-        mmap_data m(KERNEL_FILE);
-        if (!skip_comp && !COMPRESSED_ANY(check_fmt(m.buf(), m.sz())) && COMPRESSED(boot.k_fmt)) {
-            // Always use zopfli for zImage compression
-            auto fmt = (boot.flags[ZIMAGE_KERNEL] && boot.k_fmt == GZIP) ? ZOPFLI : boot.k_fmt;
-            hdr->kernel_size() = compress(fmt, fd, m.buf(), m.sz());
-        } else {
-            hdr->kernel_size() = xwrite(fd, m.buf(), m.sz());
-        }
-
-        if (boot.flags[ZIMAGE_KERNEL]) {
-            if (hdr->kernel_size() > boot.hdr->kernel_size()) {
-                fprintf(stderr, "! Recompressed kernel is too large, using original kernel\n");
-                ftruncate64(fd, lseek64(fd, - (off64_t) hdr->kernel_size(), SEEK_CUR));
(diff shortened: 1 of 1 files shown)


yujincheng08:
If you don't want the kernel to be packed, remove it.

https://github.com/topjohnwu/Magisk/blob/a82f17c59499ba2c46001c5bc54f8ec5f506e5d8/scripts/boot_patch.sh#L239

kerneltoast:
> If you don't want the kernel to be packed, remove it.
> 
> https://github.com/topjohnwu/Magisk/blob/a82f17c59499ba2c46001c5bc54f8ec5f506e5d8/scripts/boot_patch.sh#L239

magiskboot should not generate garbage images. `mkbootimg.py` explicitly doesn't do this, so why should magiskboot do it too? This is a needless pitfall that is easily fixed by this PR; I don't understand the aversion to this change.

kerneltoast:
Here is proof of the garbage output for a repacked `vendor_kernel_boot.img`:
```
$ ./unpack_bootimg.py --boot_img boot-new.img 
Traceback (most recent call last):
  File "./unpack_bootimg.py", line 563, in <module>
    main()
  File "./unpack_bootimg.py", line 558, in main
    info = unpack_bootimg(args.boot_img, args.out)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./unpack_bootimg.py", line 486, in unpack_bootimg
    info = unpack_vendor_boot_image(image_file, output_dir)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./unpack_bootimg.py", line 431, in unpack_vendor_boot_image
    boot_img.read(VENDOR_RAMDISK_NAME_SIZE))[0].decode())
                                                ^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe0 in position 5: invalid continuation byte
```

vvb2060:
Check vendor_boot instead of kernel_size

kerneltoast:
> Check vendor_boot instead of kernel_size

This is a weaker heuristic because vendor ramdisks are optional. It's more robust to check kernel size because the kernel has ever only been located in `boot.img`. If the image did not have a kernel image inside of it, there is no scenario where a kernel image should be introduced after repacking.

vvb2060:
magiskboot uses the current directory as input, so not cleaning the directory before unpacking is undefined behavior, and I think you should fix the script. If the image did not have a kernel image inside of it, there is no scenario where a kernel image should be in the directory.
However, I also agree that invalid format files should not be output.
cc @topjohnwu

aviraxp:
I am for this pr as well.

vvb2060:
https://github.com/topjohnwu/Magisk/commit/53d7cbc11bbb4e417707653db98a916ec2697f7e done

kerneltoast:
Moving one if-statement around was worse than adding a blurb to the documentation indicating that magiskboot might generate a garbage boot image?

PR #8385, opened 2024-09-11
Remove dtb command from magiskboot

yujincheng08:
They are no longer useful.


scripts/boot_patch.sh
```diff
@@ -195,18 +195,6 @@ rm -f ramdisk.cpio.orig config *.xz
 # Binary Patches
 #################
 
-for dt in dtb kernel_dtb extra; do
-  if [ -f $dt ]; then
-    if ! ./magiskboot dtb $dt test; then
```
vvb2060, line 200: This test should be safe to remove after [2 years](https://github.com/topjohnwu/Magisk/releases/tag/v25.0). I believe affected users will restore the original boot after bootloop and try again.

Kimmon322 (commented):
Leaves 30

broshop13:
No

PR #8406, opened 2024-09-21
Check sepolicy database version in add_xperm_rule

### LINKED ISSUE #8344: libsepol.avtab_write_item: policy version 29 does not support ioctl extendedpermissions rules and one was specified
Device: Quantum Go Q2
Android version: 7.1.1
Magisk version name: 8e82113b
Magisk version code: 27007

Following the [last issue](https://github.com/topjohnwu/Magisk/issues/8343), I decided to update Magisk through recovery, but after that the root was lost. The hidden app says that the root connection was lost and when restoring the default app it says that root is not available. To record the bootlog I decided to do a clean reinstall. For some reason the logcat disconnects during boot so I recorded a second one as soon as possible. When returning to TWRP, I noticed that the /cache/magisk.log file is not created so I imagine that Magisk is failing at a very early stage.
[bootlog-magisk27007-clean-install.zip](https://github.com/user-attachments/files/16776613/bootlog-magisk27007-clean-install.zip)
[magisk27007-patched-boot.zip](https://github.com/user-attachments/files/16776615/magisk27007-patched-boot.zip)
[stockboot.zip](https://github.com/user-attachments/files/16776616/stockboot.zip)

canyie:
Fix #8344

### DIFF
--- a/native/src/sepolicy/sepolicy.cpp
+++ b/native/src/sepolicy/sepolicy.cpp
@@ -272,7 +272,7 @@ bool sepol_impl::add_rule(const char *s, const char *t, const char *c, const cha
 
 void sepol_impl::add_xperm_rule(type_datum_t *src, type_datum_t *tgt, class_datum_t *cls, const Xperm &p, int effect) {
     if (db->policyvers < POLICYDB_VERSION_XPERMS_IOCTL) {
-        LOGE("policy version %u does not support ioctl extended permissions rules\n", db->policyvers);
+        LOGW("policy version %u does not support ioctl extended permissions rules\n", db->policyvers);
         return;
     }
     if (src == nullptr) {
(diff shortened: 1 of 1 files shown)


native/src/sepolicy/sepolicy.cpp
```diff
@@ -271,6 +271,10 @@ bool sepol_impl::add_rule(const char *s, const char *t, const char *c, const cha
 #define ioctl_func(x) (x & 0xFF)
 
 void sepol_impl::add_xperm_rule(type_datum_t *src, type_datum_t *tgt, class_datum_t *cls, const Xperm &p, int effect) {
+    if (db->policyvers < POLICYDB_VERSION_XPERMS_IOCTL) {
+        LOGE("policy version %u does not support ioctl extended permissions rules\n", db->policyvers);
```
vvb2060, line 275:
```suggestion
        LOGW("policy version %u does not support ioctl extended permissions rules\n", db->policyvers);
```

canyie:
The previous behavior was to abort in this case, not sure whether we should abort or not... I don't think this should be a fatal error, so this CL changes it to not abort... Should be okay.

vvb2060 (approved):

PR #8416, opened 2024-09-23
flash module: ignore META-INF

vvb2060:
sync with `magisk --install-module`

### DIFF
--- a/app/core/src/main/java/com/topjohnwu/magisk/core/tasks/FlashZip.kt
+++ b/app/core/src/main/java/com/topjohnwu/magisk/core/tasks/FlashZip.kt
@@ -47,18 +47,12 @@ open class FlashZip(
             }
         }
 
-        val isValid = try {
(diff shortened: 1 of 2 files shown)


Amdre-afk:
[magisk_log_2024-09-29T00.13.00.log](https://github.com/user-attachments/files/17178331/magisk_log_2024-09-29T00.13.00.log)

Kimmon322 (commented):
Leave 29

salvogiangri (commented):
Commit d33b077a13b34691a7f1fd3643336e09383ced78 introduced a regression that causes modules which erroneously use a _very old_ version of the [module_installer.sh](https://github.com/topjohnwu/Magisk/blob/master/scripts/module_installer.sh) script to not flash properly, it would be great to see this merged with the next release since it solves the issue

Jpac7-ai (commented):
Magik

topjohnwu (approved):

PR #8417, opened 2024-09-23
install_module: simplify script

vvb2060:


Kimmon322 (commented):
Leave 30

salvogiangri (commented):
Commit d33b077a13b34691a7f1fd3643336e09383ced78 introduced a regression that causes modules which erroneously use a _very old_ version of the [module_installer.sh](https://github.com/topjohnwu/Magisk/blob/master/scripts/module_installer.sh) script to not flash properly, it would be great to see this merged with the next release since it solves the issue

topjohnwu (approved):

PR #8475, opened 2024-10-15
Revert "init: support 2SI devices with skip_initramfs"

vvb2060:
This reverts commit b38fd1ca5f15ed22858c7737879c38d360e78eac.


aviraxp:
I think this is ready for merge.

vvb2060:
We need to detect /dev/root at runtime and then write it to config in boot.img, which is the only effective way to detect legacy SAR.

topjohnwu:
We'll find another way to detect the scenario, let's revert the behavior back to what it was

topjohnwu (approved):

PR #8515, opened 2024-10-27
App: Add the ability to hide the manager app with a non-random package name

dptsolutions:
I've encountered an instance where a root checker looking for Magisk determines that a "random" package ID like the ones used to hide the manager app are a sign of modification, and the app using the checker will refuse to work.

This PR modifies the hide app process, to allow the user to choose the package ID they want the hidden app to use, along with the custom app name. This is done by:

1. Adding a new EditText to the dialog, to allow the user to enter the package ID. If the field is left blank (the default) the hide process will use a random package ID as before
2. The new EditText has validation checks for a valid package ID
3. When the user goes to hide the app w/ a custom package ID, I check to ensure the package ID isn't already in use on the system, by quering PackageManager. If it is in use (or the check fails for some reason), an error dialog is displayed and the hide process is aborted

### DIFF
--- a/app/apk/src/main/java/com/topjohnwu/magisk/ui/settings/SettingsItems.kt
+++ b/app/apk/src/main/java/com/topjohnwu/magisk/ui/settings/SettingsItems.kt
@@ -5,7 +5,10 @@ import android.content.res.Resources
 import android.os.Build
 import android.view.LayoutInflater
 import android.view.View
+import android.view.View.OnFocusChangeListener
 import androidx.databinding.Bindable
+import com.google.android.material.textfield.TextInputEditText
+import com.google.android.material.textfield.TextInputLayout
 import com.topjohnwu.magisk.BR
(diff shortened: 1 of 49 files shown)


vvb2060:
If random applicationId will be found, then specific applicationId will also be found.

dptsolutions:
@vvb2060 - what am I missing when it comes to detection in this case? Assuming that they are using some sort of algo that looks for nonsensical package names that aren't words, if a user sets a package name that looks more normal (ie com.myapp.fun, or org.commonname), that would be more difficult to detect, no? This pr does not force every hidden install to the same id, it allows the end user of magisk manager to enter one themselves during the hiding process. 

The only thing I can think of to combat this is that a detector could scrape the major stores for legit package id's and if they see a package name that isn't recognized, flag it. But that would take actual resources on their end to maintain that package name db (rather than an algorithm in the detector to decide a package name is random) and probably throw lots of false positives.

PR #8556, opened 2024-11-14
Replaced HOME_CSC in install.md with USERDATA

LNLenost:
In the last version of Odin HOME_CSC was replaced with USERDATA so I fixed it


NBruderman:
Maybe keep both the old and the new names? For the sake of those who still use the old version

salvogiangri:
USERDATA tar only exists on a few Samsung firmware packages (mostly USA ones).

LNLenost:
> USERDATA tar only exists on a few Samsung firmware packages (mostly USA ones). This is also wrong because flashing that **does** format your data.

Sorry, didn't knew. Close this issue then.

PR #8600, opened 2024-12-05
Run app in root mode only if root solution is Magisk

aviraxp:
Current code assumes device is rooted with Magisk, so we need to enforce it or people will ask "why magisk cannot be installed with other root solutions, i have already gave it root".


vvb2060:
Installing magisk with other su is a supported use case‍

PR #8610, opened 2024-12-07
Support systemlessly deleting files or folders

aviraxp:
After we refactor the magic mount and always mount folder as tmpfs, we can easily support deleting files or folders now. We recognize dummy devices with major number 0 and minor number 0 as an indicator for removing files and folders. This indicator is borrowed from overlayfs.


aviraxp:
Diff to og pr: Fix confusing (entry->d_type & DT_BLK) == DT_CHR logic

aviraxp:
move to https://github.com/topjohnwu/Magisk/pull/6736

PR #8630, opened 2024-12-13
Introduce instrumentation tests

topjohnwu:


scripts/avd_magisk.sh
```diff
@@ -66,7 +66,7 @@ fi
 
 # Stop zygote (and previous setup if exists)
 magisk --stop 2>/dev/null
-stop zygote
+stop
```
vvb2060, line 69: This also stop adbd, will terminate script execution in interactive shell .

topjohnwu: The test will not properly execute if only using `stop zygote`

PR #8631, opened 2024-12-13
Update module installer guide about META-INF

aviraxp:
This syncs with FlashModule behavior.


vvb2060 (changes requested):
either delete all or keep all.

aviraxp:
Keep it now as-is to avoid breaking module itself's verifying scripts.

vvb2060 (approved):

Issue #8635, opened 2024-12-14
Critical Security Gap


OmarAwjy:
Device:Samsung Galaxy S21
Magisk version name:28101
Magisk version code:28101

Magisk "User Authentication" is made for protecting root access when user's phone is not in his hands, When we enable "User Authentication" all superuser requests should be assigned with user's password, but unfortunately, Magisk "Automatic Response" option completely conflicts with it, even if "User Authentication" is enabled, the "Automatic Response" option state can be changed without "User Authentication", if you changed it to "Grant", it won't prompt for User's password, and, it will grant all superuser requests for apps without User Authentication even if it's enabled, I'm asking The Magisk Team to fix this user security problem ASAP, Thanks.

EDIT : I strongly recommend to add the user's password protection for the entrie Magisk Manager app, this will protect unauthorised modules to be installed or removed, as some modules can remove the whole User's password.

akano12:
If someone have physical access to an asset, whatever protections and measure you take become useless. 


Let's not dive deep into how someone with physical access can uninstall then reinstall magisk app to reset your preferences, or flash any binary on your phone, or hack nasa using html.

Tldr, pointless

OmarAwjy:
> If someone have physical access to an asset, whatever protections and measure you take become useless.
> 
> Let's not dive deep into how someone with physical access can uninstall then reinstall magisk app to reset your preferences, or flash any binary on your phone, or hack nasa using html.
> 
> Tldr, pointless

Already thought about this before you and have the plan, Magisk will be a system app (i'll systemize it) to prevent uninstallation, and, will chmod 444 it, to prevent updates or downgrades, but this security issue need to be solved first, and yes, whatever protections and measure i take (if i can realize the security concept) are useful.

canyie:
Hello,
We have taken a look at this issue and determined that this is not a security vulnerability.
Here's our reasoning: While a physical attacker can bypass user interaction requirements, it is also possible for physical attackers to reboot the device into bootloader mode, install an image with malicious‌ programs then fully compromise and control the device. "User Authentication" is a best-effort protection measure, and does not claim to be safe against physical attacks. It is also almost impossible to "truly" protects an unlocked device from physical attacks, as most users have granted root access to other apps without setting an expiration time, and some of which can also be abused to execute malicious‌ commands or scripts with root privileges. 
This issue will be considered as a feature enhancement request, and is not enough to be considered as a security vulnerability. 
[...]

PR #8637, opened 2024-12-15
Perform authentication if needed for AutomaticResponse setting

aviraxp:
https://github.com/topjohnwu/Magisk/issues/8635


vvb2060:
This doesn't make sense, an attacker can clear the app data

topjohnwu (approved):

PR #8675, opened 2025-01-01
Add resetprop option to avoid triggering system_property_wait()

aviraxp:


native/src/core/resetprop/resetprop.cpp
```diff
@@ -167,6 +171,8 @@ static int set_prop(const char *name, const char *value, PropFlags flags) {
     if (pi != nullptr) {
         if (flags.isSkipSvc()) {
             ret = __system_property_update(pi, value, strlen(value));
+        } else if (flags.isSkipWait()) {
+            ret = __system_property_update(pi, value, strlen(value), true);
```
canyie, line 175:
```suggestion
        if (flags.isSkipWait()) {
            ret = __system_property_update(pi, value, strlen(value), true);
        } else if (flags.isSkipSvc()) {
            ret = __system_property_update(pi, value, strlen(value));
```

topjohnwu:
@aviraxp File a PR to https://github.com/topjohnwu/system_properties first

vvb2060:
What does this do?

yujincheng08 (changes requested):
To skip triggering wait, you also need to skip the following line:

https://github.com/topjohnwu/system_properties/blob/cabe129f5472ca10af82f3147fd8916d45b7d74e/system_properties.cpp#L316

https://github.com/topjohnwu/system_properties/blob/cabe129f5472ca10af82f3147fd8916d45b7d74e/system_properties.cpp#L325

PR #8729, opened 2025-01-22
Expose safe mode option on 28+

aviraxp:


vvb2060:
https://github.com/topjohnwu/Magisk/blob/95d1e69d8e21b7258dfe54e60d146a8d220991c2/app/core/src/main/java/com/topjohnwu/magisk/core/Const.kt#L26-L30

topjohnwu (approved):

PR #8736, opened 2025-01-26
Introduce post-mount stage scripts

aviraxp:
Same as post-fs-data mode, only difference is that scripts run after magic mount completes. This provides a stage where scripts can be instantly executed after module mount is ready to keep a clean memory map/mount namespace for new exec.

### DIFF
--- a/docs/details.md
+++ b/docs/details.md
@@ -51,6 +51,9 @@ SECURE_DIR=/data/adb
 # Folder storing general post-fs-data scripts
 $SECURE_DIR/post-fs-data.d
 
+# Folder storing general post-mount scripts
+$SECURE_DIR/post-mount.d
+
 # Folder storing general late_start service scripts
 $SECURE_DIR/service.d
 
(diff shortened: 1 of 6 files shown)


vvb2060:
why not fork on post-fs-data?

aviraxp:
How can i know magic mount is completed?

vvb2060:
post-fs-data done

aviraxp:
This should work as magic mount is now atomic,  but the point is providing additional stage without introducing more mounts. This stage is after magic mount and before zygote starts, making it a good stage to execute app_process, especially removing the need to run it in a standalone namespace.

vvb2060:
The magic mount is atomic, so app_process can be exec at any time. If you don't want to add additional file, you can check /system/bin/magisk

aviraxp:
> The magic mount is atomic, so app_process can be exec at any time. If you don't want to add additional file, you can check /system/bin/magisk

What about A10-？

yujincheng08:
you can `poll` on `/proc/1/mounts`

PR #8742, opened 2025-01-29
Move more code into Rust

topjohnwu:


aviraxp:
Can we have a new canary before this commit?

topjohnwu:
> Can we have a new canary before this commit?

Sure

PR #8751, opened 2025-02-02
Implement su_daemon in Rust

topjohnwu:


native/src/core/lib.rs
```diff
@@ -205,11 +204,19 @@ pub mod ffi {
...
         #[namespace = "rust"]
         fn daemon_entry();
     }
 
+    // Default constructors
+    extern "Rust" {
+        #[Self = SuRequest]
+        #[cxx_name = "New"]
+        fn default() -> SuRequest;
+    }
```
yujincheng08, line 218: May use https://github.com/dtolnay/cxx/pull/1430/commits/0e8a115ca4b82b39897bfde071336c3b12c23a7a for c++ side constructor

PR #8753, opened 2025-02-02
Define SuRequest::default as c++ constructor

yujincheng08:


yujincheng08:
Since it's not getting merge, never mind.

PR #8755, opened 2025-02-03
Store Utf8CStr as str

yujincheng08:


topjohnwu:
Why?

yujincheng08:
storing it as str semantically indicates its utf8 validated, while [u8] has no such semantic.

yujincheng08:
Since it's not getting merge, never mind.

PR #8758, opened 2025-02-03
Fix zygisk module load

yujincheng08:


topjohnwu:
I prefer fixing it in zygisk instead of module collection, resolved in 7f6c9e84117532534cef07f330a73482ff5b1c7f

topjohnwu:
I prefer fixing it in zygisk instead of module collection, resolved in 7f6c9e84117532534cef07f330a73482ff5b1c7f

PR #8774, opened 2025-02-08
Fix Utf8CString and get_context

yujincheng08:


aviraxp:
Why not just use Vector<u8>for xattr

yujincheng08:
Utf8CString is a wrapper of `Vec<u8>` with null-termination and utf8 validation.

aviraxp:
But we don't need to care about the content of con here

topjohnwu:
Replaced with #8808

PR #8827, opened 2025-03-02
get_secontext: ignore ENODATA

vvb2060:
https://man7.org/linux/man-pages/man2/getxattr.2.html
ENODATA: The named attribute does not exist, or the process has no access to this attribute.
ENOTSUP: Extended attributes are not supported by the filesystem, or are disabled.

ignore ENOTSUP too?

### DIFF
--- a/native/src/base/files.rs
+++ b/native/src/base/files.rs
@@ -733,10 +733,12 @@ impl FsPath {
                 XATTR_NAME_SELINUX.as_ptr().cast(),
                 con.as_mut_ptr().cast(),
                 con.capacity(),
-            )
-            .check_os_err()?;
+            );
             if sz < 1 {
                 con.clear();
+                if *errno() != libc::ENODATA {
+                    return Err(io::Error::last_os_error());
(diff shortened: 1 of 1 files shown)


native/src/base/files.rs
```diff
@@ -840,9 +842,14 @@ pub fn fd_get_attr(fd: RawFd) -> io::Result<FileAttr> {
                 XATTR_NAME_SELINUX.as_ptr().cast(),
                 attr.con.as_mut_ptr().cast(),
                 attr.con.capacity(),
-            )
-            .check_os_err()?;
-            attr.con.set_len((sz - 1) as usize);
+            );
+            if sz < 1 {
+                if *errno() != libc::ENODATA {
```
yujincheng08, line 847:
```suggestion
                attr.con.clear();
                if *errno() != libc::ENODATA {
```

vvb2060: where write?

topjohnwu (approved):

PR #8837, opened 2025-03-07
Update ONDK to r28.3

topjohnwu:


yujincheng08:
Also update edition?

https://github.com/topjohnwu/Magisk/blob/28d86a34541bcecf23517f74d7e52893a58b867a/native/src/base/Cargo.toml#L4

topjohnwu:
> Also update edition?
> 
> https://github.com/topjohnwu/Magisk/blob/28d86a34541bcecf23517f74d7e52893a58b867a/native/src/base/Cargo.toml#L4

This will be done in a separate PR

PR #8840, opened 2025-03-07
Introduce FsPathFollow

topjohnwu:
Make sure all operations of FsPath do not follow symlinks, and provide a way to explicitly switch over to a set of operations that DO follow symlinks by FsPath::follow_link.


saudalic (commented):
Perfect

Issue #6762, opened 2023-03-19
Fix terminal when using pipes


WandersonKalil:
`su -c busybox ps | grep su` will show unexpected  output on Terminal Emulator because we are calling `set_stdin_raw()` on the wrong side of the connection: we are setting the attributes on the Terminal's STDIN instead of the opened PTY's STDIN.

TODO: move `set_stdin_raw()` from `pts.cpp` into `su_daemon.cpp` after `close(client);` so that we keep the Terminal's STDIN intact.

Device: s20 FE
Android version: 12
Magisk version name: 981ccab 
Magisk version code: 25210

Issue #2425, opened 2020-02-09
Scripts with pipes won't run normally


eebssk1:
Recently I found one of my script didn't work at all.
Here  is another simple script i have with the problem

~~~~
#!/system/xbin/bash
sleep 6

bdev=zram0
dev=/sys/block/$bdev
ddev=/dev/block/$bdev
TM=$(free | grep Mem | cut -F 2)
UM=$(expr $TM / 100 \* 20)

if [ ! -e $ddev ]
then
exit
fi

echo 1 > $dev/reset

echo zstd > $dev/comp_algorithm
echo 2 > $dev/max_comp_streams
echo 64 > $dev/queue/read_ahead_kb
echo 1 > $dev/queue/add_random

echo $UM > $dev/disksize

mkswap $ddev
sleep 1
swapon -d $ddev

echo 1 > $dev/compact

sysctl -w vm.swappiness=55

if [ x$@ != x1 ]
then
/data/adb/service.d/swapfix.sh 1
fi
~~~~

And the following is strace log (not full,only shows the problem)
~~~~
Trace of process 11995 - /system/xbin/bash /data/adb/service.d/swap.sh 1
strace: Process 11995 attached
wait4(-1, [{WIFEXITED(s) && WEXITSTATUS(s) == 0}], 0, NULL) = 11996
rt_sigaction(SIGINT, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, {sa_handler=0x5f20bd054rt_sigprocmask(SIG_SETMASK, [], [CHLD], 8) = 0
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=11996, si_uid=0, si_status=0, siwait4(-1, 0x7fcc01ec24, WNOHANG, NULL)  = -1 ECHILD (No child processes)                  rt_sigreturn({mask=[]})                 = 0
read(255, "\nbdev=zram0\ndev=/sys/block/$bdev\nddev=/dev/block/$bdev\nTM=$(free | grep Memrt_sigprocmask(SIG_BLOCK, NULL, [], 8)  = 0
pipe2([3, 4], 0)                        = 0
[...]

Issue #1463, opened 2019-05-16
`echo id | su` doesn't work as expected


brbsix:
To illustrate the issue:

    $ echo id | su

Expected output:

    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0

Actual output:

    id

---
With SuperSU, I found `su -c COMMAND` to be very unreliable as it would frequently truncate the command's output or incorrectly report it's exit status. [It was instead recommended to create an interactive shell and pipe commands to it.][0]

As such, I have a large number of scripts that utilize the following technique:

    $ su <<<id
    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0

    $ su <<EOF
    echo 'This is the beginning of a long series of commands.'
    id
    EOF
    This is the beginning of a long series of commands.
    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0

Curiously, this doesn't work with MagiskSU. Rather than execute the commands that it receives via stdin, it prints them to stdout.

---

I've encountered quite a few inconsistencies however the issue generally seems to arise when either stdout or stderr are connected to a terminal.

    $ su <<<'touch the_command_was_clearly_run' &>/dev/null
    $ ls
    the_command_was_clearly_run

    $ out=$(su <<<id 2>/dev/null)
    $ echo "$out"
    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0

    $ out=$(su <<<cmd_does_not_exist 2>&1)
    $ echo "$?"
    127
    $ echo "$out"
    /system/bin/sh: <stdin>[1]: cmd_does_not_exist: not found

I'm not sure why, but the following typically (but not always) seems to work:

    $ (sleep 0.1; echo id) | su
    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0
    $ echo "$?"
    129

This also seems to work:

    $ socat - EXEC:su <<<id
    uid=0(root) gid=0(root) groups=0(root) context=u:r:magisk:s0

[0]: https://su.chainfire.eu/#how-call

topjohnwu:
@brbsix Magisk's `su -c` simply passes everything after the `-c` to `sh -c` with root permissions, there is no difference in how it is handled, so at least in Magisk's case the output will not be truncated.

What are you using this for? For an app? If that's the case,  would suggest you check out https://github.com/topjohnwu/libsu for that matter

mirfatif:
[...]

PR #8848, opened 2025-03-13
add "raw" flag that passes STDIN to a superuser command as-is

GamerGirlandCo:
related issues:
#2425 #6762 #1463

### DIFF
--- a/native/src/core/su/su.cpp
+++ b/native/src/core/su/su.cpp
@@ -41,6 +41,7 @@ int quit_signals[] = { SIGALRM, SIGABRT, SIGHUP, SIGPIPE, SIGQUIT, SIGTERM, SIGI
     "Usage: su [options] [-] [user [argument...]]\n\n"
     "Options:\n"
     "  -c, --command COMMAND         Pass COMMAND to the invoked shell\n"
+    "  -r                            Pipe STDIN to COMMAND as-is"
     "  -g, --group GROUP             Specify the primary group\n"
     "  -G, --supp-group GROUP        Specify a supplementary group.\n"
     "                                The first specified supplementary group is also used\n"
@@ -119,6 +120,7 @@ int su_client_main(int argc, char *argv[]) {
             strcpy(argv[i], "-M");
     }
 
+    bool is_raw = false;
     while ((c = getopt_long(argc, argv, "c:hlmps:VvuZ:Mt:g:G:", long_opts, nullptr)) != -1) {
         switch (c) {
             case 'c': {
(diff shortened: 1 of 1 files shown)


vvb2060:
In the Android world, the root shell is not the same thing as a local shell; it is a remote shell.
Think of it more like an SSH connection or [run0](https://www.freedesktop.org/software/systemd/man/257/run0.html)

vvb2060:
Can you give some test cases?

native/src/core/su/su.cpp
```diff
@@ -220,9 +225,9 @@ int su_client_main(int argc, char *argv[]) {
 
     // Determine which one of our streams are attached to a TTY
     int atty = 0;
-    if (isatty(STDIN_FILENO))  atty |= ATTY_IN;
-    if (isatty(STDOUT_FILENO)) atty |= ATTY_OUT;
-    if (isatty(STDERR_FILENO)) atty |= ATTY_ERR;
+    if (isatty(STDIN_FILENO) && !is_raw)  atty |= ATTY_IN;
+    if (isatty(STDOUT_FILENO) && !is_raw) atty |= ATTY_OUT;
+    if (isatty(STDERR_FILENO) && !is_raw) atty |= ATTY_ERR;
```
yujincheng08, line 230:
```suggestion
    if (isatty(STDERR_FILENO) && !is_raw) atty |= ATTY_ERR;
```

yujincheng08, line 230:
```suggestion
    if (!is_raw && isatty(STDIN_FILENO))  atty |= ATTY_IN;
    if (!is_raw && isatty(STDOUT_FILENO)) atty |= ATTY_OUT;
    if (!is_raw && isatty(STDERR_FILENO)) atty |= ATTY_ERR;
```

GamerGirlandCo:
> In the Android world, the root shell is not the same thing as a local shell; it is a remote shell.
> Think of it more like an SSH connection or [run0](https://www.freedesktop.org/software/systemd/man/257/run0.html)

you can still pipe stdin to ssh sessions, though. it's why you can use programs like `tar` or `dd` over ssh without needing to copy the archive over/use a temporary file.

vvb2060:
I think it's weird to switch via arguments, is there a way to detect the case where stdin is not a tty?

vvb2060:
> sure. where exactly should i put the tests? do i add them to test_common.sh?

You can reply directly, of course adding unit tests is the best way.

GamerGirlandCo:
> I think it's weird to switch via arguments, is there a way to detect the case where stdin is not a tty?

there are some cases where a TTY will be opened (albeit in raw mode), such as when using `adb exec-in` or `adb exec-out`. this isn't bad if you're piping plain text, but it does become a problem when you're trying to pipe binary data with "unprintable" characters.

yujincheng08:
See if ab9086534e16c37e244008934368754f0e67a481 has fixed this issue.

PR #8905, opened 2025-03-29
Refactor elf-cleaner with rust

yujincheng08:


topjohnwu:
@yujincheng08 switch over to a Rust implementation that also works on Windows. See #8915

PR #8927, opened 2025-04-09
Move pts to rust, and avoid using thread

yujincheng08:
After this, only one thread when running `su`:

```
$ su
# ls /proc/$(pidof -s su)/task/ | wc -l
1
```

Also fix:
- wrong format of `su | cat`
- wrong output of `echo id | su`
- cannot `crtl + z` when `su -c sleep 100`


vvb2060 (approved):

aviraxp:
@topjohnwu I am not sure if you you have noticied, but last patch in this series effectively breaks su -c behavior that some apps have been relying on since 2012 (appending tty is koush's su default behaviour since then). Now these apps cannot react to external signals.

Yes, I know it is different from linux su behaviour,  but su is not posix, there isn't a standard for this. At least, this behavior change should be noticed in change log.

native/src/core/su/pts.cpp
```diff
@@ -1,294 +0,0 @@
...
-    if (tcgetattr(STDIN_FILENO, &termios) < 0) {
-        return -1;
-    }
-
-    old_stdin = termios;
-
-    cfmakeraw(&termios);
-
-    if (tcsetattr(STDIN_FILENO, TCSAFLUSH, &termios) < 0) {
-        // https://blog.zhanghai.me/fixing-line-editing-on-android-8-0/
```
vvb2060, line 143: I think we should keep this note

Clawcore64:
Круто!

PR #8969, opened 2025-04-28
Support patch boot on device using build.py

Howard20181:


phaokip (commented):
scripts/patch_boot.sh

topjohnwu:
I don't plan to maintain yet another way of installing Magisk, sorry :/.
Let's maybe cleanup the existing installing mechanism before introducing this.
Also, taking a closer look, I don't think calling boot_patch.sh without an argument can work?

PR #8988, opened 2025-05-02
build.py: add cargo to PATH

vvb2060:
Fix build without rust development environment

### DIFF
--- a/build.py
+++ b/build.py
@@ -223,6 +223,7 @@ def build_cpp_src(targets: set):
 def run_cargo(cmds):
     ensure_paths()
     env = os.environ.copy()
+    env["PATH"] = f'{env["PATH"]}{os.pathsep}{rust_sysroot}{os.sep}bin'
     env["RUSTUP_TOOLCHAIN"] = str(rust_sysroot)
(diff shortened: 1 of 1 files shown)


topjohnwu:
I actually explicitly removed it because calling `cargo`/`rustc` directly behaves differently compared to calling the rustup proxies. For instances, some shared library paths are not set accordingly.

vvb2060:
update [doc](docs/build.md)? If rustup is not installed, `build.py all` will fail.

build.py
```diff
@@ -223,6 +223,7 @@ def build_cpp_src(targets: set):
 def run_cargo(cmds):
     ensure_paths()
     env = os.environ.copy()
+    env["PATH"] = f'{env["PATH"]}{os.pathsep}{rust_sysroot}{os.sep}bin'
```
yujincheng08, line 226: Only the following works on my machine:
```suggestion
    env["PATH"] = f'{rust_sysroot}{os.sep}bin{os.pathsep}{env["PATH"]}'
```

aviraxp:
merge？

PR #8999, opened 2025-05-08
Migrate module mounting to Rust

topjohnwu:


yujincheng08 (approved):

native/src/core/module.rs
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    // First find whether /system/bin exists
+    let bin = system.children().and_then(|c| c.get_mut("bin"));
+    if let Some(FsNode::Directory { children }) = bin {
+        inject(children);
+        return;
+    }
+
+    // If /system/bin node does not exist, use the first suitable directory in PATH
```
aviraxp, line 374: Plz keep old logic, ensure only magisk supported mount path is used

vvb2060, line 374: Maybe we can use `/system/apex/com.android.runtime/bin`, all custom ROMs are flattened apex.
https://source.android.com/docs/core/ota/apex#flattening-apex

PR #9008, opened 2025-05-10
su: better tty support

vvb2060:
> --pty, --pipe
Request allocation of a pseudo TTY for the run0 session (in case of --pty), or request passing the caller's STDIO file descriptors directly through (in case of --pipe). If neither switch is specified, or if both switches are specified, the mode will be picked automatically: if standard input, standard output and standard error output are all connected to a TTY then a pseudo TTY is allocated, otherwise the relevant file descriptors are passed through directly.

https://www.freedesktop.org/software/systemd/man/257/run0.html


yujincheng08 (changes requested):
This make `su -c sleep 10` impossible to `^Z` to move it to background.

PR #9017, opened 2025-05-13
su: pass all unknown options

vvb2060:
-c, -l and more are now handled by remote shell.
http://www.mirbsd.org/htman/i386/man1/mksh.htm
This allows users to pass more options to the remote shell, such as rcfile.

### DIFF
--- a/native/src/core/su/su.cpp
+++ b/native/src/core/su/su.cpp
@@ -40,26 +40,26 @@ int quit_signals[] = { SIGALRM, SIGABRT, SIGHUP, SIGPIPE, SIGQUIT, SIGTERM, SIGI
 
     fprintf(stream,
     "MagiskSU\n\n"
-    "Usage: su [options] [-] [user [argument...]]\n\n"
+    "Usage: su [options] [--] [user [argument...]]\n\n"
     "Options:\n"
-    "  -c, --command COMMAND         Pass COMMAND to the invoked shell\n"
-    "  -i, --interactive             Force pseudo-terminal allocation when using -c\n"
+    "  -s, --shell SHELL             Use SHELL instead of the default " DEFAULT_SHELL "\n"
+    "  -i, --interactive             Force pseudo-terminal allocation\n"
     "  -g, --group GROUP             Specify the primary group\n"
     "  -G, --supp-group GROUP        Specify a supplementary group\n"
     "                                The first specified supplementary group is also used\n"
     "                                as a primary group if the option -g is not specified\n"
     "  -Z, --context CONTEXT         Change SELinux context\n"
     "  -t, --target PID              PID to take mount namespace from\n"
+    "                                pid 0 means magisk global mount namespace\n"
     "  -d, --drop-cap                Drop all Linux capabilities\n"
-    "  -h, --help                    Display this help message and exit\n"
-    "  -, -l, --login                Pretend the shell to be a login shell\n"
     "  -m, -p,\n"
     "  --preserve-environment        Preserve the entire environment\n"
-    "  -s, --shell SHELL             Use SHELL instead of the default " DEFAULT_SHELL "\n"
     "  -v, --version                 Display version number and exit\n"
     "  -V                            Display version code and exit\n"
-    "  -mm, -M,\n"
-    "  --mount-master                Force run in the global mount namespace\n\n");
+    "  -h, --help                    Display this help message and exit\n\n"
+    "--: Force stop options parsing, and also stop when an unknown option is found\n"
+    "User: The user to switch to (default root), it can be name or uid\n"
+    "Argument: Pass it to the shell as is\n\n");
     exit(status);
 }
 
@@ -87,10 +87,9 @@ static void setup_sighandlers(void (*handler)(int)) {
 }
 
 int su_client_main(int argc, char *argv[]) {
+    opterr = 0;
     option long_opts[] = {
-            { "command",                required_argument,  nullptr, 'c' },
             { "help",                   no_argument,        nullptr, 'h' },
-            { "login",                  no_argument,        nullptr, 'l' },
             { "preserve-environment",   no_argument,        nullptr, 'p' },
             { "shell",                  required_argument,  nullptr, 's' },
             { "version",                no_argument,        nullptr, 'v' },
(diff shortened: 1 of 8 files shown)


native/src/core/su/su.cpp
```diff
@@ -224,7 +219,7 @@ int su_client_main(int argc, char *argv[]) {
     }
 
     // Determine which one of our streams are attached to a TTY
-    interactive |= req.command.empty();
+    interactive |= req.command.size() == 1;
```
yujincheng08, line 222: I recommend keeping `--shell` separated and leaving this `empty()`. Otherwise, `su some_commend` still errorously allocates tty (thus cannot `^Z`).

native/src/core/su/su.cpp
```diff
@@ -38,23 +38,20 @@ int quit_signals[] = { SIGALRM, SIGABRT, SIGHUP, SIGPIPE, SIGQUIT, SIGTERM, SIGI
     "MagiskSU\n\n"
     "Usage: su [options] [-] [user [argument...]]\n\n"
     "Options:\n"
-    "  -c, --command COMMAND         Pass COMMAND to the invoked shell\n"
-    "  -i, --interactive             Force pseudo-terminal allocation when using -c\n"
+    "  -s, --shell SHELL             Use SHELL instead of the default " DEFAULT_SHELL "\n"
+    "  -i, --interactive             Force pseudo-terminal allocation\n"
```
yujincheng08, line 42: since there's a breaking change on `-m`, how about making tty allocation `-t` as `ssh`, `docker`, and changing attach mount namespace to `-m`?

yujincheng08:
I also recommend adding `--` support to it, like other UNIX commands, to pass all stuff after `--` to the inner command.

native/src/core/su/su.cpp
vvb2060: Tested `su sleep 20` will not allocate tty

native/src/core/su/su.cpp
vvb2060: No options have changed

vvb2060:
Already supported --

native/src/core/su/su.cpp
yujincheng08: I mean, only one command. For example, `su ls`.

yujincheng08:
Add `--` to help message?

native/src/core/su/su.cpp
vvb2060: Tested `su ls` will not allocate tty

native/src/core/su/su.cpp
yujincheng08: `--mount-master` vs `-t 0`?

vvb2060: They are the same, I removed the obsolete -mm help message.

vvb2060:
done

vvb2060:
done

selfmusing:
After [PR #9017](https://github.com/topjohnwu/Magisk/pull/9017) was merged into the Magisk Alpha (starting with v29.x), several apps that previously worked without issues are now failing to function properly.

This seems to be related to the recent changes in how su is handled, as reverting to an older Alpha build (pre-v29) or switching to the official Magisk release restores full functionality.

Examples of affected apps:

Valhalla Thor ([Play Store](https://play.google.com/store/apps/details?id=com.valhalla.thor))

Google Shortcuts Launcher v5.5 ([GitHub](https://github.com/WSTxda/Google-Shortcuts-Launcher/releases/))

Is there a recommended way to make these apps work with the latest magisk alpha (v29.x & above), or is this an intentional change that breaks compatibility?

Looking forward to any guidance.

vvb2060:
I recommend that these apps use [libsu](https://github.com/topjohnwu/libsu) instead of `su -c`

selfmusing:
> I recommend that these apps use [libsu](https://github.com/topjohnwu/libsu) instead of `su -c`

Screenshot_20250725-115321_MT Manager

Screenshot_20250725-115346_MT Manager

Thanks for the response.

After inspecting the `classes.dex` of both apps, I can confirm that they indeed rely on executing `su -c` commands directly to request root access, rather than using `libsu`.

I've attached screenshots showing the relevant code references for both apps.

This likely explains why their root functionality breaks on **Magisk Alpha v29.x** and above — due to the recent changes in how `su` is handled.

Is there any way to temporarily restore compatibility for such apps without needing them to update to `libsu`?

vvb2060:
Add quotes around the command, for example `su -c "ls /data"`

selfmusing:
> Add quotes around the command, for example `su -c "ls /data"`

I've change `su -c` to `su -s` on shortcuts launcher app and it worked.
But the other app ain't working even after using `su -c "cmd"`

Screenshot_20250725-131639_Thor

vvb2060:
not you, ask the app developer to fix it

native/src/core/su/su.cpp
```diff
@@ -176,34 +162,46 @@ int su_client_main(int argc, char *argv[]) {
...
-        if (const passwd *pw = getpwnam(argv[optind]))
+        if (const passwd *pw = getpwnam(argv[optind])) {
             req.target_uid = pw->pw_uid;
-        else
-            req.target_uid = parse_int(argv[optind]);
-        optind++;
+            optind++;
+        } else if (int uid = parse_int(argv[optind]); uid >= 0) {
+            req.target_uid = uid;
+            optind++;
```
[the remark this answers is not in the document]
vvb2060: Expected behavior

native/src/core/su/su.cpp
```diff
@@ -116,29 +115,16 @@ int su_client_main(int argc, char *argv[]) {
     }
 
     bool interactive = false;
+    string shell = DEFAULT_SHELL;
 
     int c;
-    while ((c = getopt_long(argc, argv, "c:hlimpds:VvuZ:Mt:g:G:", long_opts, nullptr)) != -1) {
+    while ((c = getopt_long(argc, argv, "+:himpds:VvuZ:Mt:g:G:", long_opts, nullptr)) != -1) {
```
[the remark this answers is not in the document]
vvb2060: b0c1a6f73a205fc31a727280550211f8ae94fbc2

topjohnwu:
@vvb2060 is this behavior backwards compatible? I'd prefer to preserve the existing behavior, and only extend the functionality to forward unknown options to the inner shell.

yujincheng08:
@topjohnwu difference:

```bash
$ # before
$ su -c echo 'hello'
hello
$ # after
$ su -c echo 'hello'
$
```

topjohnwu:
This changes the CLI API in a backwards incompatible way. I'll need more consideration before merging this PR.

vvb2060:
It is already compatible with the old commands. https://github.com/topjohnwu/Magisk/pull/9017/commits/b915da2495f0845720974a895c580e309511db15

hamzabhatti5253:
Change it

Issue #9050, opened 2025-05-19
Magisk 29 shows "Installed: N/A" if Superuser Access is set to "Disabled"


willindernet:
If you set ‘Superuser Access’ to ‘Disabled’ in the settings, Magisk is no longer recognised by the app. The app then displays ‘Installed: N/A’ is displayed in the app. It is then no longer possible to reset it, as ‘Superuser Access’ is then missing in the settings.
[See here](https://xdaforums.com/t/magisk-general-support-discussion.3432382/page-3003#post-90104508)

Device: Sony Xperia XZ2 compact
Android version: 15
Magisk version name: 29.0
Magisk version code: 29000

[magisk_log_2025-05-19T16.12.32.log](https://github.com/user-attachments/files/20298541/magisk_log_2025-05-19T16.12.32.log)

aviraxp:
regression https://github.com/topjohnwu/Magisk/commit/a786801141db8e4b70e91a9e9765fd411220f109#diff-9ce04ac990de2b35afaa13dadded38891ec078367ebe6fe8c785f8a99ad7c8e9

aviraxp:
Try ci build https://github.com/aviraxp/Magisk/commit/537da2291458e3d1ac8b43e00aa4534740ef5e6d

pndwal:
> Try ci build hhttps://github.com/[/pull/9051](https://github.com/topjohnwu/Magisk/pull/9051).

@avitaxp, _anyone_ can test this; 28xxx or 29xxx... I reproduced issue with 29001 by  disabling superuser access before taking Hide the Magisk app; That fails to complete, and when stub app is uninstalled, full app doesn't regain access...

I'll try your ci build when I have a minute if no-one else does... 👍

PR #9051, opened 2025-05-19
Ensure manager can always bypass su access policy

aviraxp:
This fixes https://github.com/topjohnwu/Magisk/issues/9050.

### DIFF
--- a/native/src/core/su/daemon.rs
+++ b/native/src/core/su/daemon.rs
@@ -241,6 +241,22 @@ impl MagiskD {
                 _ => uid,
             };
 
+            let mut access = RootSettings::default();
+            self.get_root_settings(eval_uid, &mut access)?;
+
+            // We need to talk to the manager, get the app info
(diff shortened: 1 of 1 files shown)


vsaucing:
Doesn't work on A10 Huawei MatePad T8.

topjohnwu (approved):

aviraxp:
works

gqqnbig:
I created the problem in version 29, and I upgraded to version 30.4, but the Installed label still shows N/A. How to fix it?

Kast555:
I ended up fixing it but I had to factory reset the entire phone. When I
say factory reset I'm talking about going into the recovery.  wiping all
user data and cache and then installing the upgraded magisk and rebooting.

gqqnbig:
@Kast555 Since you wiped all user data, it's close to factory-reseting the phone. So I reset the phone, and the problem disappears. In this sense, I can't really confirm whether this PR fixes the issue or not.

PR #9086, opened 2025-06-02
Mount magisk bins in directory with least files

aviraxp:


vvb2060:
I think SECONDARY_READ_ONLY_PARTITIONS is enough, sorting by number of files may select strange dir

aviraxp:
https://github.com/topjohnwu/Magisk/pull/9088

PR #9088, opened 2025-06-02
Refactor magisk bins injection logic

aviraxp:


aviraxp:
@yujincheng08 Can you add split_off/drain to Utf8CString?

aviraxp:
This commit is ready for review.

PR #9091, opened 2025-06-05
Avoid throwing error if chmod a symlink

aviraxp:
It is possible that we want to clone attributes of a regular file to a symlink. In this case, we don't need to
error out if
chmod fails. Just skip it.


aviraxp:
This happens if we bind mount a symlink on a regular file.

aviraxp:
Do we need to switch to https://android.googlesource.com/platform/bionic/+/3cbc6c6?

topjohnwu (approved):

Issue #9105, opened 2025-06-13
lsetxattr '/sbin/knox_changer' 'u:object_r:system_file:s0': Read-only file system (os error 30)


Hargriv:
Device: Samsung Note9 SM-N960F
Android version: 10 OneUI  2.5
Magisk version name: Canary 29001 
Magisk version code: b1ac47a0

[magisk_log_2025-06-13T13.21.39.log](https://github.com/user-attachments/files/20724855/magisk_log_2025-06-13T13.21.39.log)

aviraxp:
Should be fixed by https://github.com/topjohnwu/Magisk/commit/0b494ed7df43c16e1dbec5e6f9247f079ad38831

tlw00988:
em....
Android's root is mounted as read-only

Hargriv:
@tlw00988 and what will be the decision on this issue?

tlw00988:
> @tlw00988 and what will be the decision on this issue?

Sorry,i've no idea

yujincheng08:
expected

oOJohn6Oo:
@aviraxp I installed magisk version 30.2, still have this issue.

```
08-09 08:28:00.700 633 633 I : Magisk 30.2(30200) daemon started
08-09 08:28:00.701 633 633 I : * Device API level: 27
08-09 08:28:00.701 633 633 E : lsetxattr '/sbin/mkfs.f2fs' 'u:object_r:system_file:s0’: Read-only file system (os error 30)
08-09 08:28:00.701 633 633 E : lsetxattr '/sbin/charger' 'u:object_r:system_file:s0’: Read-only file system (os error 30)
08-09 08:28:00.701 633 633 E : lsetxattr '/sbin/busybox' 'u:object_r:system_file:s0’: Read-only file system (os error 30)
08-09 08:28:00.711 633 635 I : ** post-fs-data mode running
08-09 08:28:00.727 633 635 I : * Initializing Magisk environment
08-09 08:28:00.744 633 635 I : * Running post-fs-data.d scripts
08-09 08:28:00.746 633 635 E : opendir '/data/adb/modules_update’: No such file or directory (os error 2)
08-09 08:28:00.746 633 635 I : * Running module post-fs-data scripts
08-09 08:28:01.275 633 635 I : ** late_start service mode running
08-09 08:28:01.275 633 635 I : * Running service.d scripts
08-09 08:28:01.281 633 635 I : * Running module service scripts
[...]

PR #9112, opened 2025-06-16
Skip symlink in restore_tmpcon()

aviraxp:
If magisktmp is /sbin, there may exist files which is symlink to files in root dir. As root is RO, setcontext will fail and break iterating loop.


aviraxp:
regression https://github.com/topjohnwu/Magisk/issues/9105

PR #9115, opened 2025-06-17
Ignore set_context() error if magisktmp is /sbin

aviraxp:
recreate_sbin() will bind mount original files in /sbin to tmpfs /sbin, so we have no choice but just log here to let the loop continue.


aviraxp:
merge

Issue #9116, opened 2025-06-17
[Canary debug b1dc47a0 29001] Zygisk won't activate on Pixel 8a with GrapheneOS


hiviah:
Device: Pixel 8a (non-carrier, unlocked, brand new)
Android version: 15 (GrapheneOS 2025061500)
Magisk version name: 29001 canary debug
Magisk version code: b1dc47a0

No Magisk modules installed.

I've reinstalled system many times (20+) now (flash_all over then install Magisk, try turn on/off Zygisk, reboot after each change, repeat fresh flash) and tried hints mentioned in Issues (like using different LPosed, disabling vbmeta...did not make difference)

I've installed at first stable release of Magisk 29.0, then debug for this next attempt and this report - https://github.com/topjohnwu/Magisk/releases/tag/canary-29001 (b1dc47a0), tried to turn on Zygisk, but it always says **Zygisk: No**

OS used is GrapheneOS release https://grapheneos.org/releases#2025061500

Attaching logs and `init_boot.img` before patch at the bottom:

* logcat
* magisk log 
* `init_boot.img` used for patching before patching

One error I see in logs that looks maybe relevant is (magisk log):

```
[ro.boot.verifiedbooterror]: [ERROR_VERIFICATION]
[ro.boot.verifiedbootstate]: [green]
[ro.boot.verifyerrorpart]: [init_boot]
```

Second possibly relevant is unability to parse certificates, which means malformer ASN.1 DER/BER encoding:

```
06-17 13:21:10.721  1277  1277 I SystemServiceManager: Starting com.android.server.security.FileIntegrityService
06-17 13:21:10.722  1277  1437 I SystemConfig: Reading permissions from /system/etc/permissions/javax.obex.xml
06-17 13:21:10.723  1277  1277 E FileIntegrityService: Invalid certificate, ignored: java.security.cert.CertificateException: com.android.org.conscrypt.OpenSSLX509CertificateFactory$ParsingException: com.android.org.conscrypt.OpenSSLX509CertificateFactory$ParsingException: java.lang.RuntimeException: error:0c00006d:ASN.1 encoding routines:OPENSSL_internal:DECODE_ERROR
06-17 13:21:10.723  1277  1437 I SystemConfig: Reading permissions from /system/etc/permissions/com.android.mediadrm.signer.xml
[...]

Issue #7489, opened 2023-11-08
SystemUI does not work with Zygisk enabled


ubergeek77:
Device: Pixel 8 Pro
Android version: Android 14 (GrapheneOS)
Magisk version name: 26.4
Magisk version code: 26401 (aa6c7c15) (D)

---

I first reported this on a similar issue that has since been closed:

https://github.com/topjohnwu/Magisk/issues/7448#issuecomment-1784467986

I also reported that a certain build of Magisk with Zygisk improvements does not experience this issue:

https://github.com/topjohnwu/Magisk/issues/7448#issuecomment-1789929581

---

When Zygisk is enabled, SystemUI does not work properly. The system boots to the launcher, and apps can be opened and used, but the following elements are not loaded and therefore do not function:

- **System navigation** - the navigation bar does not show, regardless of if "3-button" or "gesture" navigation is used. Any kind of navigation, including gestures, is not possible. The user cannot go "back," "home," and the "recents" screen is not accessible.
- **Notification bar** - the notification bar is not visible, and the pull-down gesture to access quick settings does not function.
- **Power menu** - the standard power menu is not present. The system falls back to the safe mode power menu.
- **Lock screen** - the screen will simply turn off when the power button is tapped. If the power button is tapped again, the phone will resume in whatever app it was in, but without displaying the lock screen.

This issue occurs in both 26.3 and 26.4, the latter of which had some improvements to FD sanitization. This occurs **only with Zygisk enabled.** The broken SystemUI elements work again if Zygisk is disabled.

This issue ***does not occur*** on this build which appears to change the way Zygisk is injected:

https://github.com/topjohnwu/Magisk/actions/runs/6767188369

On this build, Zygisk works properly, and can be enabled safely without breaking SystemUI.

---

Boot logcat:
[boot-logcat.log](https://github.com/topjohnwu/Magisk/files/13290965/boot-logcat.log)

Magisk log:
[magisk_log_2023-11-07T19.12.26.log](https://github.com/topjohnwu/Magisk/files/13290948/magisk_log_2023-11-07T19.12.26.log)

osm0sis:
Interesting I don't have this issue on Pixel 8 Pro stock.

Unknown78:
I have similar issue on https://github.com/topjohnwu/Magisk/issues/7472#issuecomment-1800842969, I could boot, but my screen is black. I couldn't open Magisk to check whether Zygisk is successfully enabled or not.

[...]

PR #9118, opened 2025-06-17
README note about  commonly reported issue - Zygisk on various ROMs - why it can't even be enabled sometimes

hiviah:
Adding note about Zygisk not working/being enabled, since this is commonly reported issue, but is not mentioned anywhere in README, Install, FAQ (only that it may not always hide root, not about not being enabled at all, without explanation of reason or warning).

I think due to sheer amount of related issues with this it warrants note directly in README (it hopefully should save you time on the issues).

It states clear reason on why (unlike "won't implement/not planned") - based on https://github.com/topjohnwu/Magisk/issues/9116#issuecomment-2981385567 and https://github.com/topjohnwu/Magisk/issues/7489#issuecomment-1841767693

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -39,6 +39,11 @@ For installation issues, upload both boot image and install logs.<br>
 For Magisk issues, upload boot logcat or dmesg.<br>
 For Magisk app crashes, record and upload the logcat when the crash occurs.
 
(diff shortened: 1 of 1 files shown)


hiviah:
Hm.

Issue #9174, opened 2025-07-12
Unreliable magic mount on overlayfs causes side effects after rust rewrite


aviraxp:
Bind mounts on overlayfs require access to the lower layer. By default (unless override_creds=off is set), overlayfs checks whether the mounter (typically init) has permission to access files in the lower layer. If not, the operation fails. On some devices, due to incorrect OEM SELinux policies, files like init under /product/bin are inaccessible, causing bind mount failures in that directory. In the Rust rewrite, such failures are returned immediately instead of being silently ignored as before, which led to this issue.

Device: 1
Android version: 1
Magisk version name: 1
Magisk version code: 1

aviraxp:
@topjohnwu I think we have to let bind_mount return ok regardless of whether it really successes.

PR #9175, opened 2025-07-12
Treat bind mount failure same as C++ implementation

### LINKED ISSUE #9139: Zygisk fails to start
Device: Hotdog
Android version: 12
Magisk version name: 30.0
Magisk version code: 30000

Zygisk fails to start in this version asking to reboot for starting - but never succeed.
Zygisk-Next module solves the issue, however, the question is why built-in one does not work.

aviraxp:
This fixes #9139 and #9174.


aviraxp:
Not tested

topjohnwu (approved):

Issue #8835, opened 2025-03-05
[Feature Request] Support device whose ramdisk locates in `vendor_boot/init_boot.cpio`


NeutroGe:
Device: Xiaomi Redmi pad SE 8.7 wifi - codename "flare" - model number 24075RP89G - SOC Helio G85 (mediatek MT8786) - stock fastboot rom from https://xmfirmwareupdater.com/hyperos/flare/stable/OS2.0.1.0.VHXMIXM/ - bootloader unlocked with the official xiaomi unlocker tool.
```
flare:/dev/block/bootdevice # ls /dev/block/bootdevice/by-name/
boot_a       dtbo_b     gsort   logo_b    misc          nvram     persist          proinfo   sec1     super     vbmeta_system_a
boot_b       efuse      gz_a    md1img_a  mmcblk0       oops      persistbak       protect1  seccfg   tee_a     vbmeta_system_b
boot_para    expdb      gz_b    md1img_b  mmcblk0boot0  opconfig  preloader_a      protect2  spmfw_a  tee_b     vbmeta_vendor_a
bootloader1  ffu        lk_a    md_udc    mmcblk0boot1  opcust    preloader_b      rescue    spmfw_b  userdata  vbmeta_vendor_b
bootloader2  flashinfo  lk_b    mem       nvcfg         otp       preloader_raw_a  scp_a     sspm_a   vbmeta_a  vendor_boot_a
dtbo_a       frp        logo_a  metadata  nvdata        para      preloader_raw_b  scp_b     sspm_b   vbmeta_b  vendor_boot_b
```

Android version: 15 AP3A.240905.015.A2 Kernel 6.6.30-android15-8
Magisk version name:  b62835c
Magisk version code: 28102

Hello,

I have tried to root my tablet using magisk by patching the original boot.img from the stock firmware then flashing it and also flashing the original vbmeta.img with "--disable-verity --disable-verification". There are no errors whatsoever but when booting with the patched boot image the system is not rooted.

I have tried with the release versions magisk v28.1, with v27.0 and with the latest canary debug version listed above with the same result.

There are no init_boot.img or recovery.img from the stock fastboot rom, no recovery partition either (there is a recovery mode on the device to wipe the data but i dont know in which image or partition it is located).

Original boot.img: [boot.img.zip](https://github.com/user-attachments/files/19080380/boot.img.zip)

[...]

PR #9238, opened 2025-08-04
add vendor_boot patching for some odd devices

ukriu:
This fixes (or adds) #8835

### DIFF
--- a/app/core/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
+++ b/app/core/src/main/java/com/topjohnwu/magisk/core/tasks/MagiskInstaller.kt
@@ -82,9 +82,10 @@ abstract class MagiskInstallImpl protected constructor(
 
     private fun findImage(slot: String): Boolean {
         val bootPath = (
-            "(RECOVERYMODE=${Config.recovery} " +
+            "cd $installDir && " +
+            "RECOVERYMODE=${Config.recovery} " +
             "SLOT=$slot find_boot_image; " +
-            "echo \$BOOTIMAGE)").fsh()
+            "echo \$BOOTIMAGE").fsh()
         if (bootPath.isEmpty()) {
             console.add("! Unable to detect target image")
             return false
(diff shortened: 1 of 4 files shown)


ukriu:
The same method to detect this case has been used in https://github.com/tiann/KernelSU/pull/2650 and https://github.com/KernelSU-Next/KernelSU-Next/issues/541#issuecomment-2999593794  
And has been working perfectly fine

topjohnwu:
Thank you for your contribution. I'll need further thinking to make sure if there are better ways to detect this scenario.

ukriu:
> Thank you for your contribution. I'll need further thinking to make sure if there are better ways to detect this scenario.

Just a heads up, there is a motorola device which has a compatible ramdisk in vendor_boot as well as boot, the current check picks the boot partition for this device, but you still could remove one of the magiskboot unpacking steps to save time since both work for the device.  
  
As for the speed, I think its fine since it only takes less than a second from my testing. But it still is a bit messy so I'm looking forward to new ideas if you have them.  
  
Currently the devices which need this are the Pixel 6 family, some Pixel 7 devices (not sure), and about 10 to 15 Xiaomi Mediatek devices which have officially gotten an upgrade to Android 15; so it would be nice if this case is acknowledged

vvb2060:
No pixel device has this behavior, I think you might have made a mistake

ukriu:
> No pixel device has this behavior, I think you might have made a mistake

Pixel 6/6 Pro/6a, while I only have confirmed it with 6a.
You can grab the latest "bluejay" firmware and test for yourself

vvb2060:
I checked this https://dl.google.com/dl/android/aosp/bluejay-ota-bp2a.250705.008.a1-14519571.zip
Amazing, it seems that this is because the device uses the new layout but does not want to repartition. Is adding a ramdisk to boot.img invalid? It will overwrite the file with the same name in the vendor_boot ramdisk.

ukriu:
> Is adding a ramdisk to boot.img invalid? It will overwrite the file with the same name in the vendor_boot ramdisk.

That can be done and is being used here (https://github.com/KernelSU-Next/KernelSU-Next/commit/512f84504ec18b2e73ad5350c8cad4105745c622), I have tried an early release of it on my xiaomi device and it does work but I wasn't sure how stable it was.

ukriu:
@topjohnwu heya, have you decided on it yet..?
If this is too excessive then supporting only patching via the patch script is a viable option too (since for MTK Xiaomi devices, writing to those partitions in runtime is not permitted) and it can have support partially.
Or what vvb2060 suggested could be viable too, then I'll close this PR.

topjohnwu:
@ukriu my plan is to either
a. have a way to detect this, or
b. add a new option in the app to allow user toggling

I just haven't have time to look into it yet

PR #9275, opened 2025-08-20
Fix `set_zygisk_prop` when nb is not yet reset to the original one

yujincheng08:
Fix test on AVD 28


topjohnwu:
Resolved with #9277

PR #9282, opened 2025-08-22
Use lzma-rust2 for xz and lzma compression and decompression

yujincheng08:


topjohnwu:
lz4-flex doesn't support high compression, which is kinda a deal breaker.

yujincheng08:
Oh I see

yujincheng08:
@topjohnwu revert to use lz4-sys

topjohnwu:
@yujincheng08 can you reorder the commits into these steps?

1. Convert get_decoder to `Box<dyn Read + 'a>` (which means delete FakeReader)
2. Then introduce lzma-rust2

topjohnwu:
Nevermind, I'll convert the change by myself, keep this branch as-is. Thank you!

PR #9299, opened 2025-09-01
Add Hinglish (Hindi-English) translation

MeowDump:
Aims to improve accessibility for Indian users


JasonKhew96:
is this language even exists in Android or Java locale?

ukriu:
@topjohnwu "hn" is not a valid language locale code. "Hinglish" seems to be Hindi with latin alphabets as the alternative to the Hindi ones. You cannot even access this in the app?
https://en.m.wikipedia.org/wiki/List_of_ISO_639_language_codes

PR #9397, opened 2025-10-01
Only reset NB prop when zygisk is enabled

aviraxp:


native/src/core/bootstages.rs
```diff
@@ -185,7 +185,9 @@ impl MagiskD {
 
         setup_preinit_dir();
         self.ensure_manager();
-        self.zygisk.lock().unwrap().reset(true);
+        if self.zygisk_enabled.load(Ordering::Acquire) {
```
yujincheng08: `Relaxed` is enough

native/src/core/daemon.rs
```diff
@@ -130,7 +130,9 @@ impl MagiskD {
                 info!("** zygote restarted");
                 self.prune_su_access();
                 scan_deny_apps();
-                self.zygisk.lock().unwrap().reset(false);
+                if self.zygisk_enabled.load(Ordering::Acquire) {
```
yujincheng08: `Relaxed` is enough

Issue #9409, opened 2025-10-02
argh breaks resetprop cli


aviraxp:
argh takes argument which starts with "-" in any location, which makes "resetprop a.b.c --1" does not work as "--1" is taken as a command of resetprop itself.

Device: 1
Android version: 16
Magisk version name: 30.4
Magisk version code: 30400

ayasa520:
<img width="1050" height="240" alt="Image" src="https://github.com/user-attachments/assets/37543901-6911-4604-aaa3-cd89c61d3434" />

We have to use `--` to stop option parsing.

aviraxp:
It should be fixed in resetprop, to keep same behvior with system setprop.

PR #9411, opened 2025-10-03
Stop parsing options after finding first positional argument

aviraxp:
This fixes https://github.com/topjohnwu/Magisk/issues/9409.


vvb2060:
`magiskboot unpack --abc.img` ?
```
magiskpolicy --load 
magisk --clone-attr 
magisk --denylist exec 
resetprop --file
……
```
All options that can input arbitrary content need to be processed

topjohnwu:
Replaced with #9442

PR #9481, opened 2025-11-01
Add informational guide for rooting Android 13 with Magisk

jancabigas29-source:
> informational guide on rooting Android 13 with Magisk, outlining risks, precautions, and safe practices without providing actionable instructions.


topjohnwu:
Thanks for your contribution, but we don't need these kind of documentation

PR #9543, opened 2025-12-02
Remove unnecessary file system permissions

aviraxp:
Removed permissions for mounting loop devices, mirrors, and tmpfs.


aviraxp:
Seems a leftover from module image age. I am not sure if magiskinit needs it too.

PR #9621, opened 2026-01-03
Create SECURITY.md for security policy

Arif999000:
Added a security policy document outlining supported versions and vulnerability reporting.


Arif999000:
> Added a security policy document outlining supported versions and vulnerability reporting.

PR #9746, opened 2026-03-06
Update faq.md

pndwal:
Add information on automatic safe mode trigger using bootloop counter.


pndwal:
Oh... So it was inaccurate?

PR #9755, opened 2026-03-10
Bring Genymotion support for testing

JulienBolard:
We are currently working on native Genymotion Magisk support, and this is just some contributions to unlock the rest of our work in progress.

That should also help others to test Magisk on Genymotion until then.


JulienBolard:
Could you explain why these fixes cannot land in Magisk official source code?

yujincheng08:
These are not fixes...

PR #9812, opened 2026-04-19
Fix DataSourceChannel

vvb2060:


topjohnwu (commented):
Can you remove this commit ([54f4de1](https://github.com/topjohnwu/Magisk/pull/9812/commits/54f4de1872d60563c9dfeac578881fca479cee0a)) from the PR? I would prefer to always use Android Studio's internal JDK (which is still JDK 21 at the moment), and would prefer NOT to use gradle daemon JVM resolvers.

vvb2060:
I also want to using the IDE's built-in JDK, but JEP 513 requires JDK25.

topjohnwu:
Can refactor the code to avoid using JEP 513?

topjohnwu (approved):

PR #9820, opened 2026-04-21
manager: extend versionCode to Long to exceed Int.MAX

Numbersf:
Fixed an issue where versionCode would get stuck at 2147483647. Updated data type to Long to support larger version numbers. Some modules call multiple parameters for versionCode, causing it to reach 10 digits, which may cause issues on Android 21+. Please check for any omissions.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -84,7 +84,7 @@ This is the **strict** format of `module.prop`
 id=<string>
 name=<string>
 version=<string>
-versionCode=<int>
+versionCode=<long>
 author=<string>
 description=<string>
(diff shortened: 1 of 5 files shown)


vvb2060:
The documentation requires that the module version code be of type int, and I believe modules should comply with this requirement.
https://github.com/topjohnwu/Magisk/blob/9d0b5298afd54b320767af7019645f03e96f362f/docs/guides.md?plain=1#L87

docs/guides.md
tomekwiola2-ai, line 1: Run

PR #9865, opened 2026-05-24
Backup original sepolicy db for LD_PRELOAD strategy

aviraxp:


nndmt034 (commented):
https://github.com/topjohnwu/Magisk/blob/aviraxp-patch-1/native%2Fsrc%2Finit%2Fselinux.rs

topjohnwu (approved):

PR #9970, opened 2026-07-23
init: check charger mode before force_normal_boot

### LINKED ISSUE #9967: Bootloop protection counts off-mode (charger-mode) boots -> Safe Mode falsely triggered on Motorola
Device: Motorola moto XT2435 (MediaTek MT6855). Also reproduces on Motorola XT2533 (Qualcomm Snapdragon SM7435).
Android version: 14 (also reproduced on Android 15)
Magisk version name: 30.7 (debug build, reports `30.7:MAGISK:D`)
Magisk version code: 30700

**Reproduced on a debug build with no modules installed.**

## Summary
On Motorola devices, *off-mode charging* (device powered off, plugged into a charger, showing the charging animation) runs a near-full boot that reaches `post-fs-data`, so `magiskd` increments the bootloop counter — but a charger boot never reaches boot-complete, so the counter is never reset. Two off-mode-charging boots with no full boot in between raise the counter to 2, and the next normal boot falsely enters **Safe Mode** (Zygisk turned off, all modules disabled). This happens from ordinary charging-while-off, not a real bootloop.

## Root cause (`native/src/core/bootstages.rs`)
```rust
let boot_cnt = self.get_db_setting(DbEntryKey::BootloopCount);   // read
self.set_db_setting(DbEntryKey::BootloopCount, boot_cnt + 1)…;   // increment
let safe_mode = boot_cnt >= 2 || …;                              // trigger
```
The counter is cleared only at boot-complete:
```rust
// Reset the bootloop counter once we have boot-complete
self.set_db_setting(DbEntryKey::BootloopCount, 0).log_ok();
```
On Motorola, off-mode charging is implemented as a near-full boot: the bootloader shows its normal (unlocked) warning screen, then the patched boot image + init run through `post-fs-data` with `/data` mounted, so `magiskd` runs `post_fs_data` (`ro.bootmode=charger`) and increments `BootloopCount`. It then stays at the charger UI and never reaches boot-complete, so the counter is never reset. There is no `charger`/`ro.bootmode` guard anywhere in the boot-stage code.

## Evidence (debug v30.7, no modules)
Two consecutive off-mode-charging boots, `ro.bootmode=charger`, counter (bootloop DB value) never reset:
```
charger boot #1 : ro.bootmode=charger   bootloop=1
charger boot #2 : ro.bootmode=charger   bootloop=2
```
`getprop persist.sys.boot.reason.history`:
```
reboot,charger → reboot,charger → cold,powerkey
```
The next normal boot (`magisk.log`):
```
Magisk 30.7(30700) daemon started
** post-fs-data mode running
* Initializing Magisk environment
* Safe mode triggered
** boot-complete triggered
```
Result: `zygisk=0`, all modules disabled.

## Why it is OEM-dependent, not SoC / not "low battery"
Reproduced on both a MediaTek (MT6855) and a Qualcomm (Snapdragon SM7435) Motorola, so it is not chip-specific — it is Motorola's off-mode-charging implementation running the full init path (mounts `/data`, runs `post-fs-data`). OEMs whose off-mode charging uses a lightweight path never run `magiskd`, so the counter never moves and they are unaffected.

## Steps to reproduce (stock/official build, Motorola)
1. Boot normally: `magisk --sqlite "SELECT value FROM settings WHERE key='bootloop'"` → `0`.
2. Power off. Plug in charger → charging animation (off-mode charging).
3. Unplug (device powers off), plug in again → charging animation. Do this **≥2 times with no boot to system in between**.
4. Power on into the OS → `magisk.log` shows `* Safe mode triggered`; Zygisk off; modules disabled.

## Suggested fix
Don't count charger-mode boots toward bootloop protection. In `post_fs_data`:
```rust
if get_prop(cstr!("ro.bootmode")) != "charger" {
    let boot_cnt = self.get_db_setting(DbEntryKey::BootloopCount);
    self.set_db_setting(DbEntryKey::BootloopCount, boot_cnt + 1)…;
    // safe_mode check …
}
```
i.e. treat `androidboot.mode=charger` boots as not counting toward bootloop, since they never reach boot-complete by design.

## Related
Likely the same root cause as #9530 (closed with no mechanism identified there).

vv19ca0:
## Problem

Fixes #9967.

Off-mode charging on some devices runs a near-full boot that reaches `post_fs_data` (mounts `/data`, starts magiskd) but never reaches `boot_complete`, so the bootloop counter is incremented and never reset. Two such charger boots leave it at ≥2 and the next real boot falsely enters safe mode (all modules + zygisk disabled).

`b55f597c` already aborts charger mode via the `androidboot.mode` check in `init/init.rs`, but the check is placed *after* the `force_normal_boot` branch:

```rust
} else if self.config.force_normal_boot {
    self.first_stage();
} else if recovery || boot_mode == "charger" {   // unreachable when fnb=1
    self.recovery_or_charger();
```

Devices that set `androidboot.force_normal_boot=1` during off-mode charging take `first_stage()` and never reach the charger check.

Confirmed on two Motorola devices; both expose `androidboot.mode=charger` in charger boot, differing only in `force_normal_boot`:

| device | SoC | charger `force_normal_boot` | v30.7 aborts charger? |
|--------|-----|-----------------------------|-----------------------|
| XT2533 | Snapdragon SM7435 | 0 | yes (already fixed by b55f597c) |
| XT2435 | MediaTek MT6855 | 1 | no |

## Fix

Move the charger check ahead of `skip_initramfs`/`force_normal_boot` (right after the second-stage check) so charger boots abort regardless of `force_normal_boot`, matching how AOSP treats charger mode first. Recovery detection stays after `force_normal_boot` (unchanged).

This replaces the `post_fs_data` counter-guard from the previous revision of this PR: aborting in magiskinit means Magisk never loads in charger mode at all (no magiskd, no modules, no counter to guard) — the same layer `b55f597c` operates on.

## Testing

Real hardware, on both affected Motorola devices:

- **XT2435 (fnb=1, which `b55f597c` does not fix)**: with the patch, repeated off-mode-charging boots no longer start magiskd at all (no charger-mode entry appears in the Magisk log); the bootloop counter stays `0` and safe mode is no longer triggered. Normal boots are unaffected (magiskd runs, modules load, `boot_complete` resets the counter).
- **XT2533 (fnb=0, already handled by `b55f597c`)**: still aborts charger mode after the reorder; normal boots unaffected.

### DIFF
--- a/native/src/init/init.rs
+++ b/native/src/init/init.rs
@@ -148,13 +148,19 @@ impl MagiskInit {
         let argv1 = unsafe { *self.argv.offset(1) };
         if !argv1.is_null() && unsafe { CStr::from_ptr(argv1) == c"selinux_setup" } {
             self.second_stage();
+        } else if unsafe { CStr::from_ptr(self.config.boot_mode.as_ptr()) } == c"charger" {
+            // Charger (off-mode charging) must abort before any normal-boot path
+            // (skip_initramfs / force_normal_boot). Some devices (e.g. Motorola) set
+            // androidboot.force_normal_boot=1 during off-mode charging, which would otherwise
+            // route to first_stage() and load Magisk in charger mode, bumping the never-reset
+            // bootloop counter until safe mode trips. AOSP likewise treats charger first.
+            self.recovery_or_charger();
         } else if self.config.skip_initramfs {
             self.legacy_system_as_root();
         } else if self.config.force_normal_boot {
             self.first_stage();
         } else if cstr!("/sbin/recovery").exists()
             || cstr!("/system/bin/recovery").exists()
-            || unsafe { CStr::from_ptr(self.config.boot_mode.as_ptr()) } == c"charger"
         {
             self.recovery_or_charger();
         } else if self.check_two_stage() {


aviraxp (approved):

TDA-Android:
FB_IMG_1789749816607



Coming soon... Stay tuned with Trend Deep Tech Advance.🚀

At our official warehouse
https://github.com/TDA-Android/TDA-Mox

PR #10001, opened 2026-08-08
Android Studio may not set ANDROID_SDK_ROOT or ANDROID_HOME on windows

ColdWindScholar:


yujincheng08:
Basically, you should set it manually on all platforms.

PR #10069, opened 2026-09-08
Rename docs/install.md to Samsung a31

aa03265211924-ux:
Rename docs/install.md to Samsung a31


aa03265211924-ux:
Rename docs/install.md to Samsung a31

